PR #4, opened 2024-08-16
Generate image at different resolutions

Xuzzo:
Hello and thanks for your work.

In this PR I have included the possibility to generate images at different resolutions.
This can be selected though the `width` and `height` arguments in the `Config` class.

### DIFF
--- a/main.py
+++ b/main.py
@@ -13,8 +13,8 @@ image = flux.generate_image(
     prompt="Luxury food photograph of a birthday cake. In the middle it has three candles shaped like letters spelling the word 'MLX'. It has perfect lighting and a cozy background with big bokeh and shallow depth of field. The mood is a sunset balcony in tuscany. The photo is taken from the side of the cake. The scene is complemented by a warm, inviting light that highlights the textures and colors of the ingredients, giving it an appetizing and elegant look.",
     config=Config(
         num_inference_steps=2,
-        width=256,
-        height=256,
+        height=768,
+        width=1360,
     )
 )
 
--- a/main.py
+++ b/main.py
@@ -13,6 +13,8 @@ image = flux.generate_image(
     prompt="Luxury food photograph of a birthday cake. In the middle it has three candles shaped like letters spelling the word 'MLX'. It has perfect lighting and a cozy background with big bokeh and shallow depth of field. The mood is a sunset balcony in tuscany. The photo is taken from the side of the cake. The scene is complemented by a warm, inviting light that highlights the textures and colors of the ingredients, giving it an appetizing and elegant look.",
     config=Config(
         num_inference_steps=2,
+        width=256,
+        height=256,
     )
 )
 
--- a/src/flux_1_schnell/config/config.py
+++ b/src/flux_1_schnell/config/config.py
@@ -14,7 +14,7 @@ class Config:
             width: int = 1024,
             height: int = 1024,
     ):
-        if width %16 != 0 or height % 16 != 0:
+        if width % 16 != 0 or height % 16 != 0:
             log.warning("Width and height should be multiples of 16. Rounding down.")
         self.width = 16 * (height // 16)
         self.height = 16 * (width // 16)
--- a/src/flux_1_schnell/config/config.py
+++ b/src/flux_1_schnell/config/config.py
@@ -1,6 +1,8 @@
 import mlx.core as mx
 import numpy as np
+import logging
 
+log = logging.getLogger(__name__)
 
 class Config:
     precision: mx.Dtype = mx.float16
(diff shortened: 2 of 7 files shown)


main.py
```diff
@@ -13,6 +13,8 @@
     prompt="Luxury food photograph of a birthday cake. In the middle it has three candles shaped like letters spelling the word 'MLX'. It has perfect lighting and a cozy background with big bokeh and shallow depth of field. The mood is a sunset balcony in tuscany. The photo is taken from the side of the cake. The scene is complemented by a warm, inviting light that highlights the textures and colors of the ingredients, giving it an appetizing and elegant look.",
     config=Config(
         num_inference_steps=2,
+        width=256,
```
filipstrand, line 16: Since `main.py` is probably the first script people run, I think we should default to a slightly bigger resolution, so that the output image is a bit larger. Maybe go with something like this here

```
height=768,
width=1360,
```

The default of `1024x1024` in `config` (as you have done it) is good I think

src/flux_1_schnell/flux.py
```diff
@@ -56,15 +56,15 @@ def generate_image(self, seed: int, prompt: str, config: Config = Config()) -> P
...
 
-        latents = Flux1Schnell._unpack_latents(latents)
+        latents = Flux1Schnell._unpack_latents(latents, config.width, config.height)
         decoded = self.vae.decode(latents)
         return ImageUtil.to_image(decoded)
 
     @staticmethod
-    def _unpack_latents(latents):
-        latents = mx.reshape(latents, (1, 64, 64, 16, 2, 2))
+    def _unpack_latents(latents, width, height):
```
filipstrand, line 64: Should add type hints to this method (I had forgotten it before as well for the `latents` input)

src/flux_1_schnell/config/config.py
```diff
@@ -9,7 +11,13 @@ class Config:
     def __init__(
             self,
             num_inference_steps: int = 4,
+            width: int = 1024,
+            height: int = 1024,
     ):
+        if width %16 != 0 or height % 16 != 0:
```
filipstrand, line 17: Small formatting note: Add space between `%` and `16`

src/flux_1_schnell/models/transformer/transformer.py
```diff
@@ -39,7 +39,7 @@ def predict(
         text_embeddings = self.time_text_embed.forward(time_step, pooled_prompt_embeds)
         encoder_hidden_states = self.context_embedder(prompt_embeds)
         txt_ids = Transformer._prepare_text_ids()
-        img_ids = Transformer._prepare_latent_image_ids()
+        img_ids = Transformer._prepare_latent_image_ids(config.width, config.height)
```
filipstrand, line 42: When I tested this code with a non-square aspect ratio it worked fine, but I found that height and width was swapped. For example with  

```
height=768,
width=1360,
```

I got a tall image (rather than a wider one as expected). The solution for me was to swap the height and width arguments in two places to the following:

```python
img_ids = Transformer._prepare_latent_image_ids(config.height, config.width)
```

```python
latents = Flux1Schnell._unpack_latents(latents, config.height, config.width)
```

And then rename the args in both methods respectively.

src/flux_1_schnell/models/transformer/transformer.py
```diff
@@ -67,12 +67,14 @@ def predict(
         return noise
 
     @staticmethod
-    def _prepare_latent_image_ids() -> mx.array:
-        latent_image_ids = mx.zeros((64, 64, 3))
-        latent_image_ids = latent_image_ids.at[:, :, 1].add(mx.arange(0, 64)[:, None])
-        latent_image_ids = latent_image_ids.at[:, :, 2].add(mx.arange(0, 64)[None, :])
+    def _prepare_latent_image_ids(width, height) -> mx.array:
+        r_h = height // 16
```
filipstrand, line 71: Does the `r_h` and `r_w` stand for `reduced_height` and `reduced_width` or something else? 

Maybe we should go with those full names, or something slightly more descriptive like: 

```
latent_height
latent_width
```


Also: we should add type hints to this method as well.

main.py
Xuzzo: good point. I changed the values

src/flux_1_schnell/models/transformer/transformer.py
Xuzzo: ah yes you are right. I have included your suggestions

filipstrand: Thanks for updating!

filipstrand:
@Xuzzo Thank you very much for this addition. As a last test, I also ran the same prompt and a fixed latent on the Diffusers implementation and it looks identical to mflux for the other resolutions as well. Great job!

PR #5, opened 2024-08-17
Add CLI args for easier execution

explorigin:
Add `argparse` CLI options.

- prompt
- seed
- model
- steps
- output
- width
- height

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ like [Numpy](https://numpy.org) and [Pillow](https://pypi.org/project/pillow/) f
 
 Run the provided [main.py](main.py) 
 ```
-python main.py
+python main.py --prompt "luxury food photograph" --steps 2
 ```
 
 or make a new separate script like the following
--- a/README.md
+++ b/README.md
@@ -135,4 +135,3 @@ Luxury food photograph of an italian Linguine pasta alle vongole dish with lots 
 
 - FLUX Dev implementation
 - LoRA adapters
-- Command line args
--- a/main.py
+++ b/main.py
@@ -1,5 +1,7 @@
 import os
 import sys
+import argparse
+import time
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -135,4 +135,3 @@ Luxury food photograph of an italian Linguine pasta alle vongole dish with lots
 
 - FLUX Dev implementation
 - LoRA adapters
-- Command line args
```
filipstrand, line 138: With the updates in this PR, since running `main.py` now requires a prompt (which is a good default to have as you have done it) the README can also be updated to say something like 

```
python main.py --prompt "luxury food photograph" --steps 2
```

I choose 2 steps in this example here simply because it has worked quite well for my tests and some earlier performance numbers was based on this. But the default of 4 steps in the args is good (I have seen that used elsewhere also).

main.py
```diff
@@ -1,19 +1,37 @@
...
 import sys
+import argparse
+import time
 
 sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), 'src')))
 
 from flux_1_schnell.config.config import Config
 from flux_1_schnell.flux import Flux1Schnell
 
-flux = Flux1Schnell("black-forest-labs/FLUX.1-schnell")
```
filipstrand, line 9: We just recently updated to support setting a custom resolution to the generated image so there is a small conflict in this file. 

With that change I think `--height` and `--width` can now also be input parameters defaulting to 1024 by 1024 if nothing is specified.

filipstrand:
Thanks for the updates to the CLI args and also for your performance report! After this PR is updated and merged, I can update the README a bit with the new performance numbers.

explorigin:
Sorry, I rebased instead of merged. It's pretty small though.

filipstrand:
> Sorry, I rebased instead of merged. It's pretty small though.

No worries, looks good!

PR #10, opened 2024-08-18
Include FLUX.1-Dev

Xuzzo:
Hello and thanks again for your very nice work.
This PR:

- Includes option to use FLUX.1-Dev
- changes name to some core classes, including Flux1Schnell -> Flux1

Note: the flux_1_schnell folder still has the same name here so we can have a better diff and review the changes more easily. Once they are merged, the name of the folders and files will need to be updated too.


Xuzzo:
There are still some minor things that I want to check, but this should be ready for review.
I compared the images created here with those from huggingface and they are very similar. Here an example
Diffusers:
image
Mflux:
image

filipstrand:
@Xuzzo Big thanks for all the help!  I will take a look at this PR later today when I have the time :)

src/flux_1_schnell/config/config.py
```diff
@@ -6,26 +6,36 @@
...
+        self.guidance = guidance
         self.sigmas = mx.concatenate([base_sigmas, mx.zeros(1)])
 
     @staticmethod
     def base_sigmas(num_inference_steps):
         sigmas = np.linspace(1.0, 1 / num_inference_steps, num_inference_steps)
         sigmas = mx.array(sigmas).astype(mx.float32)
         return sigmas
+    
+    def shift_sigmas(self):
```
filipstrand, line 34: To modify the sigmas here is convenient but I can see one potential problem: 

When generating several images I typically write a loop like this: 

```python
for i in range(10):
    image = flux.generate_image(
        seed=3,
        prompt="...",
        config=Config(
            num_inference_steps=3,
            height=256,
            width=256,
            guidance=3.5,
        )
    )

    ImageUtil.save_image(image, "image.png")
```

But it also quite reasonable to move the config outside the loop like so (since it is effectively static):

```python
config = Config(num_inference_steps=3, height=256, width=256, guidance=3.5, )
for i in range(10):
    image = flux.generate_image(
        seed=3,
        prompt="...",
        config=config
    )

    ImageUtil.save_image(image, "image.png")
```

However, in the latter case we will call the `shift_sigmas` multiple times and get the wrong values after the first generation.  Preferably it would be nice to not have any updating of states for the Config class after it has been created.

src/flux_1_schnell/config/config.py
```diff
@@ -6,26 +6,36 @@
 
 
 class Config:
-    precision: mx.Dtype = mx.float16
+    precision: mx.Dtype = mx.bfloat16
```
filipstrand, line 9: Was not aware of the `bfloat16` option in mlx. I am not well read up on this topic, but a quick google search suggests that this is often times preferable to `float16` in these type of applications. Out of curiosity have you seen any performance benefits or decreased memory usage for example?

src/flux_1_schnell/flux.py
```diff
@@ -47,16 +49,12 @@ def generate_image(self, seed: int, prompt: str, config: Config = Config()) -> P
                 config=config
             )
 
-            latents = FlowMatchEulerDiscreteNoiseScheduler.denoise(
-                t=t,
-                noise=noise,
-                latent=latents,
-                config=config
-            )
+            dt = config.sigmas[t + 1] - config.sigmas[t]
```
filipstrand, line 52: Super nice algebraic simplification! I more or less blindly copied it from Diffusers before (only removing some uncessary variables) but did not think of simplifying it further :). This also motivates the removal of the scheduler class as you have done. It is very neat that the denoising step is so simple in this model!  Some time ago, I remember looking at some other scheduler for SDXL (think it was the `DDIMScheduler` or something similar), and it was a lot more complicated compared to this.

src/flux_1_schnell/config/config.py
Xuzzo: The images of diffusers vs mflux were quite different and I did not understand why, cause there was nothing wrong in the code. Turns out, there is a small rounding error when `guidance` is created in `Transformer.predict. In fact
` mx.broadcast_to(3500 , (1,)).astype(mx.float16) = array([3500], dtype=float16)`
but
` mx.broadcast_to(3500 , (1,)).astype(mx.bfloat16) = array([3504], dtype=bfloat16)`
and diffusers uses bfloat16

src/flux_1_schnell/flux.py
Xuzzo: I also find it refreshing that sometimes things become simpler as they progress 🙂 schedulers for diffusion models have gone through quite a revolution recently

filipstrand:
Hi again, I have added some smaller changes on top of your work. Most notably, I have added a `ModelConfig` class that hold a few pieces of information (such as the `max_sequence_length`) for each model. Since the sigmas dependent on both the model configuration and the more "run time" configuration of the `Config` class I begin the `generate_image` method with this: `config = config.copy_with_sigmas(self.model_config)` now. It is very similar to how you did it, but I always make a new copy of the Config (since it is lightweight) to make sure we don't accidentally override anything

main.py
```diff
@@ -9,12 +9,12 @@
...
 from flux_1_schnell.post_processing.image_util import ImageUtil
 
+
 def main():
     parser = argparse.ArgumentParser(description='Generate an image based on a prompt.')
     parser.add_argument('--prompt', type=str, required=True, help='The textual description of the image to generate.')
     parser.add_argument('--output', type=str, default="image.png", help='The filename for the output image. Default is "image.png".')
-    parser.add_argument('--model', type=str, default="black-forest-labs/FLUX.1-schnell", help='The model to use. Default is "black-forest-labs/FLUX.1-schnell".')
-    parser.add_argument('--max_sequence_length', type=int, default=256, help='Max Sequence Length (Default is 256)')
+    parser.add_argument('--model', type=str, default="schnell", help='The model to use. Default is "schnell".')
```
filipstrand, line 17: To keep args short I opted for `schnell` or `dev` for model, and the actual name is resolved in `Flux1.from_alias(args.model)` The new `ModelConfig` holds the full model names

main.py
```diff
@@ -9,12 +9,12 @@
 from flux_1_schnell.flux import Flux1
 from flux_1_schnell.post_processing.image_util import ImageUtil
 
+
 def main():
     parser = argparse.ArgumentParser(description='Generate an image based on a prompt.')
     parser.add_argument('--prompt', type=str, required=True, help='The textual description of the image to generate.')
     parser.add_argument('--output', type=str, default="image.png", help='The filename for the output image. Default is "image.png".')
-    parser.add_argument('--model', type=str, default="black-forest-labs/FLUX.1-schnell", help='The model to use. Default is "black-forest-labs/FLUX.1-schnell".')
-    parser.add_argument('--max_sequence_length', type=int, default=256, help='Max Sequence Length (Default is 256)')
```
filipstrand: I also removed the `max_sequence_length` arg and made it a property of the `ModelConfig`. But if you have other ideas of why it can be good to send this in please tell me :)

src/flux_1_schnell/config/config.py
```diff
@@ -1,28 +1,13 @@
...
+import mlx.core as mx
+import numpy as np
 
-def get_sigmas(num_inference_steps):
-    sigmas = np.linspace(1.0, 1 / num_inference_steps, num_inference_steps)
-    sigmas = mx.array(sigmas).astype(mx.float32)
-    return mx.concatenate([sigmas, mx.zeros(1)])
+from flux_1_schnell.config.model_config import ModelConfig
 
-def shift_sigmas(sigmas, width, height):
```
filipstrand: I originally come from more of a java background so I am typically more fond of putting method like this as static ones under classes. I know in python this is not at all as common, but that is why you see it used so much in my code

src/flux_1_schnell/flux.py
```diff
@@ -18,44 +19,60 @@
...
-        sigmas = get_sigmas(config.num_inference_steps)
-        if self.is_dev:
-            sigmas = shift_sigmas(sigmas, config.width, config.height)
+        # Create a new config with sigmas based on what model we are running
+        config = config.copy_with_sigmas(self.model_config)
+
+        # Create the latents
         latents = LatentCreator.create(config.height, config.width, seed)
 
+        # Embedd the prompt
```
filipstrand, line 52: Added these comments more as visual "dividers" in this method since now we can do more things inline (e.g the the denoising process after your simplification), rather than to delegate to separate methods that can be named

src/flux_1_schnell/config/model_config.py
```diff
@@ -0,0 +1,29 @@
+from enum import Enum
+
+
+class ModelConfig(Enum):
```
filipstrand, line 4: This is the main addition. I modelled this as an enum for now since it holds relatively few attributes. If this were to grow and hold more information over time, maybe it would be better to have dedicated classes or something else, but I think it works for now

Xuzzo:
Thanks for your additions. Looks good to me. The only thing I am not fully sure about is the config logic: I dont think the sigma methods should be there, since they have external dependencies (i.e. model type). It does not hurt too much for now, but maybe having another class that takes both config and ModelConfig would be better? something like

```
class RuntimeConfig:
    def __init__(self, config: Config, model: ModelConfig):
        self.model = model
        self.config = Config()
        self.sigmas = self._get_sigmas(config.num_inference_steps)
        if model == ModelConfig.FLUX1_DEV:
            self.sigmas = self._shift_sigmas(self.sigmas, config.width, config.height)

    @staticmethod
    def _get_sigmas(num_inference_steps):
        ...

    @staticmethod
    def _shift_sigmas(sigmas: mx.array, width: int, height: int):
        ...
```
and in this case Config and ModelConfig can be frozen after init.

src/flux_1_schnell/config/model_config.py
Xuzzo: LGTM. Thanks 🙂

filipstrand:
> Thanks for your additions. Looks good to me. The only thing I am not fully sure about is the config logic: I dont think the sigma methods should be there, since they have external dependencies (i.e. model type). It does not hurt too much for now, but maybe having another class that takes both config and ModelConfig would be better? something like
> 
> ```
> class RuntimeConfig:
>     def __init__(self, config: Config, model: ModelConfig):
>         self.model = model
>         self.config = config
>         self.sigmas = self._get_sigmas(config.num_inference_steps)
>         if model == ModelConfig.FLUX1_DEV:
>             self.sigmas = self._shift_sigmas(self.sigmas, config.width, config.height)
> 
>     @staticmethod
>     def _get_sigmas(num_inference_steps):
>         ...
> 
>     @staticmethod
>     def _shift_sigmas(sigmas: mx.array, width: int, height: int):
>         ...
> ```
> 
> and in this case Config and ModelConfig can be frozen after init. Then you can pass around a RuntimeConfig object and refer to anything as we were doing before.

Yes that might actually be a bit more clean and separated. I'll try it out and and lets see how it fits...

src/flux_1_schnell/config/runtime_config.py
```diff
@@ -0,0 +1,58 @@
...
+
+class RuntimeConfig:
+
+    def __init__(self, config: Config, model_config: ModelConfig):
+        self.config = config
+        self.num_train_steps = 1000
+        self.sigmas = self._create_sigmas(config, model_config)
+
+    @property
+    def height(self):
```
filipstrand, line 16: Since `RuntimeConfig ` wraps the other `Config` class, I exposed some properties so that we don't need to dig one level deep each time to get things like height and width and so on.

src/flux_1_schnell/config/runtime_config.py
```diff
@@ -0,0 +1,58 @@
...
+
+from flux_1_schnell.config.config import Config
+from flux_1_schnell.config.model_config import ModelConfig
+
+
+class RuntimeConfig:
+
+    def __init__(self, config: Config, model_config: ModelConfig):
+        self.config = config
+        self.num_train_steps = 1000
```
filipstrand, line 12: Also moved `num_train_steps` to this class. One could perhaps argue that it should actually be in `ModelConfig` instead now that I think about it... Now the original `Config` class really only shows what can be set by the user at generation time. Maybe it should be renamed to something like `UserConfig` or `ImageConfig` or something similar?

Xuzzo: Probably `num_train_steps ` is better in `ModelConfig `.  No preference for the `Config` name. If we opt to be precise, then maybe something like `ImageGenerationConfig` would do. But even just `Config` works for me.

filipstrand:
Thanks again for your help. After this is merged and tested a bit more I'll probably send out a tweet about it. Just wanted to check that it is OK that I mention your twitter handle https://x.com/f_xuzzo there :) ?

Xuzzo:
ah thanks! yes that's me 🙂

filipstrand:
> ah thanks! yes that's me 🙂

Great!

PR #16, opened 2024-08-21
Add Image to Image option

Xuzzo:
Hi,
this PR:
- introduces the possibility to modify another image through flux1
- fixes a bug with width <-> height (they were swapped in config init)
- adds main_img2img for generation. Note: not sure this is needed, we could probably unify cli and keep only one main


filipstrand:
@Xuzzo Again, thanks for the contribution. Very nice that you spotted and fixed the `height/width` mistake.   

Regarding this feature, do you have a good example of it working? I only tried it quickly with the schnell model for 4 steps, and did not get any special result from it (I basically got back the input image with minor variations). Maybe it might not work as well with the distilled `schnell` model??

Xuzzo:
Using the `schnell` model I found it is quite important to tune the `strength` parameter. For example:

Original image:
image

CLI:
`python main_img2img.py --prompt "Luxury food photograph" --steps 10 --seed 2 --base-image image.png --strength 0.3`

Result:
image(37)

I found the dev to be a bit less sensitive to `strength`.

filipstrand:
> Using the `schnell` model I found it is quite important to tune the `strength` parameter. For example:
> 
> Original image: image
> 
> CLI: `python main_img2img.py --prompt "Luxury food photograph" --steps 10 --seed 2 --base-image image.png --strength 0.3`
> 
> Result: image(37)
> 
> I found the dev to be a bit less sensitive to `strength`.

Oh great, this is a very nice example! I think we should add this to the readme.

filipstrand:
Btw, I have looked at your PR, and will share some comments later today. Thanks for the patience :)

src/flux_1/config/config.py
```diff
@@ -15,9 +15,46 @@ def __init__(
...
+            num_inference_steps = 1
+        if width % 16 != 0 or height % 16 != 0:
+            log.warning("Width and height should be multiples of 16. Rounding down.")
+        self.width = 16 * (width // 16)
+        self.height = 16 * (height // 16)
+        self.num_total_denoising_steps = num_inference_steps
+        self.inference_steps = list(range(num_inference_steps))
+        self.guidance = guidance
+
+class ConfigImg2Img:
```
filipstrand, line 29: More of a personal preference, but I would prefer one file per config class.

src/flux_1/config/config.py
```diff
@@ -15,9 +15,46 @@ def __init__(
...
+    precision: mx.Dtype = mx.bfloat16
+
+    def __init__(
+            self,
+            num_inference_steps: int = 4,
+            width: int = 1024,
```
filipstrand, line 35: With the img2img option, does it make sense to also provide a height and width? I am thinking that it will always be based on the input base image?

main_img2img.py
```diff
@@ -0,0 +1,51 @@
...
+            width=args.width,
+            guidance=args.guidance,
+            strength=args.strength
+        )
+    )
+
+    ImageUtil.save_image(image, args.output)
+
+
+if __name__ == '__main__':
```
filipstrand, line 50: Regarding `main_img2img.py` I think I prefer this option, instead of unifying things too much at this point (not fully sure how often this feature will be used compared to text2img), so I actually prefer to keep this a bit separate for now as you have done

src/flux_1/flux.py
```diff
@@ -79,18 +85,52 @@ def generate_image(self, seed: int, prompt: str, config: Config = Config()) -> P
...
+        noise = mx.random.normal(
+            shape=[1, (runtime_config.height // 16) * (runtime_config.width // 16), 64],
+            key=mx.random.key(seed)
+        )
+        base_image = ImageUtil.to_array(base_image)
+        image_latents = self.vae.encode(base_image)
+        image_latents = self._pack_latents(image_latents, runtime_config.height, runtime_config.width)
+        latents = runtime_config.sigmas[config.init_timestep] * noise + (1.0 - runtime_config.sigmas[config.init_timestep]) * image_latents
+
+        return self._generate_from_latents(latents, prompt, runtime_config)
```
filipstrand, line 135: This is also a bit more of a personal opinion, but I think I even would prefer a separate `flux_img2img.py` file that holds the new class at this point.  Right now, I would also not even subclass it and deal with some code duplication for the img2img case. 

I guess the thing I like is for the standard text2img workflow to be as straight forward as possible and I also kinda like being able to view the whole `generate_image` method at once, without it delegating to another method like `_generate_from_latents`. This would mean some more code duplication right now, but I think it is OK at the moment.  

Based on your img2img logic, I have made a [branch based on this one](https://github.com/filipstrand/mflux/tree/img2img-based-on-Xuzzo-dont-merge-this-branch) of what this could look like. 

To be clear, since you have added the extra logic for img2img, I think we should continue on your current branch here, but make some changes similar to how it looks like in my other branch. 

Would be nice to hear your opinions on this, and also let me know if you have the time or not to do it, otherwise I can add some commits on top of your branch here :)

Xuzzo: No problem for me. Feel free to push directly here

iLoveBug:
any progress for this?

anthonywu:
👍🏼 to getting img2img implemented, though this change would need to be re-constructed on latest `main` changes. I propose doing this after #70 and #66 and maybe another refactoring where the Flux classes become sub-classes of a general `FluxBase` where shared implementation go.

anthonywu:
I have a draft PR that re-constructs this PR using some WIP-refactoring code.

Open to review: https://github.com/anthonywu/mflux/pull/1/files

Interface is ready. The image generation is buggy, I think I'm a "few lines of code" away from getting it to work, just ran out of time today. The structure of the re-implementation is worth a preview though for anyone interested in helping expedite this.

main_img2img.py
anthonywu: after working on my own take in https://github.com/anthonywu/mflux/pull/1 - my current opinion is the config classes should be mixins, instead of RuntimeConfig being parent object of ModelConfig, etc.

src/flux_1/config/config.py
anthonywu: my current position is we probably don't need multiple config classes that map 1 config to 1 CLI entry point.

src/flux_1/config/config.py
anthonywu: this depends on the training resolution of Flux, my understanding (from reading forums) is that snapping the output image to the training resolutions gets better results. Worth doing some evals here, or just follow evals from similar projects upstream.

anthonywu (commented):
thanks to @Xuzzo's ref implementation here I found a bug fix for my PR over at https://github.com/anthonywu/mflux/pull/1

my first test was: `mflux-generate --model schnell --init-image cake.png --init-image-strength 0.7 --prompt "a blue cake"`

image

will be doing more tests by varying the prompts and strengths, but my PR feels directionally correct

Xuzzo:
Thanks @anthonywu for moving this forward. Im closing this PR and leaving all development to #77

PR #21, opened 2024-08-23
Add Stream generate image functions

madroidmaq:
Add the `stream_generate_image` function to expose the results of each step of the inference for follow-up GUI interaction. 

The usage would look something like this:

```python
def stream_demo(args):
    seed = int(time.time()) if args.seed is None else args.seed

    flux = Flux1.from_alias(args.model)

    stream = flux.stream_generate_image(
        seed=seed,
        prompt=args.prompt,
        report_step=5,
        config=Config(
            num_inference_steps=args.steps,
            height=args.height,
            width=args.width,
            guidance=args.guidance,
        ),
    )

    step_count = 0

    for image in stream:
        step_count += 1

        intermediate_output_path = f"{args.output}_report_step_{step_count}.png"
        ImageUtil.save_image(image, "build/stream/" + intermediate_output_path)

    print(f"Generation complete. Total steps: {step_count}")

stream_demo()
```

### DIFF
--- a/main.py
+++ b/main.py
@@ -11,7 +11,7 @@ from flux_1.flux import Flux1
 from flux_1.post_processing.image_util import ImageUtil
 
 
-def main():
+def setup_arg_parser():
     parser = argparse.ArgumentParser(description='Generate an image based on a prompt.')
     parser.add_argument('--prompt', type=str, required=True, help='The textual description of the image to generate.')
     parser.add_argument('--output', type=str, default="image.png", help='The filename for the output image. Default is "image.png".')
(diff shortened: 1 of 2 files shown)


anthonywu (commented):
This PR needs a bit of catch up to the latest in `main`, but I think as a user this would be very helpful during parameter experimentation where I want to see the trade-off between compute/quality.

To extend this work, I propose a hook to "quit" the generation if the intermediate outputs are unsatisfactory and the user should be able to give a signal to stop further work. This is especially relevant in the `dev` model where generations can take over 5 minutes (on my M1 Max 64GB), and even slower on most users machines. Being able to move onto the next experiment would be a great win.

Would love to see this PR developed further.

anthonywu:
@madroidmaq check out PR #59 to see if that satisfies your original intent.

madroidmaq:
@anthonywu Good jobs ! My initial goal was to see the image generation process in Streamlit, and the implementation on my machine looks pretty good. I see your support is even stronger, so I think I can close this PR. In the future, if others have similar needs, they can refer to your PR.

filipstrand:
@madroidmaq sorry this PR and feature were deprioritized for a bit. I'll review the open PR once the ControlNet feature is merged.

PR #29, opened 2024-09-02
added a flask server for asynchronous image generation with mflux

Orbiter:
I would like to contribute a server front-end to this outstanding project.
For simplicity, I did not write a README yet for the functionality of the server, but most of what one would write down as a documentation is in this pull request below. If required/wanted I can put this in a separate pull request as `README_server.md`. So this pull request ist just a single file. I hope you like it.

# API Server for mflux

This API server is designed for asynchronous image generation tasks, particularly optimized for environments where GPU resources need to be shared across multiple tasks, such as in generative AI chat programs. This server ensures that only one image generation task runs at a time to efficiently use GPU resources.

The API supports features such as:
- Queuing of image generation tasks.
- Forecasting computation time for better user experience in multi-user environments.
- Managing task statuses and retrieving generated images.

Furthermore, the API exposes a swagger endpoint to self-document the server.

## Example usage

The server can be started with i.e.
```
python3 server.py --quantize 8
```

It then runs on port 4030 by default. Host an port can be configured by call parameters, try `server.py --help`. To see the swagger documentation, open http://localhost:4030/swagger

To produce an image, the usual workflow is:
- `/generate` to initialize the generation, this returns a `task_id`
- `/status` to check if the image generation has finished. If not, this also returns an approximated waiting time which can be used in a front-end to show a progress bar
- `/image` to retrieve the produced image as soon as the status turns to "done"

In detail - here is a call to generate an image:

```
curl -X 'POST' \
  'http://localhost:4030/generate' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "prompt": "A beautiful landscape",
  "seed": "1725311496",
  "height": 1024,
  "width": 1024,
  "steps": 4,
  "guidance": 3.5,
  "format": "JPEG",
  "quality": 85,
  "priority": false
}'
```

The response is then an object i.e. with:

```
{
  "task_id": "1fc9cc4f",
  "task_length": 1,
  "expected_time_seconds": 146.88675427253082
}
````

The `task_id` can then be used to check the image generation status:

```
curl -X 'GET' \
  'http://localhost:4030/status?task_id=1fc9cc4f' \
  -H 'accept: application/json'
```

a response would be i.e.:
```
{
  "pos": 0,
  "status": "waiting",
  "wait_remaining": 15
}
```
... which means that the image is not ready and is expected to be ready in 15 second. The position is 0 which means that no other image is in queue before.

Finally, the image can be retrieved with:

```
curl -X 'GET' \
  'http://localhost:4030/image?task_id=1fc9cc4f' \
  -H 'accept: application/json'
```

This returns the jpeg binary and removes the image from the production queue.

There are more API endpoints to list the queue and delete entries from the queue, see swagger documentation for details.

## python client (quick example)

Here are three functions which implement a client endpoint for the image generation process as shown above with curl:

```
def mflux_generate_client(mfluxendpoint, prompt, width=1280, height=720, steps=4, count=4, seed=None, format="JPEG", quality=85, priority=False):
    mfluxendpoint = "http://localhost:4030"
    data = {
        "prompt": prompt,
        "height": height,
        "width": width,
        "steps": steps,
        "format": format,
        "quality": quality,
        "priority": priority
    }
    if seed is not None:
        data["seed"] = seed
    response = requests.post(mfluxendpoint + "/generate", json=data)
    # parse the response and get the task_id
    json = response.json()
    task_id = json["task_id"]
    return task_id
    
def mflux_status_ready(mfluxendpoint, task_id):
    response = requests.get(mfluxendpoint + "/status?task_id=" + task_id)
    if response.status_code == 200:
        status = response.json()["status"]
        if status == "done":
            return 0
        else:
            # read the waiting time
            waiting_time = response.json().get("wait_remaining", 1)
            if waiting_time < 1: waiting_time = 1
            return waiting_time
    else:
        return -1

def mflux_get_image(mfluxendpoint, task_id):
    response = requests.get(mfluxendpoint + "/image?task_id=" + task_id + "&base64=false&delete=true")
    if response.status_code == 200:
        return response.content
    else:
        return None
```

The mfluxendpoint would be a string like `http://localhost:4030`. 
A single function which uses the client endpoints above to get an image can be i.e.:

```
def generate_image(mfluxendpoint, prompt, width=1280, height=720, steps=4, count=1):
    startt = time.time()
    task_id = mflux_generate_client(mfluxendpoint, prompt, width=width, height=height, steps=steps, count=count)
    for i in range(10000):
        waiting_time = mflux_status_ready(mfluxendpoint, task_id)
        print("Waiting time: ", waiting_time, " seconds")
        if waiting_time == 0: break
        nextsleep = max(min(waiting_time / 4, 10), 1)
        time.sleep(nextsleep)
    imageb = mflux_get_image(mfluxendpoint, task_id)    
    stopt = time.time()
    print("Time taken: ", stopt - startt, " seconds")
    image = Image.open(BytesIO(imageb))
    return image
```

### CODE REVIEW

File: src/mflux/server.py [hunk 1]
```diff
@@ -0,0 +1,360 @@
...
+import time
+import base64
+import hashlib
+import threading
+from PIL import Image
+from flask import Flask, request, Response, jsonify
+from flask_restx import Api, Resource, fields
+from flask_cors import CORS
+
+sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
```
anthonywu, line 18: this is a bad code smell, you're hacking package detection that should be handled by a proper venv setup where `mflux` is the package root, I suspect you developed this PR relying on relative directory paths. Perhaps a temp hack you forgot to remove after proper project config.

easy fix though - just remove this line and things should work in a proper venv

anthonywu (commented):
my community/contributor vote is to isolate these add-ons as their own GitHub repos and PyPi artifacts, e.g. `pip install mflux-flask-server`. It's going to be hard for a core project to bundle one server amongst many equally valid choices, e.g. `Flask` vs. `FastAPI` vs. `FastHTML` vs. the automatic API endpoints in `gradio`. Puts too much burden on the foundation repo to maintain frontends.

src/mflux/server.py
Orbiter: ah, no, I copied this from generate.py:
https://github.com/filipstrand/mflux/blob/ca4e3f4c7922fdb5052df937c98e8ee5363bd664/src/mflux/generate.py#L6

anthonywu: yah, anti-pattern, I will make some suggestions on packaging best practices soon in my own PR

filipstrand: Yes, this is now removed as #55 have been merged to main

Orbiter:
Sorry for not updating my code in this pull request. Meanwhile I made many changes/additions and I admit that an integration of the server application in your codebase would require a high maintenance. It's probably better to provide the server in an outside project.

Thats what I did meanwhile, the server is now here: https://github.com/Orbiter/mflux-server

It comes with (so far) two client UI applications, one is integrated and provides a multi-image generation+download function: 

screenshot

That means it's probably correct to close this pull request without pulling.

anthonywu:
> This server ensures that only one image generation task runs at a time to efficiently use GPU resources.

Yes this is a very real problem on these local apps. GPU utilization is full each run. I actually thought the CLI was a nice clean interface to ensure resource reclamation after each run.

https://github.com/anthonywu/mflux-fasthtml/blob/9a6d02ac7de37fbe19a1078a5b94a92c9646fc5f/src/mflux_fasthtml/app/worker.py#L34 was my attempt to queue one at a time and there remains many problems to overcome with resource allocation issues between generations.

Orbiter:
I see that you are performing a forced mlx cleanup here:
https://github.com/anthonywu/mflux-fasthtml/blob/9a6d02ac7de37fbe19a1078a5b94a92c9646fc5f/src/mflux_fasthtml/app/worker.py#L16

I am running my mflux-server in (multi-user) production together with several llama.cpp instances at the same time at the same server and I cannot see resource allocation problems - so far. My believe was that all resources are here just main memory and I have plenty of that. I can easily have all models (flux, llama etc) in the memory at the same time, so what do you mean exactly with "resource reclamation"?

anthonywu:
On my M1 Max 64GB I was doing iterative development of the web ui wrapper and when I would let the uvicorn app reload on file change, or if I Ctrl-C out of the program, the memory pressure remained - then in subsequent runs my Mac would freeze and I need to kill -9 the old process. I suspect it has something to do with threading usage so I have to fix it or switch that out.

Another concern is that when changing model or loras we need to reload the weights, because otherwise the memory requirement of the program would just keep growing.

Finding lots of gotchas when trying to generate many images in a row while loading/unloading between generations. Something I'll follow up on when I have time.

Orbiter:
Oh wow, this is a very important hint. While I never experienced a problem on the production machine (192GB) I should have monitored memory pressure as you did. Now I did so and observed what you described. I took inspiration from your memory cleanup function and added this in mflux-server now here: https://github.com/Orbiter/mflux-server/commit/2a9e0d6ef4929c040d9645c948575149fdf564e2

After the memory cleanup, especially `metal.clear_cache()` the memory usage of the python process always went back to about 17GB.

I also added an option to reduce the cache size that mlx may use with `metal.set_cache_limit(metal_cache_limit)`. Experiments using here only some GB (i.e. 3GB) showed that the whole memory usage actually stayed small and the computation speed did not really showed any strong increase.

anthonywu:
Glad to hear that was helpful.

I think once we find some good practices the cleanup helpers should migrate into the core repo here.

anthonywu:
@Orbiter - with your project setup at your own repo, shall we close this PR?

Orbiter:
yes

PR #31, opened 2024-09-05
Fix width and height in config

Xuzzo:
Since PR #16 has fallen behind in priority, I thought of bringing here the fix for the width and height mix-up in the config class 

This PR:
- fixes a bug in config, where width and height are mixed
- this was not causing any issues downstream and is just a code cleanup


filipstrand:
@Xuzzo Oh, great suggestion! Will merge this fix right away.

Yes sorry about having #16 open for a while with no updates... Have not forgotten about it, just that it has fallen a bit behind in priority as you said, will pick this one up again eventually :)

PR #48, opened 2024-09-14
Add controlnet

Xuzzo:
This PR:
- adds controlnet
- for now, only the Canny edges variant.
- adds cli for new code


Xuzzo:
I have tried to keep the rest of the code and this new section separated. This means that there are quite a lot of repetitions, but I think we can live with that. I hope I will be able to add other controlnets soon.

anthonywu (changes requested):
Driving by as an user/observer of this repo to help test and review this change.

- Left some suggestions in inline comments. Easy stuff.
- Withholding nitpicks on style/linting - deferring to @filipstrand on aesthetic guidelines. @filipstrand may I suggest adding a `README` section on what commands to run to validate/process a file before PR, e.g. `ruff check` and alike.

I've been testing on `dev` model with `steps=16` and getting some reasonable results. I'm not expert eyed enough to spot what might be missing, which is to say - it's generating scenes consistent with the control image and I'm happy with the state of this PR. Thanks @Xuzzo for spearheading this work!

filipstrand:
Thanks for this contribution @Xuzzo! I have not played around with control-net for Flux so this will be really interesting to try out!  I'll go over the PR soon.

src/mflux/controlnet/flux_controlnet.py
```diff
@@ -0,0 +1,321 @@
...
+
+from mflux.config.model_config import ModelConfig
+from mflux.config.runtime_config import RuntimeConfig
+from mflux.models.transformer.ada_layer_norm_continous import AdaLayerNormContinuous
+from mflux.models.transformer.embed_nd import EmbedND
+from mflux.models.transformer.joint_transformer_block import JointTransformerBlock
+from mflux.models.transformer.single_transformer_block import SingleTransformerBlock
+from mflux.models.transformer.time_text_embed import TimeTextEmbed
+import numpy as np
+import cv2
```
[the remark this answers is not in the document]
filipstrand: Great catch! Also, as of recently, the project has two places where we list the dependencies: `pyproject.toml` and `requirements.txt` (not ideal, and can probably be updated to just the toml file), so that can be good to keep in mind.

Small additional note: Some imports here (like `cv2`) are not used in this file, but this dependency have been moved to `utils_controlnet.py`

src/mflux/generate_controlnet.py
```diff
@@ -0,0 +1,65 @@
...
+def main():
+    parser = argparse.ArgumentParser(description='Generate an image based on a prompt.')
+    parser.add_argument('--prompt', type=str, required=True, help='The textual description of the image to generate.')
+    parser.add_argument('--control-image-path', type=str, required=True, help='Local path of the image to use as input for controlnet.')
+    parser.add_argument('--output', type=str, default="image.png", help='The filename for the output image. Default is "image.png".')
+    parser.add_argument('--model', "-m", type=str, required=True, choices=["dev", "schnell"], help='The model to use ("schnell" or "dev").')
+    parser.add_argument('--seed', type=int, default=None, help='Entropy Seed (Default is time-based random-seed)')
+    parser.add_argument('--height', type=int, default=1024, help='Image height (Default is 1024)')
+    parser.add_argument('--width', type=int, default=1024, help='Image width (Default is 1024)')
+    parser.add_argument('--steps', type=int, default=4, help='Inference Steps')
```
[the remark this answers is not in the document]
filipstrand: I like your suggestion here that we should automatically set the steps based on the model choice if not specified by the user.

src/mflux/post_processing/image_util.py
```diff
@@ -19,11 +19,11 @@ def to_image(
             lora_paths: list[str],
             lora_scales: list[float],
             config: RuntimeConfig,
-    ) -> Image:
```
filipstrand, line 22: Just a bit curious about the name change here. Did you switch to `GeneratedImage` here because now (and also later when we include img2img feature) we will have more "image" concepts in the application and it is good to differentiate all of these?

src/mflux/controlnet/flux_controlnet.py
```diff
@@ -0,0 +1,321 @@
+from pathlib import Path
```
filipstrand, line 1: I would suggest to add an empty `__init__.py` file to the `controlnet` directory to mark it as a proper package.

src/mflux/weights/weight_handler.py
```diff
@@ -87,6 +88,39 @@ def load_transformer(root_path: Path | None = None, lora_path: str | None = None
                         "linear2": block["ff_context"]["net"][2]
                     }
         return weights, quantization_level
+    
+    @staticmethod
+    def load_controlnet_transformer(controlnet_id: Path | None = None) -> (dict, int):
+        controlnet_path = Path(snapshot_download(repo_id=controlnet_id,allow_patterns=["*.safetensors","config.json"]))
```
filipstrand, line 94: `snapshot_download` expects a `str` type for `repo_id` but `controlnet_id` is a `Path`. Have not tried this out yet, I guess it works regardless, but nice to fix that warning.

filipstrand:
> Driving by as an user/observer of this repo to help test and review this change.
> 
> * Left some suggestions in inline comments. Easy stuff.
> * Withholding nitpicks on style/linting - deferring to @filipstrand on aesthetic guidelines. @filipstrand may I suggest adding a `README` section on what commands to run to validate/process a file before PR, e.g. `ruff check` and alike.
> 
> I've been testing on `dev` model with `steps=16` and getting some reasonable results. I'm not expert eyed enough to spot what might be missing, which is to say - it's generating scenes consistent with the control image and I'm happy with the state of this PR. Thanks @Xuzzo for spearheading this work!

@anthonywu It is a great suggestion that we should add a style guide with some tooling like `ruff` to help us enforce that. Up to now, I have mostly relied on the built-in defaults of PyCharm to set the standard, but for a nicer development experience for all, we should use something that works well across other editors too). Will look into this a bit more soon.

Xuzzo:
@filipstrand @anthonywu thanks for the reviews! Regarding styling: I agree we should have some enforced guidelines. I would recommend using [pre-commit](https://pre-commit.com/) with the ruff and black hooks. Here is a pre-commit config I normally use
```
default_install_hook_types: [commit-msg, pre-commit]
default_stages: [commit, manual]
fail_fast: false
repos:
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.4.0
    hooks:
      - id: check-added-large-files
        args: ["--maxkb=15000"]
  - repo: local
    hooks:
      - id: ruff
        name: ruff
        entry: poetry run ruff check
        require_serial: true
        language: system
        types: [python]
      - id: black
        name: black
        entry: poetry run black
        require_serial: true
        language: system
        types: [python]
```

src/mflux/generate_controlnet.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Generate an image based on a prompt.')
+    parser.add_argument('--prompt', type=str, required=True, help='The textual description of the image to generate.')
+    parser.add_argument('--control-image-path', type=str, required=True, help='Local path of the image to use as input for controlnet.')
+    parser.add_argument('--output', type=str, default="image.png", help='The filename for the output image. Default is "image.png".')
+    parser.add_argument('--model', "-m", type=str, required=True, choices=["dev", "schnell"], help='The model to use ("schnell" or "dev").')
+    parser.add_argument('--seed', type=int, default=None, help='Entropy Seed (Default is time-based random-seed)')
+    parser.add_argument('--height', type=int, default=1024, help='Image height (Default is 1024)')
```
filipstrand, line 21: Not fully sure how this works myself yet, but does it make sense to specify the `height` and `width` of the generated image if we have a `control-image-path`, or should those dimensions always be the same as the controlnet input image?

filipstrand:
@Xuzzo Will do some more testing and playing around with this later today and just curious, do you know if this works with both `schnell` and `dev` or just the latter?

Xuzzo:
It was trained with dev, so performance with schnell is not guaranteed. Anyway, the tests I did were quite good with both

src/mflux/generate_controlnet.py
```diff
@@ -0,0 +1,65 @@
...
+    parser.add_argument('--width', type=int, default=1024, help='Image width (Default is 1024)')
+    parser.add_argument('--steps', type=int, default=4, help='Inference Steps')
+    parser.add_argument('--guidance', type=float, default=3.5, help='Guidance Scale (Default is 3.5)')
+    parser.add_argument('--controlnet-strength', type=float, default=0.7, help='Controls how strongly the control image influences the output image. A value of 0.0 means no influence. (Default is 0.7)')
```
filipstrand, line 25: Recently I also added the ability to save image-metadata (see the `_embed_metadata` method in `image.py`) I think we should save this and any other controlnet metadata (such as reference image name etc) there too.

filipstrand:
@Xuzzo Just wanted to say really great job with this one! I have tried it out on a few examples (first got some bad results, but turned out the strength was too high, then after tuning that a bit I got really good results.) and I think we can merge this after the comments are fixed and make a new release. Just had a few smaller comments above.

Also out of interest: I have not followed the controlnet space closely, but it sounded like there are more kinds to be supported? Would be interesting to hear your opinions on what some other candidates are here and if we should keep track of them in a table [like we do with the LoRA adapters](https://github.com/filipstrand/mflux?tab=readme-ov-file#supported-lora-formats-updated) for example.

filipstrand:
@anthonywu Thanks once again for your review efforts and for highlighting the UX with the default step sizes for the models. I remember reading that some other user also accidentally ran the dev model with 4 steps and got bad results. In the worst case, I guess users assume something is broken and move on...

Small things like good defaults makes the user experience so much nicer but this is so easy to overlook, especially if already familiar with the project, so thanks for pointing this out :)

src/mflux/post_processing/image_util.py
Xuzzo: Yes when adding the types I had PIL.Image.Image and the returned Image class and I thought it could generate confusion. We can revert back or find a better name than GeneratedImage (like AnnotatedImage or so) if you want to keep the Image return type

src/mflux/config/runtime_config.py
```diff
@@ -35,6 +35,13 @@ def precision(self) -> mx.Dtype:
     @property
     def num_train_steps(self) -> int:
         return self.model_config.num_train_steps
+    
+    @property
+    def controlnet_conditioning_scale(self) -> float:
+        if isinstance(self.config, ConfigControlnet):
+            return self.config.controlnet_strength
+        else:
+            return ValueError("Controlnet conditioning scale is only available for ConfigControlnet")
```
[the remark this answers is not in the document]
Xuzzo: thanks for the suggestion. Changed it to NotImplementedError

src/mflux/generate_controlnet.py
Xuzzo: The control image is stretched to the config height and width and a warning is returned if they dont match (see [here](https://github.com/filipstrand/mflux/pull/48/files#diff-a1f520a8130a2f4fcb56145cf5b29725d1828d0cf5e1116a53c95353944916c5R111-R113)) . This can be helpful e.g. if you want to upscale image quality

Xuzzo:
@filipstrand thanks! a control strength below 1 should more or less always give decent results. If not let me know! Regarding the other controlnet types: they all depend on what is used to extract the control features. For now the features are extracted with canny edges detector, which is a relatively simple technique (and does not require a model), but there are others (like depth mapping) that require some (simple) neural nets. I would like to bring [this model](https://huggingface.co/InstantX/FLUX.1-dev-Controlnet-Union) next, which has a bunch of controlnet types altogether. The standard way would be to accompany it with [this dependency](https://github.com/huggingface/controlnet_aux), but I am not sure how well it works on Macs and will need to do a few tests

filipstrand:
@Xuzzo I pushed some additional commits, mostly minor changes, moving some things around and some renaming. The biggest thing which I have wanted to do even before this PR was to separate out the model saving logic a bit from the main Flux file (since those are big enough already). With this addition of controlnet, I think it warrants that change even more so I went ahead and did it here. 

I'd like to merge this to main today, but won't have the time before work, so will have to do it later today. I also want to do a bit more manual testing to make sure everything works as expected (which reminds me of the need for some automatic tests we need to add in future). If any of my changes here seem weird or unintuitive, please comment and continue in this PR :)

filipstrand:
@Xuzzo Forgot to mention before, but I will upload some examples of the controlnet  to the readme and then publish this in a new release this weekend.

src/mflux/controlnet/transformer_controlnet.py
```diff
@@ -0,0 +1,96 @@
...
+            block_sample = controlnet_block(block_sample)
+            controlnet_block_samples = controlnet_block_samples + (block_sample,)
+
+        single_block_samples = ()
+        for block in self.single_transformer_blocks:
+            ctrlnet_hidden_states = block.forward(
+                hidden_states=ctrlnet_hidden_states,
+                text_embeddings=text_embeddings,
+                rotary_embeddings=image_rotary_emb
+            )
```
anthonywu, line 84: @Xuzzo @filipstrand 

my `ruff check` found a problem here:

```
src/mflux/controlnet/transformer_controlnet.py:80:31: F821 Undefined name `ctrlnet_hidden_states`
   |
78 |         for block in self.single_transformer_blocks:
79 |             ctrlnet_hidden_states = block.forward(
80 |                 hidden_states=ctrlnet_hidden_states,
   |                               ^^^^^^^^^^^^^^^^^^^^^ F821
81 |                 text_embeddings=text_embeddings,
82 |                 rotary_embeddings=image_rotary_emb
   |
```

`ctrlnet_hidden_states` is circularly defined on its first encounter at L81. I think it needs an initial value before the `for`-block

Xuzzo: ah oupsy. It should be

```suggestion
        for block in self.single_transformer_blocks:
            hidden_states = block.forward(
                hidden_states=hidden_states,
                text_embeddings=text_embeddings,
                rotary_embeddings=image_rotary_emb
            )
```
It does not give problems because the Controlnet in use does not have single blocks (only double).

filipstrand: Thanks, I have added this small fix to main now.

PR #55, opened 2024-09-16
dev ux: simplify import paths of frequently used items in scripts and demos

anthonywu:
I propose making common scripts/demos less verbose (i.e. developer onboarding), we can make the most commonly imported objects auto-imported at the top level of the package.

before

```py
from mflux.config.model_config import ModelConfig
from mflux.config.config import Config
from mflux.flux.flux import Flux1
```

after

```py
from mflux import Flux1, Config, ModelConfig
```

considerations:

- any package/module re-organizations in the future can be transparent to the user
- this is soft change, any pre-existing code that does the full path import will still work
- imports within the package implementation should still use full path imports following `explicit is better than implicit` value, as code at that level is meant for maintainers
- added a basic test that would exercise the `import`s via `generate.py`

-----

also taking opportunity to address the seemingly unnecessary (probably debug artifact from the past):

`os.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))`

is `src/mflux` and was being inserted as the first element of `sys.path`, which probably fixed an import path error in the past. I think this is a no-op removal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,8 +105,7 @@ mflux-generate --model dev --prompt "Luxury food photograph" --steps 25 --seed 2
 Or, with the correct python environment active, create and run a separate script like the following:
 
 ```python
-from mflux.flux.flux import Flux1
-from mflux.config.config import Config
+from mflux import Flux1, Config
 
 # Load the model
 flux = Flux1.from_alias(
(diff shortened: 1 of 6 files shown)


filipstrand:
I like it. Will rebase this on top of latest main make similar changes to the new controlnet file

PR #57, opened 2024-09-20
Introduce project makefile for setup and contributor operation

anthonywu:
This `Makefile` enables contributors to use consistent and recommended commands to achieve contributor guidelines that will be forthcoming, e.g. formatting/linting requirements as well as running a test suite.

The major change is officially recommending `uv venv` to drive the setup process. It's an opinion that I think is acceptable for a new project from August 2024.

PR #58, opened 2024-09-20
introduce standardized `ruff check` and `ruff format` config

anthonywu:
Following #57 which soft introduces `ruff` as linter/formatter, this PR defines the contributor configs for submitting changes.


pyproject.toml
```diff
@@ -46,3 +46,50 @@ mflux-generate-controlnet = "mflux.generate_controlnet:main"
 [tool.setuptools.packages.find]
 where = ["src"]
 include = ["mflux*"]
+
+[tool.ruff]
+line-length = 88         # same as 'black' default
```
anthonywu, line 51: `88` is somehow the community default from `black` formatter, which `ruff format` reasonably dupes.

Personally I did 80 or 90 or 120 depending on the project type (whether the viewers would experience the code on wide-screens would view it on remote terminals in split windows)

I think for mflux, most people would experience it on the Mac screens so up to 120 is reasonably IMO.

So lets start with 88, I want help preview what the diff would be going up to 220.

```sh
$ for len in $(seq 88 4 220); do rich "line length $len" --rule -w 40; ruff format --check --line-length $len --diff --respect-gitignore | wc -l; done
──────────── line length 88 ────────────
65 files would be reformatted, 20 files already formatted
    2962
──────────── line length 92 ────────────
65 files would be reformatted, 20 files already formatted
    2863
──────────── line length 96 ────────────
65 files would be reformatted, 20 files already formatted
    2661
─────────── line length 100 ────────────
65 files would be reformatted, 20 files already formatted
    2561
─────────── line length 104 ────────────
65 files would be reformatted, 20 files already formatted
    2429
─────────── line length 108 ────────────
65 files would be reformatted, 20 files already formatted
    2257
─────────── line length 112 ────────────
65 files would be reformatted, 20 files already formatted
    2183
─────────── line length 116 ────────────
65 files would be reformatted, 20 files already formatted
    2067
─────────── line length 120 ────────────
65 files would be reformatted, 20 files already formatted
    2020
─────────── line length 124 ────────────
65 files would be reformatted, 20 files already formatted
    1988
─────────── line length 128 ────────────
65 files would be reformatted, 20 files already formatted
    1978
─────────── line length 132 ────────────
65 files would be reformatted, 20 files already formatted
    1962
─────────── line length 136 ────────────
65 files would be reformatted, 20 files already formatted
    1951
─────────── line length 140 ────────────
65 files would be reformatted, 20 files already formatted
    1946
─────────── line length 144 ────────────
65 files would be reformatted, 20 files already formatted
    1940
─────────── line length 148 ────────────
65 files would be reformatted, 20 files already formatted
    1929
─────────── line length 152 ────────────
65 files would be reformatted, 20 files already formatted
    1925
─────────── line length 156 ────────────
65 files would be reformatted, 20 files already formatted
    1905
─────────── line length 160 ────────────
65 files would be reformatted, 20 files already formatted
    1919
─────────── line length 164 ────────────
65 files would be reformatted, 20 files already formatted
    1893
─────────── line length 168 ────────────
65 files would be reformatted, 20 files already formatted
    1882
─────────── line length 172 ────────────
65 files would be reformatted, 20 files already formatted
    1885
─────────── line length 176 ────────────
65 files would be reformatted, 20 files already formatted
    1857
─────────── line length 180 ────────────
65 files would be reformatted, 20 files already formatted
    1858
─────────── line length 184 ────────────
65 files would be reformatted, 20 files already formatted
    1858
─────────── line length 188 ────────────
65 files would be reformatted, 20 files already formatted
    1849
─────────── line length 192 ────────────
65 files would be reformatted, 20 files already formatted
    1849
─────────── line length 196 ────────────
65 files would be reformatted, 20 files already formatted
    1849
─────────── line length 200 ────────────
65 files would be reformatted, 20 files already formatted
    1848
─────────── line length 204 ────────────
65 files would be reformatted, 20 files already formatted
    1842
─────────── line length 208 ────────────
65 files would be reformatted, 20 files already formatted
    1837
─────────── line length 212 ────────────
65 files would be reformatted, 20 files already formatted
    1837
─────────── line length 216 ────────────
65 files would be reformatted, 20 files already formatted
    1838
─────────── line length 220 ────────────
65 files would be reformatted, 20 files already formatted
    1838
```

no matter what, 65 files would be reformatted in this range, what do you feel comfortable with @filipstrand.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+    # Ruff version.
+    rev: v0.6.6
+    hooks:
+      # Run the linter.
+      - id: ruff
+        types_or: [python, pyi]
+        args: [--fix]
+      # Run the formatter.
+      - id: ruff-format
+        types_or: [python, pyi]
```
anthonywu, line 12: from official guide: https://github.com/astral-sh/ruff-pre-commit?tab=readme-ov-file

if accepted, these would be added to the Makefile helpers:

1. `uv tool install pre-commit`
2. `pre-commit install` (adds the hooks to local .git directory)
3. add a helper `make check` for `pre-commit run --all-files` which is the same as `make lint` and `make format` in #57 but some ppl prefer to run these things separately 🤷🏼‍♂️ 
4. users can still bypass anytime with `git commit --no-verify` anything that violates `pre-commit`. Anyone can also just... not `install pre-commit`, so everything really is a soft suggestion until GitHub Actions is setup

pyproject.toml
filipstrand: I think 120 should work fine. This is purely a matter of personal taste, but there are some cases where exceeding this kind of limit makes sense IMO and makes code a bit easer to read, but mostly it is a good rule. First I was thinking of disabling the `line-length` check, but it is probably better to add a few exceptions where it makes sense. I'll go over the updates and make any necessary changes...

anthonywu: 120 is nice place. I think having the long def and call arg lists be forced to new lines is much more readable.

pyproject.toml
```diff
@@ -48,13 +48,27 @@ where = ["src"]
 include = ["mflux*"]
 
 [tool.ruff]
-line-length = 88         # same as 'black' default
+line-length = 120
 indent-width = 4
 target-version = "py310"
 respect-gitignore = true
+exclude = [
+    "src/mflux/generate.py",
```
filipstrand, line 56: There were some formatting results which became a bit more visually chaotic to me, mostly when there is a group of lines doing similar things, such as defining a nn layer or parser arguments etc. (as in `generate.py` for example), but where some of the lines exceed the length limit and the others don't and you end up with something like the following:

```python
    parser.add_argument("--prompt", type=str, required=True, help="The textual description of the image to generate.")
    parser.add_argument(
        "--output", type=str, default="image.png", help='The filename for the output image. Default is "image.png".'
    )
    parser.add_argument(
        "--model",
        "-m",
        type=str,
        required=True,
        choices=["dev", "schnell"],
        help='The model to use ("schnell" or "dev").',
    )
``` 

Personally I think the above looks worse than what was there before (even if you have to scroll a bit to see the full content): 

```python
    parser.add_argument("--prompt", type=str, required=True, help="The textual description of the image to generate.")
    parser.add_argument("--output", type=str, default="image.png", help="The filename for the output image. Default is \"image.png\".")
    parser.add_argument("--model", "-m", type=str, required=True, choices=["dev", "schnell"], help="The model to use (\"schnell\" or \"dev\").")
```

I ran the formatting on all files, then reverted the changes I was a bit unsure about and then excluded those files for future formatting, (uncommenting all excluded files here and re-running the formatting will show all those cases). To exclude whole files like this is probably not the best thing long term, but I think it is a start.

filipstrand: An alternative is to have all arguments be formatted like so 

```python
parser.add_argument(
        "--model",
        "-m",
        type=str,
        required=True,
        choices=["dev", "schnell"],
        help='The model to use ("schnell" or "dev").',
    )
```

regardless of how long the single line was. I think the mix of styles is the worst option.

.pre-commit-config.yaml
filipstrand: I added the application of the formatting as new commits on top of your branch here and now both formatting and linting should complete successfully. If you want, please feel free to add your proposed pre-commit additions in this branch, otherwise I am happy with the state of this PR and can merge it to main and we can take that separately, both works equally well for me :)

anthonywu:
LGTM, merge at your convenience.

pyproject.toml
anthonywu: you can flank those long parser lines with: https://docs.astral.sh/ruff/formatter/#format-suppression

```
# fmt: off
not_formatted=3
also_not_formatted=4
# fmt: on
```

or 

```
parser.add(... very long args)  # fmt: skip at 
```

filipstrand:
> LGTM, merge at your convenience. The `fmt: off` and `fmt: skip` magic comments may be what you need for the intentionally-long one liners without exempting whole files.

Great, thanks once again! I'll merge this one to main now, and will take a look at your other PR tomorrow! For the 0.3.0 release, I was planning to include the controlnet feature by Fabio and your additions here and post it on twitter as I've done with the previous releases. Just wanted to check with you first if it's OK that I mention you there too :)  https://x.com/anthonywu

anthonywu:
> check with you first if it's OK that I mention you there too 

fine by me 🤝

PR #59, opened 2024-09-20
option to save stepwise and composite images, add interruptibility

anthonywu:
This is my own original take on the objectives of #21 by user @madroidmaq, who wanted the stepwise images to be saved to disk. I commented in the PR that I'd like to see the objective completed.

In this PR, I am proposing:

1. make the generation flow more CLI `Ctrl-C` friendly during long generations, especially relevant with `dev` model and 10+ steps. This opens up the possibility that users can quit early if the early steps do not fit their expectations for any reason. (todo: a diff interrupt mechanism needs to be introduced for hypothetical GUI operations, TBD)
2. You can choose to output the stepwise images one by one as they become available, or wait for the end and a composite will be stitched together for you. The panel can help users evaluate diff time/quality tradeoffs in adjusting the steps parameter.
3. there is some duplication of structure and code in `src/mflux/flux/flux.py` vs `src/mflux/controlnet/flux_controlnet.py`, but in prior PRs I think @filipstrand you were okay with the (hopefully) temporary duplication, and now that we have a working baseline, we can use the tests to refactor forward (I can help, but it'll be a P2/P3 given things work well right now)

- [x] This is a PR that will need to be rebased on expected major formatting changes in other pending PRs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ respect-gitignore = true
 # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`) codes by default.
 # Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
 # McCabe complexity (`C901`) by default.
-select = ["E4", "E7", "E9", "F"]
+select = ["BLE", "E4", "E7", "E9", "F", "ICN", "LOG", "PERF", "W"]
 ignore = []
 
 # Allow fix for all enabled rules (when `--fix`) is provided.
--- a/src/mflux/controlnet/flux_controlnet.py
+++ b/src/mflux/controlnet/flux_controlnet.py
@@ -1,16 +1,16 @@
 import logging
-from typing import TYPE_CHECKING
-
 import mlx.core as mx
 from mlx import nn
+from pathlib import Path
 from tqdm import tqdm
-
+from typing import TYPE_CHECKING
 from mflux.config.config import ConfigControlnet
 from mflux.config.model_config import ModelConfig
 from mflux.config.runtime_config import RuntimeConfig
 from mflux.controlnet.controlnet_util import ControlnetUtil
 from mflux.controlnet.transformer_controlnet import TransformerControlnet
 from mflux.controlnet.weight_handler_controlnet import WeightHandlerControlnet
+from mflux.exceptions import StopImageGenerationException
 from mflux.models.text_encoder.clip_encoder.clip_encoder import CLIPEncoder
 from mflux.models.text_encoder.t5_encoder.t5_encoder import T5Encoder
 from mflux.models.transformer.transformer import Transformer
@@ -110,7 +110,9 @@ class Flux1Controlnet:
             output: str,
             controlnet_image_path: str,
             controlnet_save_canny: bool = False,
-            config: ConfigControlnet = ConfigControlnet()
+            config: ConfigControlnet = ConfigControlnet(),
+            stepwise_output_dir: Path = None,
+            stepwise_composite_only=False
     ) -> "GeneratedImage":  # fmt: off
         # Create a new runtime config based on the model type and input parameters
         config = RuntimeConfig(config, self.model_config)
@@ -139,34 +141,68 @@ class Flux1Controlnet:
         prompt_embeds = self.t5_text_encoder.forward(t5_tokens)
         pooled_prompt_embeds = self.clip_text_encoder.forward(clip_tokens)
 
+        step_wise_images = []
+        if stepwise_output_dir:
+            stepwise_output_dir.mkdir(parents=True, exist_ok=True)
         for t in time_steps:
-            # Compute controlnet samples
-            controlnet_block_samples, controlnet_single_block_samples = self.transformer_controlnet.forward(
-                t=t,
-                prompt_embeds=prompt_embeds,
-                pooled_prompt_embeds=pooled_prompt_embeds,
-                hidden_states=latents,
-                controlnet_cond=controlnet_cond,
-                config=config,
-            )
+            try:
+                # Compute controlnet samples
+                controlnet_block_samples, controlnet_single_block_samples = self.transformer_controlnet.forward(
+                    t=t,
+                    prompt_embeds=prompt_embeds,
+                    pooled_prompt_embeds=pooled_prompt_embeds,
+                    hidden_states=latents,
+                    controlnet_cond=controlnet_cond,
+                    config=config,
+                )
 
-            # 3.t Predict the noise
-            noise = self.transformer.predict(
-                t=t,
-                prompt_embeds=prompt_embeds,
-                pooled_prompt_embeds=pooled_prompt_embeds,
-                hidden_states=latents,
-                config=config,
-                controlnet_block_samples=controlnet_block_samples,
-                controlnet_single_block_samples=controlnet_single_block_samples,
-            )
+                # 3.t Predict the noise
+                noise = self.transformer.predict(
+                    t=t,
+                    prompt_embeds=prompt_embeds,
+                    pooled_prompt_embeds=pooled_prompt_embeds,
+                    hidden_states=latents,
+                    config=config,
+                    controlnet_block_samples=controlnet_block_samples,
+                    controlnet_single_block_samples=controlnet_single_block_samples,
+                )
 
-            # 4.t Take one denoise step
-            dt = config.sigmas[t + 1] - config.sigmas[t]
-            latents += noise * dt
+                # 4.t Take one denoise step
+                dt = config.sigmas[t + 1] - config.sigmas[t]
+                latents += noise * dt
 
-            # Evaluate to enable progress tracking
-            mx.eval(latents)
+                # Evaluate to enable progress tracking
+                mx.eval(latents)
+
+                if stepwise_output_dir:
+                    stepwise_decoded = self.vae.decode(
+                        Flux1Controlnet._unpack_latents(latents, config.height, config.width)
+                    )
+                    # performance todo: Pillow mostly uses CPU,
+                    # can try to improve image generation performance by offloading
+                    # stepwise image processing to the CPU via threading
+                    stepwise_img = ImageUtil.to_image(
+                        decoded_latents=stepwise_decoded,
+                        seed=seed,
+                        prompt=prompt,
+                        quantization=self.bits,
+                        generation_time=time_steps.format_dict["elapsed"],
+                        lora_paths=self.lora_paths,
+                        lora_scales=self.lora_scales,
+                        config=config,
+                    )
+                    step_wise_images.append(stepwise_img)
+                    if not stepwise_composite_only:
+                        stepwise_img.save(
+                            path=stepwise_output_dir / f"seed_{seed}_step{t+1}of{len(time_steps)}.png",
+                            export_json_metadata=False,
+                        )
+            except KeyboardInterrupt:  # noqa: PERF203
+                raise StopImageGenerationException(f"Stopping image generation at step {t+1}/{len(time_steps)}")
+            finally:
+                if step_wise_images and stepwise_output_dir:
+                    composite_img = ImageUtil.to_composite_image(step_wise_images)
+                    composite_img.save(stepwise_output_dir / f"seed_{seed}_composite.png")
 
         # 5. Decode the latent array and return the image
         latents = Flux1Controlnet._unpack_latents(latents, config.height, config.width)
--- a/src/mflux/exceptions.py
+++ b/src/mflux/exceptions.py
@@ -1,0 +1,18 @@
+class MFluxException(Exception):
+    """base class for all custom exceptions in mflux package."""
+
+
+class ImageSavingException(MFluxException):
+    """error ocurred while attempting to save image to storage."""
+
+
+class MetadataEmbedException(MFluxException):
+    """error ocurred while attempting to embed metadata in image"""
+
+
+class MFluxUserException(MFluxException):
+    """an exception raised by user behavior or intention."""
+
+
+class StopImageGenerationException(MFluxUserException):
+    """user has requested to stop a image generation in progress."""
--- a/src/mflux/exceptions.py
+++ b/src/mflux/exceptions.py
@@ -3,11 +3,11 @@ class MFluxException(Exception):
 
 
 class ImageSavingException(MFluxException):
-    """error ocurred while attempting to save image to storage."""
+    """error occurred while attempting to save image to storage."""
 
 
 class MetadataEmbedException(MFluxException):
-    """error ocurred while attempting to embed metadata in image"""
+    """error occurred while attempting to embed metadata in image"""
 
 
 class MFluxUserException(MFluxException):
--- a/src/mflux/flux/flux.py
+++ b/src/mflux/flux/flux.py
@@ -1,10 +1,12 @@
 import mlx.core as mx
+from pathlib import Path
 from mlx import nn
 from tqdm import tqdm
 
 from mflux.config.config import Config
 from mflux.config.model_config import ModelConfig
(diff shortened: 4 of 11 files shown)


src/mflux/flux/flux.py
```diff
@@ -85,22 +89,56 @@ def generate_image(self, seed: int, prompt: str, config: Config = Config()) -> G
...
+                    hidden_states=latents,
+                    config=config,
+                )
+
+                # 4.t Take one denoise step
+                dt = config.sigmas[t + 1] - config.sigmas[t]
+                latents += noise * dt
+
+                # Evaluate to enable progress tracking
+                mx.eval(latents)
```
anthonywu, line 111: all this is just indented

src/mflux/flux/flux.py
```diff
@@ -85,22 +89,56 @@ def generate_image(self, seed: int, prompt: str, config: Config = Config()) -> G
...
+                        lora_paths=self.lora_paths,
+                        lora_scales=self.lora_scales,
+                        config=config,
+                    )
+                    step_wise_images.append(stepwise_img)
+                    if not stepwise_composite_only:
+                        stepwise_img.save(
+                            path=stepwise_output_dir / f"seed_{seed}_step{t+1}of{len(time_steps)}.png",
+                            export_json_metadata=False
+                        )
```
anthonywu, line 135: option to skip the per-step image saves and only accept the composite

src/mflux/flux/flux.py
```diff
@@ -85,22 +89,56 @@ def generate_image(self, seed: int, prompt: str, config: Config = Config()) -> G
...
+            except KeyboardInterrupt:
+                raise StopImageGenerationException(f"Stopping image generation at step {t+1}/{len(time_steps)}")
+            finally:
+                if step_wise_images and stepwise_output_dir:
+                    composite_img = ImageUtil.to_composite_image(step_wise_images)
+                    composite_img.save(stepwise_output_dir / f"seed_{seed}_composite.png")
```
anthonywu, line 141: this is a decision (for now, under consideration) to generate the composite on every step, so that even if the user quits at step 8 of 16, they'd have a composite of the 8 steps.

- on one hand, this is convenient and probably best DX
- on the other hand, we generate the composite N times instead of once at the end
- on the other hand, if you choose to see stepwise images, you're already in a debugging mindset so maybe maximally leaning into the debugging experience is okay?

users can always opt out of these expensive mid-processing steps by not outputting stepwise at all

src/mflux/post_processing/image_util.py
```diff
@@ -39,6 +40,17 @@ def to_image(
...
+    def to_composite_image(generated_images: t.List[GeneratedImage]) -> Image:
+        # stitch horizontally
+        total_width = sum(gen_img.image.width for gen_img in generated_images)
+        max_height = max(gen_img.image.height for gen_img in generated_images)
+        composite_img = Image.new("RGB", (total_width, max_height))
+        current_x = 0
+        for index, gen_img in enumerate(generated_images):
+            composite_img.paste(gen_img.image, (current_x, 0))
+            current_x += gen_img.image.width
+        return composite_img
```
anthonywu, line 52: example, "blue bird", schnell 6 steps

image

src/mflux/generate.py
```diff
@@ -43,20 +44,24 @@ def main():
...
+            stepwise_output_dir=Path(args.stepwise_image_output_dir) if args.stepwise_image_output_dir else None
         )
-    )
 
-    # Save the image
-    image.save(path=args.output, export_json_metadata=args.metadata)
+        # Save the image
+        image.save(path=args.output, export_json_metadata=args.metadata)
+    except StopImageGenerationException as stop_exc:
+        print(stop_exc)
```
anthonywu, line 64: works as expected, we do have to wait for each step to finish, but much improve over the prior behavior where you have to smash Ctrl-C many times to hope for a catch

```sh
mflux-generate --model schnell --prompt "blue bird" --steps 6 --metadata --stepwise-image-output-dir /tmp/foo2

  0%|                                                                                                                                                | 0/6 [00:00<?, ?it/s]
  0%|                                                                                                                                                | 0/6 [00:23<?, ?it/s]
Stopping image generation at step 1/6
```

src/mflux/flux/flux.py
anthonywu, line 135: example, "blue bird", 6 steps

image

tests/test_readme_example.sh
```diff
@@ -1,7 +1,31 @@
...
+
+mflux-generate \
+    --prompt "blue bird, morning, spring" \
+    --model schnell \
+    --steps 2 \
+    --seed 24 \
+    --height 512 \
+    --width 512 \
+    --stepwise-image-output-dir /tmp/mflux-test \
+    --output /tmp/mflux-test/sf_blue_bird.png
```
anthonywu, line 20: this makes a blue bird image which is the control input image for the next step

image

tests/test_readme_example.sh
```diff
@@ -1,7 +1,31 @@
...
+mflux-generate-controlnet \
+    --prompt "yellow bird, afternoon, snowy mountain" \
+    --model schnell \
+    --control-image-path /tmp/mflux-test/sf_blue_bird.png \
+    --steps 2 \
+    --seed 24 \
+    --height 512 \
+    --width 512 \
+    --output /tmp/mflux-test/controlnet_sf_yellow_bird.png \
+    --stepwise-image-output-dir /tmp/mflux-test
```
anthonywu: the control net produces

image

note the images have consistent edges despite starting with diff seeds

anthonywu (commented):
added tests to cover the new stepwise generation

the files produced end up in `/tmp/mflux-test` and are:

```
luxury_food.png
seed_24_step1of2.png
seed_24_step2of2.png
seed_24_composite.png
sf_blue_bird.png
seed_42_step1of2.png
seed_42_step2of2.png
seed_42_composite.png
controlnet_sf_yellow_bird.png
```

anthonywu (commented):
I feel pretty good about the functional state of this PoC. interruptibility is very nice when I want to quit early and get my GPU cycles back ASAP, and stepwise has helped me make some decisions about inference step tradeoffs.

The duplication between the two modules does bother me, but that cleanup is going to be an ugly diff and should be done as a separate undertaking, and only after tests are put in to guarantee the refactoring will be a functional no-op.

src/mflux/exceptions.py
```diff
@@ -0,0 +1,10 @@
+class MFluxException(Exception):
+    """base class for all custom exceptions in mflux package."""
+
+
+class MFluxUserException(MFluxException):
+    """an exception raised by user behavior or intention."""
+
+
+class StopImageGenerationException(MFluxUserException):
+    """user has requested to stop a image generation in progress."""
```
anthonywu, line 10: defining custom exceptions specific to this module, and specific to scenarios can allow for fine-grained error and flow handling at the CLI/GUI level going forward

tests/test_readme_example.sh
```diff
@@ -1,7 +1,37 @@
...
+
 mflux-generate \
     --prompt "Luxury food photograph" \
     --model schnell \
     --steps 2 \
     --seed 2 \
-    --height 1024 \
-    --width 1024
+    --height 512 \
+    --width 512 \
```
anthonywu, line 13: reducing the test image size from 1024 to 512 dim should make the test times more tolerable.

src/mflux/post_processing/generated_image.py
```diff
@@ -66,7 +66,7 @@ def save(self, path: str, export_json_metadata: bool = False) -> None:
             # Embed metadata
             self._embed_metadata(file_path)
             log.info(f"Metadata embedded successfully at: {file_path}")
-        except Exception as e:
+        except Exception as e:  # noqa: BLE001
             log.error(f"Error saving image: {e}")
```
anthonywu, line 70: I see why this is `exception Exception` because a variety of fs/permission problems can come up during saving, but over time we should be able to recognize and catch `OSError` and other anticipatable problems.

filipstrand:
This looks interesting, and it’s good that this feature has come back into focus again. I’ll finish the last part of the ControlNet additions, then after that I'll take a look at this feature.

anthonywu:
a self-concern I have right now is how to properly name the stepwise output files, I would like to find a way to auto-name and auto-index the output files by prompt content, perhaps using some reliable NLP library to summarize the prompt in 5-6 words and use that summary + the unix timestamp of generation start time to always uniquely identify each set of images (stepwise, composite, final)

filipstrand:
@anthonywu Apologies for the late response, been a bit busy the last week, but will finally have some time to look at this again :)

src/mflux/exceptions.py
```diff
@@ -0,0 +1,18 @@
+class MFluxException(Exception):
+    """base class for all custom exceptions in mflux package."""
+
+
+class ImageSavingException(MFluxException):
+    """error ocurred while attempting to save image to storage."""
```
filipstrand, line 6: Small typo for `ocurred`

src/mflux/post_processing/image_util.py
filipstrand: Nice! I recently did a presentation of this project and wanted to illustrate the denoising steps as you have done here, but obviously had to go in and hack around a bit to get them, so definitely nice to have support for this feature. I can add this graphic to the readme too.

src/mflux/exceptions.py
anthonywu: Will fix, and I can suggest adding this to pre-commit hook: https://crates.io/crates/typos-cli

anthonywu:
> @anthonywu Apologies for the late response, been a bit busy the last week, but will finally have some time to look at this again :) 

There is plenty of work to follow-up, such as having a UI (Gradio, Jupyter notebook, etc) that show and cache the partially computed image from many variations, then the user can continue or stop any images based on early signs of goodness. I think this is especially relevant for local compute where images come back after many minutes, and only one image can reliably run at a time on the GPU.

src/mflux/flux/flux.py
```diff
@@ -86,22 +88,54 @@ def generate_image(self, seed: int, prompt: str, config: Config = Config()) -> G
...
+                        quantization=self.bits,
+                        generation_time=time_steps.format_dict["elapsed"],
+                        lora_paths=self.lora_paths,
+                        lora_scales=self.lora_scales,
+                        config=config,
+                    )
+                    step_wise_images.append(stepwise_img)
+                    if not stepwise_composite_only:
+                        stepwise_img.save(
+                            path=stepwise_output_dir / f"seed_{seed}_step{t+1}of{len(time_steps)}.png",
```
filipstrand, line 130: Got a warning here for path vs string. It works fine, but for type hinting  maybe the better thing is for the `save` method to support both `str` and `Path`

src/mflux/post_processing/image_util.py
```diff
@@ -47,6 +47,17 @@ def to_image(
             controlnet_strength=config.controlnet_strength if isinstance(config.config, ConfigControlnet) else None,
         )
 
+    def to_composite_image(generated_images: t.List[GeneratedImage]) -> Image:
```
filipstrand, line 50: Can be marked as static

src/mflux/exceptions.py
filipstrand: Yes this sounds like a good thing to have

src/mflux/controlnet/flux_controlnet.py
```diff
@@ -139,34 +141,68 @@ def generate_image(
...
+                latents += noise * dt
+
+                # Evaluate to enable progress tracking
+                mx.eval(latents)
+
+                if stepwise_output_dir:
+                    stepwise_decoded = self.vae.decode(
+                        Flux1Controlnet._unpack_latents(latents, config.height, config.width)
+                    )
+                    # performance todo: Pillow mostly uses CPU,
```
filipstrand, line 181: Could we remove the todos comments? A good reminder to have, but I sometimes find that the codebase ends up with a lot of them after some time, so I typically like put these elsewhere (unless it is something to be address very soon). Maybe we could have a section in the readme or elsewhere if there are obvious performance todos or other things that could be worked on in the future?

filipstrand:
@anthonywu I have tried running this and it looks very good. Thanks for the great job! 

I wanted to hear your opinion on the following: I kind of think of this feature as important and a good UX, but also as a smaller thing that should not be that visible in the main flow. I was thinking of ways to hide it a bit, and one solution could be to introduce a `StepwiseHandler` or similar that could handle the stepwise outputs. Since it could be configured at init time to be active/inactive based on the `stepwise_output_dir` args, we would not need to check if we should call it, but simply do `stepwise_handler.process_step(t, latents)` and it would internally know if it should produce the outputs or not. 

The end result could be something like this (have not had time to test it thoroughly, will try to do so later today), but I think it makes overall flow a bit more easy to glance at:

```python
    def generate_image(
        self,
        seed: int,
        prompt: str,
        config: Config = Config(),
        stepwise_output_dir: Path = None,
        stepwise_composite_only: bool = False,
    ) -> GeneratedImage:
        # Create a new runtime config based on the model type and input parameters
        config = RuntimeConfig(config, self.model_config)
        time_steps = tqdm(range(config.num_inference_steps))
        stepwise_handler = StepwiseHandler(  # <----- This is new
            self,
            config,
            seed,
            prompt,
            stepwise_output_dir,
            stepwise_composite_only,
            time_steps
        )

        # 1. Create the initial latents
        latents = mx.random.normal(
            shape=[1, (config.height // 16) * (config.width // 16), 64],
            key=mx.random.key(seed)
        )  # fmt: off

        # 2. Embed the prompt
        t5_tokens = self.t5_tokenizer.tokenize(prompt)
        clip_tokens = self.clip_tokenizer.tokenize(prompt)
        prompt_embeds = self.t5_text_encoder.forward(t5_tokens)
        pooled_prompt_embeds = self.clip_text_encoder.forward(clip_tokens)

        try:
            for t in time_steps:
                noise = self.transformer.predict(
                    t=t,
                    prompt_embeds=prompt_embeds,
                    pooled_prompt_embeds=pooled_prompt_embeds,
                    hidden_states=latents,
                    config=config,
                )

                # 4.t Take one denoise step
                dt = config.sigmas[t + 1] - config.sigmas[t]
                latents + noise * dt

                # Handle stepwise output if enabled
                stepwise_handler.process_step(t, latents)

                mx.eval(latents)

        except KeyboardInterrupt:
            stepwise_handler.handle_interruption()
            raise StopImageGenerationException(f"Stopping image generation at step {t + 1}/{len(time_steps)}")

        # 5. Decode the latent array and return the image
        latents = ArrayUtil.unpack_latents(latents, config.height, config.width)
        decoded = self.vae.decode(latents)
        return ImageUtil.to_image(
            decoded_latents=decoded,
            seed=seed,
            prompt=prompt,
            quantization=self.bits,
            generation_time=time_steps.format_dict["elapsed"],
            lora_paths=self.lora_paths,
            lora_scales=self.lora_scales,
            config=config,
        )
```

Any thoughts on this? It will add another class, so some complexity there I guess. Any other downsides you can think of? 
If you agree on this, I think we could merge this PR and then I could to this kind of refactoring soon thereafter.

pyproject.toml
```diff
@@ -57,7 +57,7 @@ respect-gitignore = true
 # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`) codes by default.
 # Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
 # McCabe complexity (`C901`) by default.
-select = ["E4", "E7", "E9", "F"]
+select = ["BLE", "E4", "E7", "E9", "F", "ICN", "LOG", "PERF", "W"]
```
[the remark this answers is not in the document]
filipstrand: Sounds good

anthonywu:
Slow reply due to travel. The extraction into stepwise handler pattern LGTM, I will check out #70 now.

PR #61, opened 2024-09-21
quality of life maintenance for ruff/pre-commit

anthonywu:
Follow-up to today's PR discussions in #58 etc

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,17 +12,17 @@ authors = [{ name = "Filip Strand", email = "strand.filip@gmail.com" }]
 maintainers = [{ name = "Filip Strand", email = "strand.filip@gmail.com" }]
 requires-python = ">=3.10"
 dependencies = [
-    "huggingface-hub>=0.24.5",
-    "mlx>=0.16.0",
-    "numpy>=2.0.1",
-    "opencv-python>=4.10.0",
-    "piexif>=1.1.3",
-    "pillow>=10.4.0",
-    "safetensors>=0.4.4",
-    "sentencepiece>=0.2.0",
-    "torch>=2.3.1",
-    "tqdm>=4.66.5",
-    "transformers>=4.44.0",
+    "huggingface-hub>=0.24.5,<1.0",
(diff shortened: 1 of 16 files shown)


anthonywu (commented):
in this PR I try to find a good tradeoff between whole-file format exclusion and keeping the good-long-lines in place

pyproject.toml
```diff
@@ -12,17 +12,17 @@ authors = [{ name = "Filip Strand", email = "strand.filip@gmail.com" }]
...
+    "mlx>=0.16.0,<1.0",
+    "numpy>=2.0.1,<3.0",
+    "opencv-python>=4.10.0,<5.0",
+    "piexif>=1.1.3,<2.0",
+    "pillow>=10.4.0,<11.0",
+    "safetensors>=0.4.4,<1.0",
+    "sentencepiece>=0.2.0,<1.0",
+    "torch>=2.3.1,<3.0",
+    "tqdm>=4.66.5,<5.0",
+    "transformers>=4.44.0,<5.0",
```
anthonywu, line 25: Good practice to constrain against unwanted major version upgrades of dependencies. Lessons learned from many years of solving issues during the first buggy days of a dependency's major release.

anthonywu (commented):
@filipstrand feel free to edit and merge at your convenience

pyproject.toml
filipstrand: Oh, nice! I recently had an issue that resulted from the recent numpy 2.0 transition, and this kind of thing would have caught that, so very good to just explicitly state this for every package.

filipstrand:
@anthonywu Thanks for these qol updates! This is better than what I did with ignoring whole files. I think this looks good and will merge as is.

pyproject.toml
anthonywu: The transition from Pillow 8 -> 9 -> 10 also burned me a few times. It's especially sensitive in the first days of a new major version when wheels aren't available for all platform/archs yet and somewhere will be a broken build due to inability to build from source.

PR #71, opened 2024-10-09
introduce pre-commit typos check

anthonywu:
@filipstrand I see you've already done some of this cleanup in #70, but here's the formal introduction to the repo.


src/mflux/models/transformer/transformer.py
```diff
@@ -5,7 +5,7 @@
 
 from mflux.config.model_config import ModelConfig
 from mflux.config.runtime_config import RuntimeConfig
-from mflux.models.transformer.ada_layer_norm_continous import (
+from mflux.models.transformer.ada_layer_norm_continuous import (
```
anthonywu, line 8: file is renamed without modifications, happy to find a low friction fix here

_typos.toml
```diff
@@ -0,0 +1,5 @@
+# this is the config file for typos-cli, part of "pre-commit run"
+# use "typos --dump-config" to disocver other configs to override
+
+[default]
+extend-ignore-identifiers-re = ["EmbedND", "embed_nd", "alle"]
```
anthonywu, line 5: `alle` the Italian word from README. IMO not worth trying to introduce non-`en` locale settings. Just exempt this.

filipstrand: Agree

filipstrand:
Nice tool to have!

PR #72, opened 2024-10-09
introduce ruff isort

anthonywu:
this is a disruptive diff that I wanted to isolate from prior PRs.

- Updated the `pyproject.toml` file to enable `ruff` linter class `I` for `isort` and proposed [starter configs](https://docs.astral.sh/ruff/settings/#lintisort)
- Applied `ruff check --fix` on all files as one-time remediation, `pre-commit` will flag all future violations

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,7 +62,7 @@ respect-gitignore = true
 # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`) codes by default.
 # Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
 # McCabe complexity (`C901`) by default.
-select = ["BLE", "E4", "E7", "E9", "F", "ICN", "LOG", "PERF", "W"]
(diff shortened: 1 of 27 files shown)


pyproject.toml
```diff
@@ -36,9 +36,7 @@ classifiers = [
 ]
 
 [project.optional-dependencies]
-dev = [
-    "pytest>=8.0.0,<9.0"
-]
+dev = ["pytest>=8.0.0,<9.0"]
```
anthonywu, line 39: I think we just have to sync on the ruff version and config file that is causing the auto-format drifts here. I'm going through `make check` helper on the CLI and ending up with this diff.

src/mflux/controlnet/flux_controlnet.py
```diff
@@ -183,7 +185,7 @@ def generate_image(
                 # Evaluate to enable progress tracking
                 mx.eval(latents)
 
-            except KeyboardInterrupt:
+            except KeyboardInterrupt:  # noqa: PERF203
```
anthonywu, line 188: needed this to pass `make check` on my local

filipstrand:
Looks good to me!

PR #73, opened 2024-10-10
design discussion: generators yield

anthonywu:
For now, just draft PR to discuss potential interface change with @filipstrand 

# Motivation

- I want to create CLI and UI interfaces that generate many images at once
- However, due to limited compute for target users, generating one image to finish takes too long when user wants to iterate over e.g. N seeds or N prompt variations without waiting for one image to finish at a time

# Proposed Change

I don't like to break existing interfaces, or introduce complexity, so this proposal is an option that does not break existing docs, but does introduce behavior developers need to know (when to use it like a iterator, when to use it like a normal function)

Future work this enables:

- a different CLI e.g. `mflux-generate-many --attempts 5` or `mflux-generate-many --seeds 1 2 3 4 5` or `mflux-generate-prompts --prompt-file <some file in some format>`
- for now, I hijack the existing `mflux-generate` to demo the proposed iterator behavior - I would isolate the "create-many" behavior into a new script in the final implementation.

### DIFF
--- a/src/mflux/flux/flux.py
+++ b/src/mflux/flux/flux.py
@@ -80,6 +80,7 @@ class Flux1:
         prompt: str,
         config: Config = Config(),
         stepwise_output_dir: Path = None,
+        stepwise_yield: bool = False,
     ) -> GeneratedImage:
         # Create a new runtime config based on the model type and input parameters
         config = RuntimeConfig(config, self.model_config)
@@ -121,10 +122,12 @@ class Flux1:
                 latents += noise * dt
 
                 # Handle stepwise output if enabled
-                stepwise_handler.process_step(t, latents)
+                stepwise_output = stepwise_handler.process_step(t, latents)
 
                 # Evaluate to enable progress tracking
                 mx.eval(latents)
+                if stepwise_yield:
+                    yield stepwise_output
 
             except KeyboardInterrupt:
                 stepwise_handler.handle_interruption()
--- a/src/mflux/generate.py
+++ b/src/mflux/generate.py
@@ -16,6 +16,7 @@ def main():
     parser.add_argument("--width", type=int, default=1024, help="Image width (Default is 1024)")
     parser.add_argument("--steps", type=int, default=None, help="Inference Steps")
     parser.add_argument('--stepwise-image-output-dir', type=str, default=None, help='[EXPERIMENTAL] Output dir to write step-wise images and their final composite image to. This feature may change in future versions.')
+    parser.add_argument('--resume-from-time-step', type=int, default=0, help='[EXPERIMENTAL] Resume from saved latents at this step.')
     parser.add_argument("--guidance", type=float, default=3.5, help="Guidance Scale (Default is 3.5)")
     parser.add_argument("--quantize",  "-q", type=int, choices=[4, 8], default=None, help="Quantize the model (4 or 8, Default is None)")
     parser.add_argument("--path", type=str, default=None, help="Local path for loading a model from disk")
@@ -42,11 +43,11 @@ def main():
     )
 
     try:
-        # Generate an image
-        image = flux.generate_image(
+        image_gen_steps = flux.generate_image(
             seed=int(time.time()) if args.seed is None else args.seed,
             prompt=args.prompt,
             stepwise_output_dir=Path(args.stepwise_image_output_dir) if args.stepwise_image_output_dir else None,
+            stepwise_yield=True,
             config=Config(
                 num_inference_steps=args.steps,
                 height=args.height,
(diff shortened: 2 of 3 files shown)


src/mflux/generate.py
```diff
@@ -42,24 +42,32 @@ def main():
     )
 
     try:
-        # Generate an image
-        image = flux.generate_image(
+        image_gen_steps = flux.generate_image(
             seed=int(time.time()) if args.seed is None else args.seed,
             prompt=args.prompt,
             stepwise_output_dir=Path(args.stepwise_image_output_dir) if args.stepwise_image_output_dir else None,
+            stepwise_yield=True,
```
anthonywu, line 49: this is the compromise I'm proposing to not break existing behavior

if some interface wants to use `generate_image` like a python generator, then they just request it with this flag, the usual `GeneratedImage` is captured in `StopIteration.value`

src/mflux/flux/flux.py
```diff
@@ -121,10 +122,12 @@ def generate_image(
                 latents += noise * dt
 
                 # Handle stepwise output if enabled
-                stepwise_handler.process_step(t, latents)
+                stepwise_output = stepwise_handler.process_step(t, latents)
 
                 # Evaluate to enable progress tracking
                 mx.eval(latents)
+                if stepwise_yield:
+                    yield stepwise_output
```
anthonywu, line 130: ## alternatives to this

I thought about extracting the entire `for t in time_steps` into a classmethod `iter_generate_latents`, so that the pre-existing `generate_image` would just for-loop it and then we have a no-op refactoring

Then, we expect the controllers (CLI, UI) that want stepwise yielding to itself re-implement the `# 2. Embed the prompt` and `# 5. Decode the latent array and return the image` sections, depending on how it wants to vary the prompt and seeds.

But this is actually more work for the CLI/UI vs. just passing in `stepwise_yield` and getting the desired behavior.

src/mflux/generate.py
```diff
@@ -42,24 +42,32 @@ def main():
...
                 height=args.height,
                 width=args.width,
                 guidance=args.guidance,
             ),
         )
-
-        # Save the image
+        while True:
+            stepwise_output = next(image_gen_steps)
+            print(f"Stepwise output: {stepwise_output}")
```
anthonywu, line 59: demonstrating what a CLI/UI could do with stepwise output

1. render step-wise images in series
2. if generating N images, prove that for all requests, step 0 is done, then for all requests, step 1 is performed, so that if you have a panel of N generations, each panel takes one turn, then we can create future hooks to cancel undesirable "directions" and save some compute

all of this is motivated by my use case of sending in hundreds of prompts each with a small iteration, but many of those were clearly going in a bad direction early on, and I could have saved the compute cycles by canceling them

src/mflux/generate.py
```diff
@@ -42,24 +42,32 @@ def main():
...
                 num_inference_steps=args.steps,
                 height=args.height,
                 width=args.width,
                 guidance=args.guidance,
             ),
         )
-
-        # Save the image
+        while True:
+            stepwise_output = next(image_gen_steps)
```
anthonywu, line 58: in a "create-many" situation, I would instantiate N python generators by calling `flux.generate_image` with diff args, then collect them into a list.

then something like:

```python
for image_count, image_gen in enumerate(all_image_generators, 1):
    # for one image request, generate one step
    stepwise_output = next(image_gen)
```

not implemented yet: a way to uniquely id each generation request, so the CLI/UI can "cancel next steps" when the generators are already queued up

filipstrand:
Finally got around to looking at this :) 

I definitely find your use cases here valid and I think the idea of returning a generator is a more general approach than to return the final image as `generate_image` currently does. It adds slightly more abstraction, but it think the uses cases here are valid. 

Please correct me if I am wrong, but yielding the `stepwise_output` in the time_step loop effectively returns that step to the caller and then some other piece of code would be responsible for iterating and actually doing something with with the output, like

```
while True:
     stepwise_output = next(image_gen_steps)
     print(f"Stepwise output: {stepwise_output}")
```

If so, I guess what we are doing here is really exposing each generation step to the "outside" to enable more flexibility. I feel this is natural and the caller would have more control over each generator to progress as they want it to (like in your example of having many parallel ones all executing one step at a time or similar).  

I like your default approach of not breaking any existing interface when proposing a feature, and maybe this example was only temporary,  but if we were to go this route, I feel it is a bit weird to have `generate_image()` returning *either* a `GeneratedImage` or the `stepwise_output` as a generator. Would it not be clearer to always yield the stepwise_output? Then it would be the responsibility of another piece of code to actually decide what to do with the step (like saving it, not saving it and only returning the final image etc).

filipstrand:
Just a short comment regarding generating multiple image at a time: 

Another way to do it would be a more low-level approach by creating several initial latent arrays and concatenating them and doing the matrix math on this concatenated array (I think the mlx-examples have support for this). 

I guess depending on what mlx does under the hood, it could be faster than doing it for one image at a time as we do now. However, I guess this approach would not allow us to cancel certain images halfway through the generation (since the math would be done on the full concatenated array). Also, I purposely avoided taking this approach when starting the project to keep things simple and not having to worry about concatenating/slicing arrays and whatever complications that might bring (maybe this is actually easy - e.g just concatenating each new latent instance along a new dimension - but I have not investigated this yet).

src/mflux/generate.py
filipstrand: Very valid use-cases

anthonywu:
We can hold on this significant operational change for now and re-engage after higher visibility feature releases.

Since posting the draft PR I have also been using some system level locks to run/queue multiple mflux programs at a time. Look for those proposals in coming weeks.

anthonywu:
Withdrawing this PR for now due to its staleness. I think the stepwise functionality and interruptibility can be improved but I will re-approach this at a later time.

PR #77, opened 2024-10-12
`mflux-generate` – support image-to-image

anthonywu:
Implemented with help from upstream alike projects as well as @Xuzzo's #16 - I have a working "img2img" implementation inlined with the standard generate workflow, where the "initial image" and "initial image strength" just provides initial image/latents and fast forwards to a step > 0 for the otherwise identical workflow.


src/mflux/config/config.py
```diff
@@ -14,13 +16,19 @@ def __init__(
         width: int = 1024,
         height: int = 1024,
         guidance: float = 4.0,
+        init_image: Path | None = None,
+        init_image_strength: float | None = None,
+        seed: float | None = None,
```
anthonywu, line 21: in prior PRs, @filipstrand you commented that at that time you preferred to keep separate scripts and interfaces, but in this case I think we can keep the if/else logic between txt2img and img2img readable.

In the future, if there's a is-a/has-s refactoring of the related generators, they can still end up in separate modules/classes, but for now I think the if/else inlined here works well.

src/mflux/config/runtime_config.py
```diff
@@ -1,15 +1,25 @@
...
+from mflux.post_processing.image_util import ImageUtil
+
+logger = logging.getLogger(__name__)
+
+VAE: t.TypeAlias = "mflux.models.vae.vae.VAE"  # noqa: F821
 
 
 class RuntimeConfig:
-    def __init__(self, config: Config | ConfigControlnet, model_config: ModelConfig):
+    def __init__(self, config: Config | ConfigControlnet, model_config: ModelConfig, vae: VAE):
```
anthonywu, line 18: the vae **instance** is needed here in the later step to encode the init image

src/mflux/config/runtime_config.py
```diff
@@ -19,6 +29,10 @@ def height(self) -> int:
     def width(self) -> int:
         return self.config.width
 
+    @property
+    def seed(self) -> int:
+        return self.config.seed
```
anthonywu, line 34: need the seed to generate the initial noise

src/mflux/config/runtime_config.py
```diff
@@ -35,6 +49,41 @@ def precision(self) -> mx.Dtype:
...
+        if self.config.init_image is None:
+            # text to image, always begin at time step 0
+            return 0
+        else:
+            # we skip to the time step as informed by the init_image_strength
+            # the higher the strength number, the more time steps we skip
+            strength = max(0.0, min(1.0, self.config.init_image_strength))
+            # if the strength is too small to even influence the image
+            # help the user round up so the init_image has influence at step 1
+            t = max(1, int(self.num_inference_steps * strength))
```
anthonywu, line 63: not sure I **should** do this, but it feels right as a user experience thing.

For example, if I have 8 steps and I set strength to 0.1 then the result would be `0.8` and get truncated to `0.0` in which case the user might be left wondering why the init image has no influence at all!

src/mflux/flux/flux.py
```diff
@@ -93,19 +97,13 @@ def generate_image(
...
-        )  # fmt: off
-
         # 2. Embed the prompt
         t5_tokens = self.t5_tokenizer.tokenize(prompt)
         clip_tokens = self.clip_tokenizer.tokenize(prompt)
         prompt_embeds = self.t5_text_encoder.forward(t5_tokens)
         pooled_prompt_embeds = self.clip_text_encoder.forward(clip_tokens)
 
-        for t in time_steps:
+        for gen_step, t in enumerate(time_steps, 1):
```
anthonywu, line 106: we now need to differentiate the time-step for noise and the generation step actually computed

src/mflux/generate.py
```diff
@@ -11,6 +11,7 @@ def main():
     parser.add_model_arguments()
     parser.add_lora_arguments()
     parser.add_image_generator_arguments()
+    parser.add_image_to_image_arguments(required=False)
```
anthonywu, line 14: demoing a future where parser elements can be selectively required based on the controller

src/mflux/post_processing/image_util.py
```diff
@@ -97,8 +96,21 @@ def to_array(image: PIL.Image.Image) -> mx.array:
         return array
 
     @staticmethod
-    def load_image(path: str) -> Image.Image:
-        return Image.open(path)
+    def load_image(path: str) -> PIL.Image.Image:
```
anthonywu, line 99: consolidated the 3 diff `PIL` imports here to be same style

src/mflux/controlnet/flux_controlnet.py
```diff
@@ -131,7 +131,7 @@ def generate_image(
 
         # Embed the controlnet reference image
         control_image = ImageUtil.load_image(controlnet_image_path)
-        control_image = ControlnetUtil.scale_image(config.height, config.width, control_image)
+        control_image = ImageUtil.scale_to_dimensions(control_image, config.height, config.width)
```
anthonywu, line 134: todo here: we should give users some options/warnings here

1. perhaps introduce a `--strict` mode where we do not implicitly do things for them, if they pass in diff dimensions for in/output we error out and let the user resolve
2. introduce a `height=auto` and `width=auto` to add some auto smart cropping/padding for the users, there are various python libraries that find subjects in a photo and smartly crops for their focus. Out of scope for now, but I would choose one of those based on library quality and recency of releases.

src/mflux/post_processing/stepwise_handler.py
```diff
@@ -45,11 +51,10 @@ def process_step(self, step: int, latents: mx.array):
             self.step_wise_images.append(stepwise_img)
 
             stepwise_img.save(
-                path=self.output_dir / f"seed_{self.seed}_step{step + 1}of{len(self.time_steps)}.png",
+                path=self.output_dir / f"seed_{self.seed}_step{gen_step}of{len(self.time_steps)}.png",
                 export_json_metadata=False,
             )
+            self.save_composite()
```
anthonywu, line 57: this was subtlety lost in #70 refactoring, the result was that successful workflow did not create a composite but interrupted workflows had one

this does introduce progressive composite on every-step - I think in this workflow, `PIL.Image.save` is not the performance hog relative to running the prediction, and having the composites stream to storage is quite nice.

anthonywu:
# compare against fal.ai

this is a init image I borrowed from https://huggingface.co/docs/diffusers/en/using-diffusers/img2img

image

> mflux-generate --{height,width}=512 --model dev --init-image ~/Downloads/astro-init.png --init-image-strength 0.1 --steps 40 --prompt "Astronaut in a jungle, cold color palette, muted colors, detailed, 8k" --guidance 3.5 --seed 2961630356

generates this locally

image

-----

altering the strength to `0.2` holding all else constant:

image

-----

altering the strength to default value `0.4` holding all else constant:

image

-----

and using 0.8 - we get almost the identical image (subtly different, stare)

image

src/mflux/config/config.py
filipstrand: Yes that is right. Initially I wanted to keep things as separated as possible, but over time as we have added more features, I also realise that a lot of them can also combine in interesting ways, and it might not be so good to separate every feature (i.e have separate classes for `FluxImg2Img` etc). Obviously also good for avoiding too much code duplication.

src/mflux/config/runtime_config.py
filipstrand: To me, it feels a bit unintuitive that the Config knows about one of the nn modules such as the `vae`. I think it might be better to bring back the `LatentCreator` or similar that I had before 

```python
        # 1. Create the initial latents
        latents = LatentCreator.create(
            seed=seed,
            vae=self.vae,
            config=config,
        )
```

and let it handle the creation of the latents. I think it is fine that this method depends on the `vae` 

```python
class LatentCreator:
    @staticmethod
    def create(
        seed: int,
        vae: nn.Module,
        config: RuntimeConfig,
    ) -> (mx.array, mx.array):
        noise = mx.random.normal(
            shape=[1, (config.height // 16) * (config.width // 16), 64],
            key=mx.random.key(seed)
        )  # fmt: off

        if config.init_image is None:
            return noise
        else:
            image = ImageUtil.load_image(config.init_image).convert("RGB")
            encode = vae.encode(ImageUtil.to_array(ImageUtil.scale_to_dimensions(image, config.width, config.height)))
            latents = ArrayUtil.pack_latents(encode, config.width, config.height)
            sigmas_for_init_image_strength = config.sigmas[config.init_time_step]
            latents_adjusted = latents * (1.0 - sigmas_for_init_image_strength)
            noise_adjusted = noise * sigmas_for_init_image_strength
            return latents_adjusted + noise_adjusted
```

anthonywu: Perhaps this can be a follow-up refactoring. I was eyeing the commonalities between Flux/Controlnet implementations, and was trying to hold off on this proposed scope:

- introduce a base class `FluxBase`, find the copy/paste commons and move here
- `Flux1` and `Flux1Controlnet` becomes a sub-class of `FluxBase`

and relevant to this discussion

`FluxBase` should have a `@classmethod` `create(seed, config)` + use its own `self.vae` instantiated earlier in the flow.

This accomplishes a few things:

- acknowledge the latent creation is tightly coupled with the generator
- and also does not proliferate another util file or name/organize

src/mflux/config/config.py
anthonywu: the permutations of how to use the generators can be realized by adding more special-purpose CLI entry points that use the same argparser configs.

I agree that continually adding the Nth flag to a single CLI is unsustainable UX-wise.

src/mflux/config/runtime_config.py
filipstrand: Yes we could do this in a followup PR. I'll try to merge this soon and then look into that a bit. I also plan on adding a readme section for img2img and a test case that covers it

anthonywu: As long as the interface isn't public facing (and this refactoring won't be), internal changes should be merged between releases. I like keeping well-scoped PRs so it's easy to review and iterate on.

src/mflux/ui/cli/parsers.py
```diff
@@ -27,6 +27,10 @@ def add_image_generator_arguments(self) -> None:
         self.add_argument("--seed", type=int, default=None, help="Entropy Seed (Default is time-based random-seed)")
         self._add_image_generator_common_arguments()
 
+    def add_image_to_image_arguments(self, required=False) -> None:
+        self.add_argument("--init-image", type=Path, required=required, help="Local path to init image")
```
anthonywu, line 31: @filipstrand I realized this should be `--init-image-path` and everywhere else `init_image_path` to be consistent with `--controlnet-image-path`.

I have a big PR in a follow-up change that addresses this: https://github.com/anthonywu/mflux/pull/2 - this PR is based on `image-to-image` checkpoint. When this PR77 is merged, I'll be able to stage a clean diff PR in your repo here.

See https://github.com/anthonywu/mflux/pull/2 for its change scope for now! It's a big diff.

PR #81, opened 2024-10-19
support metadata files as CLI arg supplier

anthonywu:
Oct 19 - this is a preview of a big change that I think we'll discuss for several days. It probably should be in the (n+1)th release. I expect we'll go back and forth a few times here because the decisions here influence how other subsequent features are done.

-----

# Change

1. Add a new CLI option to consume the run configs (but not the model configs) from `.json` files. This allows for very long commands with many path args to be easily re-usable for subsequent runs. You can imagine this being useful for anyone experimenting with _changing one thing_ and seeing how that single variable affects the outcome.
1. Promote the utility of the existing metadata upgrade files to serve as the input for run configs. The configs are simple enough to be handcrafted, but the metadata output is the automatically compatible one we will offer.
1. There is some metadata schema change here, but it was not consumed anywhere by mflux so I would not consider this compatibility breaking for SemVar purposes.
1. See more inline comments for todos and design concerns.

# UX Demo

What the change allows are commands like:

1. `mflux-generate --model dev --metadata --config-from-metadata <file>` - re-generate from yours or other's prior run configs. You can use these as references for checking into team repos, or to use as the control config in a test suite where the before/after of a image gen change can from the same config file defined in a repo.
1. `mflux-generate --model schnell --metadata --config-from-metadata <file from dev run>` - re-do the image gen, but using a diff model than the one that saved the config
1. `mflux-generate --model dev --metadata --config-from-metadata <file> --guidance 5.0` - re-genenerate, but change 1 variable
1. `mflux-generate --model dev --metadata --config-from-metadata <file> --quantize 4` - re-generate, but with the lighter quantized model
1. put it in a for-loop somewhere (shell or python), and each iteration you change one thing. This is where my output namer proposal in #80 is relevant, i.e. I don't want to rename things as I iterate through large batches, I would opt into a smart auto-namer in that scenario.

PR #82, opened 2024-10-20
tests and fixes for image to image

anthonywu:
This is a necessary follow-up to #77 before @filipstrand can make the next pypi release.

- Remove VAE from config class, add a **Flux1** method to create initial latents
- Following your format for the first 3 tests, add a test to cover #77 
- Adjust the test harness just enough to accommodate the new variant, TIL image-to-image does not work well for rectangular dimensions
- rename "init_image" to "init_image_path" everywhere, to be consistent with `controlnet_image_path`
- I will rebase #81 on top of this work when merged

# Tests

- `tests/test_generate_image.py` PASSING for me
- ~~test_generate_image_controlnet.py FAIL for me - I am out of time trying to debug this today, but wanted to post this PR to let you know I'm following up on fixing after #77~~
- `test_generate_image_controlnet.py` PASSING for me


src/mflux/controlnet/controlnet_util.py
```diff
@@ -17,6 +17,13 @@ def preprocess_canny(img: PIL.Image) -> PIL.Image:
         image_to_canny = np.concatenate([image_to_canny, image_to_canny, image_to_canny], axis=2)
         return PIL.Image.fromarray(image_to_canny)
 
+    @staticmethod
+    def scale_image(height: int, width: int, img: PIL.Image) -> PIL.Image:
+        if height != img.height or width != img.width:
+            log.warning(f"Control image has different dimensions than the model. Resizing to {width}x{height}")
+            img = img.resize((width, height), PIL.Image.LANCZOS)
+        return img
+
```
anthonywu, line 26: Attempted to consolidate this to one implementation in #77 , but turns out there's a subtle bug introduced regarding resizing. Rolling back for now to make `tests/test_generate_image_controlnet.py` pass again for upcoming release. We'll worry about refactoring/merging implementations later.

src/mflux/controlnet/flux_controlnet.py
```diff
@@ -131,7 +131,7 @@ def generate_image(
 
         # Embed the controlnet reference image
         control_image = ImageUtil.load_image(controlnet_image_path)
-        control_image = ImageUtil.scale_to_dimensions(control_image, config.height, config.width)
+        control_image = ControlnetUtil.scale_image(config.height, config.width, control_image)
```
anthonywu, line 134: `ImageUtil.scale_to_dimensions` does not produce the same outcome

the rollback is towards `ControlnetUtil.scale_image`

anthonywu (commented):
Fixed `tests/test_generate_image_controlnet.py`!

src/mflux/controlnet/flux_controlnet.py
filipstrand: Forgot to rerun the tests myself before merging #77, but at least the tests caught the error.

src/mflux/controlnet/controlnet_util.py
filipstrand: Thanks for noticing, forgot to run the tests myself before/after merge to main. Ideally we would have something like a Github Action workflow to run the tests automatically.  Just checked this quickly and it looks like Github have support for Apple silicon runners, and that they are free? https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

Do you have any experience with these? If it is not too much work and if the machines can run the model (thinking limited memory could be an issue) and it is free, then it would be a good thing to have.

src/mflux/config/runtime_config.py
```diff
@@ -6,20 +6,17 @@
...
-from mflux.post_processing.image_util import ImageUtil
 
 logger = logging.getLogger(__name__)
 
 VAE: t.TypeAlias = "mflux.models.vae.vae.VAE"  # noqa: F821
 
 
 class RuntimeConfig:
-    def __init__(self, config: Config | ConfigControlnet, model_config: ModelConfig, vae: VAE):
+    def __init__(self, config: Config | ConfigControlnet, model_config):
```
filipstrand, line 16: Great, thanks for moving the vae out from here. As you wrote, the config classes can probably be refactored a bit for improved clarity.

src/mflux/controlnet/controlnet_util.py
anthonywu: > The new runner executes Actions workflows with a 3 vCPU, 7 GB RAM, and 14 GB of storage VM

A obvious next step for PRs. I think the 7GB RAM may be a challenge for generation time, but here's what I think we can do soon:

1. to run in the 7GB RAM env, create q=4 test resources that runs before the q=8 tests, if the q=4 test cases fails don't bother running q=8, there are some 
1. turn up the CI job timeout to max, as this likely exceeds typical CI timeout
1. use more `unittest.mock.patch` to mock out the actual step-wise iteration generation steps, this will allow for full coverage of the controller code surrounding the image generation, will be valid tests for refactoring efforts. I expect most of the image gen tests fail because ultimately the args passed into the inner most steps are mutated during breakages.

PR #84, opened 2024-10-25
support HF models - freepik as first example

anthonywu:
### Change

I nominate that we support compatible and trending 3rd party models based on:

1. curation for quality and reputation
2. compatibility

The first model to meet this criteria for me is: https://huggingface.co/Freepik/flux.1-lite-8B-alpha - it's a distilled version of flux.1 dev, read more at the link.

the claim:

> Flux.1 Lite, an 8B parameter transformer model distilled from the FLUX.1-dev model. This version uses 7 GB less RAM and runs 23% faster while maintaining the same precision (bfloat16) as the original model.

### Tradeoffs

the model configs are hard-coded for now, to hypothetically let anyone choose any huggingface-hosted model is possible but requires some config refactoring, but also invites anyone to attempt an incompatible model and creating Issues tracker noise

For now, I think this can make some users happy (those with less RAM) while mflux can be seen as inclusive of third party releases.

### DIFF
--- a/src/mflux/config/model_config.py
+++ b/src/mflux/config/model_config.py
@@ -2,8 +2,12 @@ from enum import Enum
 
 
 class ModelConfig(Enum):
+    # ==== black-forest-labs official models =====
     FLUX1_DEV = ("black-forest-labs/FLUX.1-dev", "dev", 1000, 512)
     FLUX1_SCHNELL = ("black-forest-labs/FLUX.1-schnell", "schnell", 1000, 256)
+    # ==== third party compatible models - best effort compatibility =====
+    # https://huggingface.co/Freepik/flux.1-lite-8B-alpha distilled from dev
+    FLUX1_FREEPIK_LITE_8B_ALPHA = ("Freepik/flux.1-lite-8B-alpha", "freepik-lite-8b-alpha", 1000, 512)
 
     def __init__(
         self,
--- a/src/mflux/ui/defaults.py
+++ b/src/mflux/ui/defaults.py
@@ -2,9 +2,10 @@ CONTROLNET_STRENGTH = 0.4
 GUIDANCE_SCALE = 3.5
 HEIGHT, WIDTH = 1024, 1024
 INIT_IMAGE_STRENGTH = 0.4  # for image-to-image init_image
-MODEL_CHOICES = ["dev", "schnell"]
+MODEL_CHOICES = ["dev", "schnell", "freepik-lite-8b-alpha"]
 MODEL_INFERENCE_STEPS = {
     "dev": 14,
     "schnell": 4,
+    "freepik-lite-8b-alpha": 22,
 }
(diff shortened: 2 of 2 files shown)


src/mflux/ui/defaults.py
```diff
@@ -2,9 +2,10 @@
 GUIDANCE_SCALE = 3.5
 HEIGHT, WIDTH = 1024, 1024
 INIT_IMAGE_STRENGTH = 0.4  # for image-to-image init_image
-MODEL_CHOICES = ["dev", "schnell"]
+MODEL_CHOICES = ["dev", "schnell", "freepik-lite-8b-alpha"]
 MODEL_INFERENCE_STEPS = {
     "dev": 14,
     "schnell": 4,
+    "freepik-lite-8b-alpha": 22,
```
anthonywu, line 9: from their readme:

> Flux.1 Lite is ready to unleash your creativity! For the best results, we strongly recommend using a guidance_scale of 3.5 and setting n_steps between 22 and 30.

I feel conflicted about having this value higher than dev, will play around to see if we can live with 10-14 range and give users faster generations by default.

src/mflux/config/model_config.py
```diff
@@ -2,8 +2,12 @@
 
 
 class ModelConfig(Enum):
+    # ==== black-forest-labs official models =====
     FLUX1_DEV = ("black-forest-labs/FLUX.1-dev", "dev", 1000, 512)
     FLUX1_SCHNELL = ("black-forest-labs/FLUX.1-schnell", "schnell", 1000, 256)
+    # ==== third party compatible models - best effort compatibility =====
+    # https://huggingface.co/Freepik/flux.1-lite-8B-alpha distilled from dev
+    FLUX1_FREEPIK_LITE_8B_ALPHA = ("Freepik/flux.1-lite-8B-alpha", "freepik-lite-8b-alpha", 1000, 512)
```
anthonywu, line 10: considered naming this `FLUX1_FREEPIK_LITE_8B` and pointing to the beta/final releases as the model matures. However, in python zen we should be more explicit than not..., so I prefer to still let users know exactly what they have to ask for and receive.

anthonywu:
not able to exactly match the `diffusers` output for now, I suspect I'm missing a few key parameters.

using the example in the model card: https://huggingface.co/Freepik/flux.1-lite-8B-alpha#text-to-image


`mflux-generate -m freepik-lite-8b-alpha --prompt "A close-up image of a green alien with fluorescent skin in the middle of a dark purple forest" --seed 11 --steps 28 --guidance 3.5`

creates

image

while running the diffusers example as is (but changing device to `mps`) creates

output

a head scratcher for now, I think this may be worth solving to see where mflux might differ from diffusers on default implementations.

anthonywu:
closing for now, not interesting enough to prioritize. Issue #101 is much more interesting and can use some of the work already done here.

PR #87, opened 2024-10-29
Fix img2img for non-square resolutions

filipstrand:
Simple height <---> width swap error. 

Before
output
After
reference_dev_image_to_image_result


src/mflux/controlnet/flux_controlnet.py
```diff
@@ -139,7 +139,7 @@ def generate_image(
         controlnet_cond = ImageUtil.to_array(control_image)
         controlnet_cond = self.vae.encode(controlnet_cond)
         controlnet_cond = (controlnet_cond / self.vae.scaling_factor) + self.vae.shift_factor
-        controlnet_cond = ArrayUtil.pack_latents(controlnet_cond, config.height, config.width)
```
filipstrand, line 142: Could also imagine having dedicated classes/types for height/width but I think this will do for now

src/mflux/config/config.py
```diff
@@ -37,5 +37,5 @@ def __init__(
         guidance: float = 4.0,
         controlnet_strength: float = 1.0,
     ):
-        super().__init__(num_inference_steps, width, height, guidance)
+        super().__init__(num_inference_steps=num_inference_steps, width=width, height=height, guidance=guidance)
```
filipstrand, line 40: Could also imagine having dedicated classes/types for height/width but I think naming the args every time height and width are passed in is good for now. 

Tests with non-square aspect ratio should hopefully catch errors (will add such a test for img2img)

tests/test_generate_image.py
```diff
@@ -46,13 +46,13 @@ def test_image_generation_dev_lora(self):
     def test_image_generation_dev_image_to_image(self):
         ImageGeneratorTestHelper.assert_matches_reference_image(
             reference_image_path="reference_dev_image_to_image_result.png",
-            init_image_path="reference_dev_image_to_image_init.png",
-            init_image_strength=0.126,  # fill 1/8 steps
+            init_image_path="reference_dev_lora.png",
```
filipstrand, line 49: Just picked on our existing images as the init

tests/test_generate_image.py
```diff
@@ -46,13 +46,13 @@ def test_image_generation_dev_lora(self):
...
+            init_image_strength=0.4,
             output_image_path=TestImageGenerator.OUTPUT_IMAGE_FILENAME,
             model_config=ModelConfig.FLUX1_DEV,
             steps=8,
-            seed=42,
-            height=768,
+            seed=44,
+            height=341,
             width=768,
-            prompt="astronauts in a jungle",
```
filipstrand, line 57: For safety, before swapping out the test_reference image, I check that this implementation produced the image as this test asserted for in the case of a squared aspect ratio. 

Decided not to keep this test since I think the non-square one covers more ground and we reduce the time it takes to run the suite by not having both

filipstrand:
Will merge this and do a `0.4.1` release

PR #97, opened 2024-11-16
Add GitHub Action workflow for Build & Test

filipstrand:
This won't be feasible unless we can store the weights on disk somehow...

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -17,8 +17,9 @@ jobs:
 
       - name: Configure additional swap space
         run: |
-          sudo diskutil erasevolume HFS+ "swap" $(sudo hdiutil attach -nomount ram://8388608) # Creates 4GB swap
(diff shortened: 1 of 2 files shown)


anthonywu:
Perhaps explore the combination of git submodule + git sparse checkout of the weights from HF repo? Would need to store the HF token with the Action runner.

filipstrand:
Do you think that would remove the need to re-download the weights every time the runner starts? 

My thinking is that since we test both dev + schnell + controlnet we will have to download roughly 34 + 34 + 3.5 GB  according to [the size of the weights](https://github.com/filipstrand/mflux?tab=readme-ov-file#-size-comparisons-for-quantized-models). Maybe if the tests were rewritten for 4bit (loading them from local storage somehow), we could get that number down quite a bit, but they are still large. (I also remember there were some issues loading LoRA weights, when the model was loaded from the 4bit quantized weights on disk, but don't remember exactly now...)

My current feeling is that testing the full image generation might not be worth it at the moment...

filipstrand:
Closing this for now

Issue #109, opened 2024-12-27
Training on Non-Sequential Block Layers


azrahello:
Reading a bit here and there, I understood that training on non-sequential blocks might be useful based on certain needs. So, I wanted to do a small test, but I couldn’t find a way to train on a series of non-contiguous block layers. I tried inserting the “range” of blocks, “start” and “end” in series, on separate numbered layers, and even when the block series was noted in the JSON file, in the adapter file uploaded to Hugging Face, I saw that the trained layers were only the last ones in the series listed in the JSON file. By tinkering with the code and modifying some classes, I managed to get what I wanted. The files I modified are:
```
@dataclass
class BlockRange:
    start: Optional[int] = None
    end: Optional[int] = None
    indices: Optional[List[int]] = None

    def get_blocks(self) -> List[int]:
        if self.indices:
            return self.indices
        if self.start is not None and self.end is not None:
            return list(range(self.start, self.end - 1))
```
This is from the training_spec.py file. Part of this:
```
def _construct_layers(
    block_spec: TransformerBlocks | SingleTransformerBlocks,
    blocks: list[JointTransformerBlock] | list[SingleTransformerBlock],
    block_prefix: str,
) -> dict:
    block_indices = block_spec.block_range.get_blocks()  # Uses the method with ()
    lora_layers = {}
    for idx in block_indices:
        if idx >= len(blocks):
            raise IndexError(f"Index {idx} out of bounds of the block list.")

        block = blocks[idx]
        for layer_type in block_spec.layer_types:
            original_layer = LoRALayers._get_nested_attr(block, layer_type)
            is_list = isinstance(original_layer, list)

            lora_layer = LoRALinear.from_linear(
                linear=original_layer[0] if is_list else original_layer,
                r=block_spec.lora_rank,
            )
            layer_path = f"{block_prefix}.{idx}.{layer_type}"

            lora_layers[layer_path] = [lora_layer] if is_list else lora_layer

    return lora_layers
```
This is from the lora_layers.py file.
The LoRA layers created on Hugging Face show the trained layers, and in the JSON file, this appears: 
```
"lora_layers": {
        "single_transformer_blocks": {
            "block_range": {
                "indices": [
                    0,
                    1,
                    7,
                    19,
                    20
                ]
            },
            "layer_types": [
                "proj_out"
            ],
            "lora_rank": 128
        }
    },
```

filipstrand:
[...]

PR #110, opened 2025-01-03
Verbose graphic and layered block indices

azrahello:
It's the first time I'm doing this operation and I'm afraid of breaking something! I've tried to be as non-intrusive as possible, especially since I didn't really have a clear understanding of what I was doing. I find the graphic particularly comfortable, but of course, it's just an idea


src/mflux/dreambooth/lora_layers/lora_layers.py
```diff
@@ -64,18 +64,20 @@ def from_spec(flux: "Flux1", training_spec: TrainingSpec) -> "LoRALayers":
 
             return LoRALayers(weights=weights)
 
+    @staticmethod
```
filipstrand, line 67: Accidentally added an extra `@staticmethod` here that is not needed

src/mflux/dreambooth/lora_layers/lora_layers.py
```diff
@@ -64,18 +64,20 @@ def from_spec(flux: "Flux1", training_spec: TrainingSpec) -> "LoRALayers":
...
     @staticmethod
     def _construct_layers(
         block_spec: TransformerBlocks | SingleTransformerBlocks,
         blocks: list[JointTransformerBlock] | list[SingleTransformerBlock],
         block_prefix: str,
     ) -> dict:
-        start = block_spec.block_range.start
-        end = block_spec.block_range.end
+        block_indices = block_spec.block_range.get_blocks()  # Usa il metodo con ()
```
filipstrand, line 74: If the code is mostly self explanatory, as I think in this case, then we don't need the comments. 

If there is a need to have a comment, we always try to have them in English for the benefit of everyone.

src/mflux/dreambooth/state/training_spec.py
```diff
@@ -54,10 +54,19 @@ class StatisticsSpec:
     state_path: str | None = None
 
 
+# indices and Optional
```
filipstrand, line 57: This comment can also be removed

src/mflux/dreambooth/state/training_spec.py
```diff
@@ -54,10 +54,19 @@ class StatisticsSpec:
...
+    start: Optional[int] = None
+    end: Optional[int] = None
+    indices: Optional[List[int]] = None
+
+    def get_blocks(self) -> List[int]:
+        if self.indices:
+            return self.indices
+        if self.start is not None and self.end is not None:
+            return list(range(self.start, self.end - 1))
+        raise ValueError("Devono essere forniti 'start' e 'end' o 'indices'.")
```
filipstrand, line 69: The error message here should be in English

src/mflux/dreambooth/statistics/plotter.py
```diff
@@ -1,12 +1,17 @@
...
+from datetime import timedelta
+
 import matplotlib.pyplot as plt
 
 from mflux.dreambooth.state.training_spec import TrainingSpec
 from mflux.dreambooth.state.training_state import TrainingState
 
 
 class Plotter:
+    start_time = time.time()  # Class variable to track time
```
filipstrand, line 11: I would not start a timer in this way.  The plotter class should only be concerned with actual plotting of graphs and the various methods that would need to know the `start_time` would probably need to have this passed in as an argument. 

Also, doing it in this way would start a timer when the class is loaded, and it could lead to timing that is not what we want (actual training time which is what I would guess you intend here...). Also, with the elapsed time, it is a bit more complicated since we can stop and resume training at any time (as we talked about here https://github.com/filipstrand/mflux/issues/108), and this is something that should be considered when plotting the training time.

filipstrand:
Great first try contributing and a great initiative to try your own solution the your original feature request. 
Appreciate your transparency with this being your first time etc., it makes it easier to know how much effort to put into code-review etc :)   

Some general notes: 

- As a good practice, it is nice to outline a bit in the PR what the solution is, what assumptions you have made etc.  In this case, since you already have opened the issue #109 some can be picked up from there, but still nice to have it in this PR. For example, a screenshot of how the updated plot looks like would be really great to have.

- Some comments and error messages were in non-english which should be updated.

- Maybe a bit more advanced, especially if you have not used git much before, but there is a nice ability in git to "squash commits" - this means that even though you might have made several back-and-forth changes during development, when you are happy with your efforts and want to present it for review, you can essentially present your final solution as one single change - you can think of this as only presenting your polished/final work without revealing the history. When taking this approach it would be fine to have your own commits in non-english and cryptic (I do so myself during development) because they would be hidden at the time of presenting the PR.  

Great job, and feel free to address  any of changes that you feel comfortable with (I can help with git squash thing later)  and I'll have another look :)

src/mflux/dreambooth/statistics/plotter.py
filipstrand: For clarity: Since handling the elapsed time part properly might be slightly bigger thing, I think it can be done in a separate PR.

azrahello: You’re absolutely right, and I fully admit that my approach to handling time might seem a bit primitive. I didn’t want to make too many changes as I was constantly worried about breaking something. :P I opted for a method that kept everything neatly contained within the task execution.

azrahello:
First of all, thank you! I truly value everything that comes out of this space. Exploring MFlux has helped me deepen my understanding of my field of research (I work in photography, and my background in computer science is stuck 20 years in the past). I’ve only revisited this environment about a year ago as part of my work on a final project. 

What I did by modifying the code was to create visual reference points for a more immediate evaluative analysis compared to opening a file. I tried to include details that would help me identify key differences between one training session and another, and the graph makes this much easier. Additionally, when quickly skimming through the plot files, I liked the idea of a kind of time-lapse effect, so I removed the empty space on the right to make the visualization more central, ending it right at the edge of the graph (aesthetic-wise, I find it works better with the legend).

I wanted to implement an easing-like approach, but my mathematical skills aren’t strong enough to find a proper method, so I improvised and found a solution to adjust the flow of the code. I added an extra 0.01% margin to prevent the graph from hitting the exact right edge. This tweak also gives me a better sense of training progress without relying on other metrics. Lastly, I included the tracking of maximum and minimum loss peaks, which I find particularly helpful
Screenshot 2025-01-21 alle 9 26 51 PM

filipstrand:
@azrahello I took the liberty of rebasing your branch ontop of the latest mflux main commit. Since this requires a force push on your version of main, I don't think I have the access to do so, so I opened up this PR https://github.com/filipstrand/mflux/pull/121 with your changes (you are still author of the main commit there)

filipstrand:
Now I changed so that we don't show that much empty space on the y-axis 

Screenshot 2025-01-27 at 23 49 39

filipstrand:
@azrahello I think this looks good, and I have tried it with both the old rages and the new `indices` input. I'll go ahead and merge https://github.com/filipstrand/mflux/pull/121 to main. Thanks for the help!

PR #114, opened 2025-01-12
Flowedit

filipstrand:
On the same topic of editing methods [inversion techniques](https://github.com/filipstrand/mflux/pull/111), this is an attempt to implement [FlowEdit](https://github.com/fallenshock/FlowEdit). This technique does not invert the image, but only uses regular the forward pass (with two calls to the transformer per step), and with some clever blending of noise from the source and target image. 

Screenshot 2025-01-12 at 12 04 36

Screenshot 2025-01-12 at 19 27 02

### DIFF
--- a/src/mflux/config/runtime_config.py
+++ b/src/mflux/config/runtime_config.py
@@ -27,6 +27,10 @@ class RuntimeConfig:
     def guidance(self) -> float:
         return self.config.guidance
 
+    @guidance.setter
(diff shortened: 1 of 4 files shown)


filipstrand:
@anthonywu Probably not, my guess is that Kontext will work a lot better in practice (same thing with the inversion PRs). I'll close these since there has not been any work done for a long time.

Issue #115, opened 2025-01-14
Is there a way to keep model running instead of loading again each time I want to generate an image?


Baitur5:
Currently if i want to generate an image, every time i gotta load flux model again, is there a way to keep model running?

anthonywu:
I think this dupes #112

filipstrand:
Yes this is effectively what #112 also asks for. Good to note as this might be something worth supporting natively as more people ask for it... 

@Baitur5 I assume you are using the command line, but if you are using Python it is as easy as wrapping the `generate_image` code in a for loop (in this example I loop 10 times producing 10 images, but only load the model once):

```python
from mflux import Flux1, Config

# Load the model
flux = Flux1.from_alias(
   alias="schnell",       # "schnell" or "dev"
   quantize=8,            # 4 or 8
)

# Generate an image
for i in range(10):
    image = flux.generate_image(
    seed=i,
    prompt="Luxury food photograph",
    config=Config(
        num_inference_steps=2,  # "schnell" works well with 2-4 steps, "dev" works well with 20-25 steps
        height=1024,
        width=1024,
    )
    )

    image.save(path="image.png")
```

anthonywu:
I refactored the argparser components so we could add more interfaces like this without duplicating the arg logic. I propose creating new entry points that are specifically for batch modes, would allow for the batch flags to not clutter up the single gen interface.

filipstrand:
That sounds like a good idea!

filipstrand:
This feature is now in main and will be included in the upcoming v.0.6.0 release thanks to Anthony.

Issue #112, opened 2025-01-09
Feature/Enhancement request: txt2image batch generation


notapreppie:
I wasn't able to find whether this functionality is already present, but my Google-Fu may be weak.

Would it be possible to implement batch generation with one or two command-line arguments? Basically just repeats all of the parameters except the seed, which would be incremented or randomized (or an option to choose which).

clivinn-shla81092:
Hello, we tried to solve the issue.

This is what we did:

Added batch generation functionality to the CLI parser and main generation script. This allows users to generate multiple images with the same parameters, incrementing or randomizing the seed for each generation.

You can review changes in this commit: https://github.com/clivinn-shla81092/filipstrand-mflux-112/commit/9ea72002f9aa954bc2d9cb0354163a9337eb215c.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

gmorain:
Hi, IMHO it does not really make sense to include the batch generation inside MFlux since a simple loop over the CLI execution achieves exactly the same thing (in bash, Python or whatever language able to execute system commands). No my decision of course as I am just a regular user of MFlux, but opening the door to this feature would be opening the pandora box: why stop at the seed when you can loop on steps too? Or any other parameter?

notapreppie:
> Hi, IMHO it does not really make sense to include the batch generation inside MFlux since a simple loop over the CLI execution achieves exactly the same thing (in bash, Python or whatever language able to execute system commands). No my decision of course as I am just a regular user of MFlux, but opening the door to this feature would be opening the pandora box: why stop at the seed when you can loop on steps too? Or any other parameter?

So, [slippery-slope informal logical fallacy](https://www.logical-fallacy.com/articles/slippery-slope/)?
"Why stop?" Well, the dev decides to stop when they decide to stop. It's not like they have any obligation to any of us. They're doing this because they want to and they can stop when they want or add/not-add features however they want.

[...]

PR #118, opened 2025-01-22
support multiple seeds batch generation

anthonywu:
This partially addresses the requests in #112 and #115.

Iterating on `seed` values while keeping all other values constant is the easiest of all the variable parameters. I think this satisfies a common user intent where they just want to see many variations of the same prompt/configs without reloading the model files.

In this PR, I attempt to:

- keep the `--seed` arg as is, but allow it to support n > 1 args.
- introduce a `--auto-seeds` arg, where you can allow the program to generate N seeds, giving up control of how the seeds are generated (it's just `random` library underneath)
- when seed count > 1, _smartly_ manipulate the `--output` name to embed each output file with the seed value so each file in the batch is named differently

# Examples

```sh
mflux-generate --model schnell --steps 4 \
  --prompt "a male lion protects his lion cub at Pride Rock" \
  --seed 1 2 3
```

produces `image_seed_1.png` `image_seed_2.png` `image_seed_3.png`

```sh
mflux-generate --model schnell --steps 4 \
  --prompt "a male lion protects his lion cub at Pride Rock" \
  --auto-seeds 5
```

produces something like: `image_seed_1561024.png` `image_seed_4756892.png` `image_seed_5976159.png` `image_seed_6479222.png` `image_seed_9434512.png`

# Future

It's possible to keep the model files loaded and iterate over N values of `width`, `height`, `steps`, `guidance`, and `init_image_*` values, but supporting all of those possible combinations can make the library code very complicated, so we may have to stop here for the official CLIs.

In my personal work, I have custom Python scripts that iterate over the configs without reloading model files, so we should just provide cookbook recipes for more complex examples, or defer to a separate CLI to be developed later in-library or as a third party tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,9 +155,11 @@ mflux-generate --model dev --prompt "Luxury food photograph" --steps 25 --seed 2
 
 - **`--model`** or **`-m`** (required, `str`): Model to use for generation (`"schnell"` or `"dev"`).
 
-- **`--output`** (optional, `str`, default: `"image.png"`): Output image filename.
+- **`--output`** (optional, `str`, default: `"image.png"`): Output image filename. If `--seed` `--auto-seeds` establishes N > 1 seed values, the "stem" of the output file name will automatically append `_seed_{value}`.
 
-- **`--seed`** (optional, `int`, default: `None`): Seed for random number generation. Default is time-based.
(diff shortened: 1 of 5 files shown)


filipstrand:
This looks great!

Issue #80, opened 2024-10-18
proposal: automatic output file naming


anthonywu:
Currently the pattern for automatic image naming is `image.*` and the subsequent images are smartly indexed as `image-N.*` etc.

However, I think we can make some improvements:

1. always add the dimensions to the image name like `image-512x512.png`
2. automatically summarize the prompt into a file name - so if you had a prompt about a dog playing in a park, the filename can be summarized by some tool as `dog-play-in-park.*`. The tool can be something traditional/fast like [`nltk`](https://www.nltk.org) or some other modern embedding/ranking tool for finding the most relevant keywords.
3. allow users to opt-in to automatic placeholders such as
  - `{iso_date} ISO date `yyyy-mm-dd`
  - `{unix_timestamp}` via `date +%s` or python str fmt
  - `{seed}`
  - other placeholders that represent the various params: `guidance`, `quantize`, etc

not trying to scope creep on what we support - just enough that every file name can be reasonably expected to be unique (seed and prompt summary, at the minimum)

We can also make a API for a `OutputFileNamer` - we'll provide reasonable defaults, and any users of the library can inject their own customization as needed.

azrahello:
I would like to ask if it is possible to embed the prompt as EXIF data in a PNG image, so that when the image is imported into photographic software, the prompt can appear as a description.

filipstrand:
@azrahello We do export this information with the created png file since a while back (even if you do not include the -metadata flag). For example, for an image  `image.png`  generated by mflux, if you run 

`exiftool image.png`

then you will see this kind of information (including the prompt): 

```
File Type                       : PNG
File Type Extension             : png
MIME Type                       : image/png
Image Width                     : 1024
Image Height                    : 1024
Bit Depth                       : 8
Color Type                      : RGB
Compression                     : Deflate/Inflate
Filter                          : Adaptive
Interlace                       : Noninterlaced
Exif Byte Order                 : Big-endian (Motorola, MM)
Warning                         : Invalid EXIF text encoding for UserComment
User Comment                    : {'mflux_version': '0.2.1', 'model': 'schnell', 'seed': '1728244816', 'steps': '6', 'guidance': 'None', 'precision': 'mlx.core.bfloat16', 'quantization': 'None', 'generation_time': '610.10 seconds', 'lora_paths': 'None', 'lora_scales': 'None', 'prompt': 'blue bird', 'controlnet_image': 'None', 'controlnet_strength': 'None'}
[...]

PR #120, opened 2025-01-22
Batch Image Renamer tool as an isolated uv run script 

anthonywu:
Implements #80 following user suggestion in #118, and in relation to Issues #112 and #115.

See README update for details.

Basically this is a script that reads the EXIF metadata and renames image files, and is executed using `uv run` to isolate its requirements from the main `mflux` project.

This type of functionality makes sense to be standalone from the python library, and as a reference implementation because users have many choice on keyword extraction models and specific requirements on how much complexity they want to introduce to their file name templates.

With this implementation, users have a usable tool as-is, but also have a starting point to implement their own customizations.

Versioning: with 0.6.0 in `main` not published yet - I think this can be part of 0.6.0.

### DIFF
--- a/src/mflux/post_processing/image_util.py
+++ b/src/mflux/post_processing/image_util.py
@@ -154,7 +154,7 @@ class ImageUtil:
     def _embed_metadata(metadata: dict, path: str) -> None:
         try:
             # Convert metadata dictionary to a string
-            metadata_str = str(metadata)
+            metadata_str = json.dumps(metadata)
 
             # Convert the string to bytes (using UTF-8 encoding)
             user_comment_bytes = metadata_str.encode("utf-8")
--- a/tools/rename_images.py
+++ b/tools/rename_images.py
@@ -1,0 +1,131 @@
+# /// script
+# requires-python = ">=3.10"
+# dependencies = [
+#   "Pillow",
+#   "keybert",
+# ]
+# ///
+
+import ast
+import json
+from pathlib import Path
+
+from keybert import KeyBERT
+from PIL import Image, UnidentifiedImageError
+
+PNG_SIGNATURE = b"\x89PNG\r\n\x1a\n"
+EXIF_USER_COMMENT_KEY = 37510
+
+
+class UnsupportedMetadata(Exception):
+    pass
+
+
+def parse_metadata_from_image(image: Image, insecure=False) -> dict:
+    try:
+        exif_data = image._getexif()
+    except AttributeError:
+        # when exif data is not available for some image type/format
+        return None
+
+    metadata = None
+    if exif_data:
+        try:
+            for tag, value in exif_data.items():
+                if tag == EXIF_USER_COMMENT_KEY:
+                    metadata = json.loads(value.decode())
+        except KeyError:
+            metadata = None
+        except json.decoder.JSONDecodeError:
+            if not insecure:
+                raise UnsupportedMetadata(
+                    "The metadata is likely stored as a str literal of a Python dict before mflux 0.6.0. "
+                    "This tool cannot guarantee safety of running eval(...) on the value. "
+                    "However, you can bypass this caution by passing the --insecure flag. "
+                    f"Metadata: {value.decode()}"
+                )
+            try:
+                # try to parse the str(dict) data
+                obj = ast.literal_eval(value.decode())
+                if isinstance(obj, dict):
+                    metadata = obj
+                else:
+                    raise UnsupportedMetadata(
+                        "The metadata is not a dict output recognized by this tool. " f"Metadata: {value.decode()}"
+                    )
+            except (ValueError, SyntaxError):
+                raise UnsupportedMetadata("The metadata is not parseable by this tool. " f"Metadata: {value.decode()}")
+
+    return metadata
+
+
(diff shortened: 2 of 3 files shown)


src/mflux/post_processing/image_util.py
```diff
@@ -154,7 +154,7 @@ def save_image(
     def _embed_metadata(metadata: dict, path: str) -> None:
         try:
             # Convert metadata dictionary to a string
-            metadata_str = str(metadata)
```
anthonywu, line 157: @filipstrand not sure why this wasn't a `json.dumps` from the beginning, because json can be safely parsed but a `str(dict)` needs to be evaluated.

In the script below, I parse the prior output but only if users supply `--insecure` for `ast.literal_eval` to be used to read the `str(metadata)` data

tools/rename_images.py
```diff
@@ -0,0 +1,115 @@
+# /// script
+# requires-python = ">=3.10"
+# dependencies = [
+#   "Pillow",
+#   "keybert",
+# ]
+# ///
```
anthonywu, line 7: https://docs.astral.sh/uv/guides/scripts/ reads this

tools/rename_images.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+    parser.add_argument(
+        "--yes", action="store_true", default=False, help="Allow renaming without interactive confirmation."
+    )
+    parser.add_argument(
+        "--insecure",
+        action="store_true",
+        default=False,
+        help="At your own risk, allow insecure parsing of literal Python values saved by mflux versions < 0.6.0",
+    )
```
anthonywu, line 73: read the usage here

src/mflux/post_processing/image_util.py
filipstrand: Nice that you caught this!

tools/rename_images.py
filipstrand: Just tried this out now for the first time, really convenient for small isolated scripts!

filipstrand:
Looks good. Definitely the right choice to have this separated from the main project. 

As this was a user suggested feature it makes some sense to include it here but, of course, the tradeoff is always what to include and what to skip... 

As AI/LLM tool gets smarter and easer to work with over time, my hope is that more users will be able to build tools themselves (sometimes completely automatically), even with very little code experience and that we would not need to provide many such pre-defined tools in the main repo. But to help enable this, a set of good examples displaying the API is probably very valuable to have as a starting point (for example, as you provided here). 

On this topic, do you think it would be valuable to even have a standalone script like `tools/generate_image.py` like [the one we have in the readme](https://github.com/filipstrand/mflux?tab=readme-ov-file#%EF%B8%8F-generating-an-image), also in the tools folder?, For example, like: 

```python
# /// script
# requires-python = ">=3.10"
# dependencies = [
#   "mflux",
# ]
# ///
from pathlib import Path

from mflux import Flux1, Config

# Load the model
flux = Flux1.from_alias(
    alias="dev",           # "schnell" or "dev"
    quantize=8,            # 4 or 8
)

# Generate an image
image = flux.generate_image(
    seed=2,
    prompt="Luxury food photograph",
    stepwise_output_dir=Path("~/Desktop/imgs"),
    config=Config(
        num_inference_steps=20,  # "schnell" works well with 2-4 steps, "dev" works well with 20-25 steps
        height=1024,
        width=1024,
        guidance=4.0,
    )
)

image.save(path="~/Desktop/image.png")
```

or would it be redundant?

anthonywu:
👍🏼 to moving the demo scripts into `tools/` folder as runnable (and therefore testable) snippets, but there some readability loss if we remove it from the README inline.

I propose a reasonable effort:

1. inline the easiest examples (1 or 2) in the README.md, and mention they also exist in the tools folder. These examples should rarely change, therefore zero maintenance for the owners.
2. Keep a suite of example scripts in the repo, and list them bullet-point-wise in the README. Create a light weight process for accepting more example scripts.  Solicit more examples from the community via PRs. Accept the PRs that properly use the library and has clean implementation.
3. Add README section that directs users to have agency to request modifications from AI assistants. Tactic: snap to an energing practice such as [llm.txt](https://github.com/AnswerDotAI/llms-txt) so that AI assistants can ingest the repo documentation as a plain text input in its context window.

filipstrand:
> 👍🏼 to moving the demo scripts into `tools/` folder as runnable (and therefore testable) snippets, but there some readability loss if we remove it from the README inline.
> 
> I propose a reasonable effort:
> 
> 1. inline the easiest examples (1 or 2) in the README.md, and mention they also exist in the tools folder. These examples should rarely change, therefore zero maintenance for the owners.
> 2. Keep a suite of example scripts in the repo, and list them bullet-point-wise in the README. Create a light weight process for accepting more example scripts.  Solicit more examples from the community via PRs. Accept the PRs that properly use the library and has clean implementation.
> 3. Add README section that directs users to have agency to request modifications from AI assistants. Tactic: snap to an energing practice such as [llm.txt](https://github.com/AnswerDotAI/llms-txt) so that AI assistants can ingest the repo documentation as a plain text input in its context window.

Agree with all of these suggestions!

PR #124, opened 2025-02-09
Refactor and add callback mechanism

filipstrand:
This PR does the following : 

- Refactors the init setup for `Flux` and `Flux1Controlnet` into a separate class `FluxInitializer`. Enables more reuse and moves some logic to separate file which makes the length of the Flux classes more manageable.  (Have been thinking of unifying the standard and controlnet variants of flux, right now it still feels best to keep them separated, but with shared methods where it makes sense. 

- Deletes `ConfigControlnet` and simply include the single additional parameter `controlnet_strength` in the `Config`

- Implements a more general callback mechanism - my idea was that some things we do feels a bit "outside" the main flow, and can be abstracted slightly (at the expense of some new indirections).  E.g instead of calling the stepwise handler in place, we now have 3 points (before-loop, in-loop, interruption) where we simply call any subscriber with the relevant information - currently, two classes are refactored to fit this pattern: `StepwiseHandler` and `CannyImageSaver`, but more could be added in the future. This introduces another layer of abstraction, but I think this is fine as it has nothing to do with the "main flow" and also has the benefit of cleaning up the interfaces for the `generate_image` methods


filipstrand:
Yes exactly, that was my idea too. And it should be pretty easy to update the callback interfaces if other information are needed in the future (e.g if someone finds a good use for having the `latents` in the `BeforeLoopCallback` etc).

PR #131, opened 2025-02-28
Memory usage optimizations

ssakar:
The default MLX cache size causes disk swapping, even with quantization and 36 GB of free memory (MacBook Pro with 48GB RAM). Here is a similar issue [mlx-example issue](https://github.com/ml-explore/mlx-examples/issues/724#issuecomment-2079315281). 

So this change limits the MLX cache size (default: 1 GB). I also noticed that the mlx-example for flux uses less memory because it frees up model components early on. Inspired by that I have implemented a callback handler, which frees up the text encoders before the flux loop and the transformer after the loop. 

With the panda example (schnell model, q8) and low-ram set, I get around 21.44 GB of MLX peak memory (+1 GB cache). Without the option I see a peak of 33 GB. With low-ram enabled, but not deleting the encoders, I get around 26.63 GB of MLX peak memory. Omitting the deleting of the transformer yields 21.44 GB, so no difference, but with larger image sizes it might be noticeable. All these runs took around 1:09 to 1:23 min, so I don't see a big performance hit. 

Example command:

mflux-generate \
    --model schnell \
    --steps 2 \
    --seed 2 **--low-ram** \
    --quantize 8 \
    --height 1920 \
    --width 1024 \
    --prompt "Tranquil pond in a bamboo forest at dawn, the sun is barely starting to peak over the horizon, panda practices Tai Chi near the edge of the pond, atmospheric perspective through the mist of morning dew, sunbeams, its movements are graceful and fluid — creating a sense of harmony and balance, the pond’s calm waters reflecting the scene, inviting a sense of meditation and connection with nature, style of Howard Terpning and Jessica Rossier"

Peak memory summary:

- low-ram: 22.44 GB (incl. mlx cache)
- normal run: 33 GB
- low-ram plus keeping encoders: 27.63 GB (incl. mlx cache)
- low-ram plus keeping transformer: 22.44 GB (incl. mlx cache)

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,8 @@ mflux-generate --model dev --prompt "Luxury food photograph" --steps 25 --seed 2
 
 - **`--config-from-metadata`** or **`-C`** (optional, `str`): [EXPERIMENTAL] Path to a prior file saved via `--metadata`, or a compatible handcrafted config file adhering to the expected args schema.
 
+- **`--low-ram`** (optional): Reduces GPU memory usage by constraining the MLX cache size and releasing text encoders and transformer components after use. This option is only compatible with single image generation. While it may slightly decrease performance, it helps prevent system memory swapping to disk, allowing generation on systems with limited RAM.
+
 <details>
 <summary>parameters supported by config files</summary>
 
--- a/src/mflux/callbacks/callback.py
+++ b/src/mflux/callbacks/callback.py
@@ -32,6 +32,17 @@ class InLoopCallback(Protocol):
         ...
 
 
+class AfterLoopCallback(Protocol):
+    def call_after_loop(
+        self,
+        seed: int,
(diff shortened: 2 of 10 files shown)


filipstrand:
This looks really good and a great use of the newly introduced callbacks! Could you please also add this functionality to `generate_controlnet.py` as well?

ssakar:
Thanks! I have updated the pull request with the controlnet changes. The joker controlnet example uses around 18 GB (incl. cache) with low-ram enabled. Without the option, I had to kill the process when it reached 30 GB.

filipstrand:
Great, thank you! Very good that you pointed out that this technique is only valid for the single-image case since we delete the resources progressively as we go through the pipeline. I'll go ahead and merge this to main.

PR #134, opened 2025-03-01
Inpaint workflow: add cv2 mask creator and args for --masked-image-path

anthonywu:
This PR bases on #129 and can be merged into @filipstrand's WIP branch to continue his work as-is without merge issues.

With the inpaint feature, we're going to expect the user or client GUIs to provide the complementary mask image.

In this PR I propose:

1. `--masked-image-path` argparser addition
2. with no additional python deps, use the already installed `opencv-python`/`import cv2` library to display a basic window that lets the user draw a binary mask over their original image

For discussion but not implemented: maybe the `--masked-image-path` can be set to a value of literal `interactive` (or empty?) and the CLI can bring up the cv2 GUI to allow a live mask drawing. IMO should be pursued in a separate PR after the basic inpaint functionality is done, to not scope creep.

Also, if we move the `cv2` tool inside the library, maybe there should be a `mflux.pre_processing` package for this and alike tools. (while reasonably restricting to existing library dependencies)

# Usage

`python tools/inpaint_mask_tool.py ~/Downloads/presidio-view.webp`

# GUI

image

the mask image can be seen in Preview.app as soon as you hit `s` for save.

image

# Example

image

via `mflux-generate -m dev-fill --init-image-path ~/Downloads/presidio-view.webp --masked-image-path ~/Downloads/presidio-view_mask.png --prompt "godzilla towering over the hills" --steps 8 --seed 1` (in my WIP workspace)

### DIFF
--- a/src/mflux/ui/cli/parsers.py
+++ b/src/mflux/ui/cli/parsers.py
@@ -10,7 +10,7 @@ from mflux.ui import defaults as ui_defaults
 
 class ModelSpecAction(argparse.Action):
     def __call__(self, parser, namespace, values, option_string=None):
-        if values in ["dev", "schnell"]:
+        if values in ["dev", "dev-fill", "schnell"]:
             setattr(namespace, self.dest, values)
             return
 
@@ -68,6 +68,7 @@ class CommandLineParser(argparse.ArgumentParser):
         self.supports_image_to_image = True
         self.add_argument("--init-image-path", type=Path, required=required, default=None, help="Local path to init image")
         self.add_argument("--init-image-strength", type=float, required=False, default=ui_defaults.INIT_IMAGE_STRENGTH, help=f"Controls how strongly the init image influences the output image. A value of 0.0 means no influence. (Default is {ui_defaults.INIT_IMAGE_STRENGTH})")
+        self.add_argument("--masked-image-path", type=Path, required=False, default=None, help="Local path to separate masked image as complement to --init-image-path")
 
     def add_batch_image_generator_arguments(self) -> None:
         self.add_argument("--prompts-file", type=Path, required=True, default=argparse.SUPPRESS, help="Local path for a file that holds a batch of prompts.")
--- a/tools/inpaint_mask_tool.py
+++ b/tools/inpaint_mask_tool.py
@@ -1,0 +1,154 @@
+from pathlib import Path
+
+import cv2
+import numpy as np
+
+
+class MaskCreator:
+    BRUSH_SIZES = {"1": 2, "2": 4, "3": 8, "4": 16, "5": 32, "6": 48, "7": 96, "8": 192, "9": 384}
+
+    def __init__(self, image_path: Path):
+        self.original_image = cv2.imread(image_path)
+        if self.original_image is None:
+            raise FileNotFoundError(f"Could not open or find the image: {image_path}")
+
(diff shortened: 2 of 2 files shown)


src/mflux/ui/cli/parsers.py
```diff
@@ -68,6 +68,7 @@ def add_image_to_image_arguments(self, required=False) -> None:
         self.supports_image_to_image = True
         self.add_argument("--init-image-path", type=Path, required=required, default=None, help="Local path to init image")
         self.add_argument("--init-image-strength", type=float, required=False, default=ui_defaults.INIT_IMAGE_STRENGTH, help=f"Controls how strongly the init image influences the output image. A value of 0.0 means no influence. (Default is {ui_defaults.INIT_IMAGE_STRENGTH})")
+        self.add_argument("--masked-image-path", type=Path, required=False, default=None, help="Local path to separate masked image as complement to --init-image-path")
```
anthonywu, line 71: @filipstrand I'll leave it to you to remove the hard coding in #129's `generate.py`

tools/inpaint_mask_tool.py
```diff
@@ -0,0 +1,154 @@
...
+
+class MaskCreator:
+    BRUSH_SIZES = {"1": 2, "2": 4, "3": 8, "4": 16, "5": 32, "6": 48, "7": 96, "8": 192, "9": 384}
+
+    def __init__(self, image_path: Path):
+        self.original_image = cv2.imread(image_path)
+        if self.original_image is None:
+            raise FileNotFoundError(f"Could not open or find the image: {image_path}")
+
+        self.mask_output_path = image_path.with_name(f"{image_path.stem}_mask").with_suffix(".png")
```
anthonywu, line 15: this will save the mask image in the same parent dir as the source image (it does assume user has write access to the source image's parent directory)

trying to not over-complicate things for now, this IMO is the most obvious/friendly implementation

tools/inpaint_mask_tool.py
```diff
@@ -0,0 +1,154 @@
...
+        alpha = 0.5  # Transparency level
+        cv2.addWeighted(self.overlay, alpha, self.display_image, 1 - alpha, 0, self.display_image)
+
+        # Draw brush size indicator in the corner
+        text = f"Brush Size: {self.brush_size} " "(Hotkeys 1-9: change brush size) "
+        cv2.putText(
+            self.display_image,
+            text,
+            (10, 30),
+            cv2.FONT_HERSHEY_DUPLEX,
```
anthonywu, line 84: font is funky but the Hershey font family is the only one that ships with opencv and therefore no additional dependencies

filipstrand:
@anthonywu Looks really good, thanks a lot for the help with the GUI! 

I'll continue finishing up the last parts of the in-painting feature and include it in the release after v.0.6.0.   

There is very little work left for the in-context lora feature now and I'm planning on releasing v.0.6.0 tomorrow. Currently, only the boring parts (release notes, readme, tests etc.) are left haha :)

src/mflux/ui/cli/parsers.py
filipstrand: Absolutely, I'll clean this up!

tools/inpaint_mask_tool.py
filipstrand: Make a lot of sense!

PR #135, opened 2025-03-04
Update arg parser test fixture

DanaCase:
Hi there,

When I was testing the latest updates I noticed that the parser tests were failing.

I updated the test fixture to include the new low-ram parameter.


DanaCase:
Oh one update, the low-ram param loading was also missing from the train script.

PR #136, opened 2025-03-04
clear prompt cache for training

DanaCase:
Hi again :)

I noticed that the prompt cache was breaking the training loop.

I just added a line to clear the prompt cache after image generation in training.


DanaCase:
Thanks!

A note for writing tests:
I did notice the training loop failing even on the _example data set + training.json
It crashes on the first ``` training_state.optimizer.optimizer.update(model=flux, gradients=grads)``` after test image generation.

So I think somehow there is a side effect causing the embeddings to be stale?

PR #138, opened 2025-03-08
remedy missed ruff check/formats in 0.6.0 release

anthonywu:
I am currently trying to rebase #129 inpainting PR on the latest main, and running into lots of unrelated diffs, much of it just related to missed `make format` and `make check` calls prior to merging #131.

Getting this out of the way so we can move forward without having to heal the repo state from a feature-focused branch.

### DIFF
--- a/src/mflux/dreambooth/state/training_state.py
+++ b/src/mflux/dreambooth/state/training_state.py
@@ -108,13 +108,13 @@ class TrainingState:
     def get_current_validation_image_path(self, training_spec: TrainingSpec) -> Path:
         output_path = Path(training_spec.saver.output_path) / DREAMBOOTH_PATH_VALIDATION_IMAGES
         output_path.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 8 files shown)


src/mflux/dreambooth/state/training_state.py
```diff
@@ -108,13 +108,13 @@ def should_generate_image(self, training_spec: TrainingSpec) -> bool:
     def get_current_validation_image_path(self, training_spec: TrainingSpec) -> Path:
         output_path = Path(training_spec.saver.output_path) / DREAMBOOTH_PATH_VALIDATION_IMAGES
         output_path.mkdir(parents=True, exist_ok=True)
-        path = output_path / Path(f"{self.iterator.num_iterations :07d}_{DREAMBOOTH_FILE_NAME_VALIDATION_IMAGE}.png")
```
anthonywu, line 111: this looks like a bug that introduces extra space, but is actually a no-op.

anyway, the only trouble it causes is drifting away from `ruff format` style expectations

filipstrand:
@anthonywu Thanks for noticing and fixing this! Maybe my pre-commit hooks was not working as expected... Will reinstall to make sure

PR #140, opened 2025-03-09
Fix: image gen should stop even if no callbacks are registered

anthonywu:
# Bug

#124 introduced a change where the `KeyboardInterrupt` event in the time step loop was refactored into the stepwise callback handler, so the effect was that we can only interrupt a multi-step generation iff we enabled the stepwise outputs.

The original intention was to make any generation stoppable, the quickest test is running a schnell query with a small number of steps:

```
mflux-generate -m schnell --prompt "stop sign" --steps 3
 33%|████████████████████████████████████████████████████████████                                                                                                                        | 1/3 [00:42<01:24, 42.11s/it] 33%|████████████████████████████████████████████████████████████                                                                                                                        | 1/3 [01:17<02:35, 77.66s/it]
Stopping image generation at step 2/3
```

# Fix

This change restores the `StopImageGenerationException` to all generations.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "mflux"
-version = "0.6.0"
+version = "0.6.1"
 description = "A MLX port of FLUX based on the Huggingface Diffusers implementation."
 readme = "README.md"
 keywords = ["diffusers", "flux", "mlx"]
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "mflux"
-version = "0.6.0"
+version = "0.6.1"
```
anthonywu, line 7: @filipstrand up to you if you want to make a small fix-it release

filipstrand:
Great! This was a mistake on my end to remove.

pyproject.toml
filipstrand: I'll do another release later today with this fix included

PR #142, opened 2025-03-10
Fix test: auto seed count == 1 sporadic test failure

anthonywu:
The sporadic error was due to the output file name having different behavior when auto seeds count is 1 vs > 1. By adjusting the test to always ask for n > 1, we can bypass this issue.

The offending code in `src/mflux/ui/cli/parsers.py`:

```python
if self.supports_image_generation and len(namespace.seed) > 1:
    # auto append seed-$value to output names for multi image generations
    # e.g. output.png -> output_seed_101.png output_seed_102.png, etc
    output_path = Path(namespace.output)
    namespace.output = str(output_path.with_stem(output_path.stem + "_seed_{seed}"))
```


filipstrand:
Nice!

PR #143, opened 2025-03-10
More low-ram improvements

ssakar:
This is the second part of memory optimizations. This time I have applied the ruff format and run some tests (the image generation failed on my system even without this changeset). 

I have removed the limitation of just single image generation. As we have the prompt cache, keeping the transformers allows us to repeatedly generate images. That increases memory but it might be more useful for some.

I have also added an explicit evaluation in the decoder (and encoder) so things don't accumulate and consume memory. That made some noticeable difference (with 1024x1024) The reported peak MLX memory usage now also captures the decoding phase. Here are some stats for this changeset:

### schnell q4, 4 steps, 1024x1024

- low-ram: 10.35 GB
- low-ram without eval during decode: 12.66 GB
- low-ram multiple seeds: 15.86 GB
- low-ram multiple seeds without eval during decode: 19.35 GB

### schnell q4, 4 steps, 512x512

- low-ram: 8.38 GB
- low-ram without eval during decode: 8.38 GB
- low-ram multiple seeds: 10.90 GB
- low-ram multiple seeds without eval during decode: 10.90 GB

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ mflux-generate --model dev --prompt "Luxury food photograph" --steps 25 --seed 2
 
 - **`--config-from-metadata`** or **`-C`** (optional, `str`): [EXPERIMENTAL] Path to a prior file saved via `--metadata`, or a compatible handcrafted config file adhering to the expected args schema.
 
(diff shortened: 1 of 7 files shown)


filipstrand:
@ssakar Thanks again for your contribution! I have rebased the branch, removed the two mx.eval calls and also added the `keep_transformer` option for `generate_in_context.py` and `generate_fill.py`

PR #147, opened 2025-03-13
Fix the parser error in save

Goekdeniz-Guelmez:
comand:
```text
mflux-save \
    --path /Users/gokdenizgulmez/Library/Mobile\ Documents/com\~apple\~CloudDocs/Transformer\ Models/MLX/dev-realism-8bit \
    --model dev \
    --quantize 8 \
    --lora-repo-id XLabs-AI/flux-RealismLora \
    --lora-scales 0.7
```

Error before:
```text
Traceback (most recent call last):
  File "/Users/gokdenizgulmez/.local/bin/mflux-save", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/gokdenizgulmez/.local/share/uv/tools/mflux/lib/python3.12/site-packages/mflux/save.py", line 9, in main
    args = parser.parse_args()
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/gokdenizgulmez/.local/share/uv/tools/mflux/lib/python3.12/site-packages/mflux/ui/cli/parsers.py", line 199, in parse_args
    if namespace.low_ram and len(namespace.seed) > 1:
       ^^^^^^^^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'low_ram'
```

After Fix:
```text
Saving model dev with quantization level 8

Fetching 8 files: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 8/8 [00:00<00:00, 15534.46it/s]
Fetching 7 files: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 7/7 [00:00<00:00, 68759.08it/s]
Model saved at /Users/gokdenizgulmez/Library/Mobile Documents/com~apple~CloudDocs/Transformer Models/MLX/dev-realism-8bit
```

### DIFF
--- a/src/mflux/ui/cli/parsers.py
+++ b/src/mflux/ui/cli/parsers.py
@@ -196,7 +196,7 @@ class CommandLineParser(argparse.ArgumentParser):
         if self.supports_image_generation and namespace.steps is None:
             namespace.steps = ui_defaults.MODEL_INFERENCE_STEPS.get(namespace.model, None)
 
-        if namespace.low_ram and len(namespace.seed) > 1:
+        if getattr(namespace, 'low_ram', False) and len(namespace.seed) > 1:
             self.error("--low-ram cannot be used with multiple seeds")
 
         return namespace


anthonywu (approved):

anthonywu:
This is a symptom of not using argparser sub parsers so we're mingling all the different command modes together. A thorough refactoring could do sub-commands: https://docs.python.org/3/library/argparse.html#sub-commands

However the current setup is simple to maintain for now, caveat being we need to use `getattr` check to allow for some modes to not expect/have some flags.

filipstrand:
@anthonywu Yes. I feel like I have lost track a bit of all things that happens in `parse_args`, so a future refactor would be nice. 

I guess some kind of tests for the entry-points files (`generate.py`, `save.py` etc.) could have caught this type of error, but at the same time it feels wrong to test at that level, especially since we have the full image-generation tests already, and the only additional thing a full end-to-end test would check would be the parser setup.

PR #148, opened 2025-03-13
adding a openai api compatible server

Goekdeniz-Guelmez:
# Examples
Bildschirmfoto 2025-03-13 um 11 20 25

# Setting in OpenWebUI
Bildschirmfoto 2025-03-13 um 11 20 38


anthonywu (changes requested):
This is very reasonable, I support shipping something like this in the library itself. Thank you for using modern libraries and keeping the implementation simple. Will need to 

- discuss module location
- isolate the additional requirements into optional dependencies in pyproject.toml
- create a command line launcher for this that is consistent with pre-existing commands
- make sure everything passes the lint checks

src/mflux/server.py
```diff
@@ -0,0 +1,246 @@
...
+                lora_paths=lora_paths,
+                lora_scales=lora_scales,
+            )
+            
+            # Store in cache
+            model_cache[model_key] = flux
+        
+        # Handle the keep_alive parameter
+        if request.keep_alive != -1:  # -1 means keep forever
+            if request.keep_alive == 0:  # 0 means unload immediately after use
```
anthonywu, line 125: suggestion, use module level constants like:

```
MODEL_KEEP_ALIVE_FOREVER = -1
MODEL_UNLOAD_IMMEDIATELY = 0
```

and the logical operators usage throughout the file will be self-documenting

src/mflux/server.py
```diff
@@ -0,0 +1,246 @@
...
+            CallbackRegistry.register_before_loop(memory_saver)
+            CallbackRegistry.register_in_loop(memory_saver)
+            CallbackRegistry.register_after_loop(memory_saver)
+        
+        images = []
+        
+        try:
+            for i in range(request.n or 1):
+                # Use provided seed or generate a new one for each image
+                current_seed = request.seed if request.seed is not None else int(time.time()) + i
```
anthonywu, line 161: suggestion for replacing the `int(time.time()) + i` logic

In Python, you can use the `time.monotonic()` function from the built-in time module to get monotonic time. Monotonic time is guaranteed to never go backwards, even if the system clock is changed.

```
import time

# Get current monotonic time
start = time.monotonic()

# Do some work
time.sleep(1)  # Simulate work

# Get elapsed time
end = time.monotonic()
elapsed = end - start

print(f"Elapsed time: {elapsed} seconds")
```

src/mflux/server.py
```diff
@@ -0,0 +1,246 @@
...
+        raise HTTPException(status_code=500, detail=f"Generation failed: {str(e)}")
+    
+    return GenerationResponse(
+        created=int(time.time()),
+        data=images
+    )
+
+if __name__ == "__main__":
+    import uvicorn
+    uvicorn.run(app, host="0.0.0.0", port=8800)
```
anthonywu, line 246: suggestion to make consistent with other CLIs with parameters:

```python
import argparse

def create_parser():
    parser = argparse.ArgumentParser(description='mflux OpenAI DALL·E-compatible API Server configuration')
    parser.add_argument('--host', default='0.0.0.0', type=str,
                        help='Host address to bind to (default: 0.0.0.0)')
    parser.add_argument('--port', default=8080, type=int,
                        help='Port to listen on (default: 8800)')
    return parser

if __name__ == '__main__':
    parser = create_parser()
    args = parser.parse_args()
    print(f"Starting server on {args.host}:{args.port}")
```

anthonywu (changes requested):
Hope this review helps @filipstrand make a decision on inclusion.

README.md
```diff
@@ -612,6 +613,65 @@ In the examples above the following LoRAs are used [Sketching](https://civitai.c
...
+# 🌐 Running the OpenAI API Compatible Server
+
+MFLUX includes a server component that provides an OpenAI API-compatible interface, allowing you to integrate MFLUX with existing tools and workflows that support the OpenAI API.
+
+### 📦 Installation
+
+To install the server component, use the `[server]` extra when installing MFLUX:
+
+```sh
+pip install mflux[server]
```
anthonywu, line 625: Lets sync this to the recommended `uv pip install ...`. Sorry I didn't just say this in the last review.

README.md
```diff
@@ -612,6 +613,65 @@ In the examples above the following LoRAs are used [Sketching](https://civitai.c
...
+```
+
+### 🚀 Starting the Server
+
+Once installed, you can start the server using the `mflux-server` command:
+
+```sh
+mflux-server
```
anthonywu, line 633: lets discuss with @filipstrand whether the OpenAI server should be recognized as just 1 of N options.

Ways we can go about this:

1. `mflux-server openai`
2. `mflux-server automatic1111`
3. `mflux-server anyothervariant`

or for now just `mflux-openai-server`

thinking ahead to future adjacent implementations I want to advocate for being able to stay backwards compatible to now without ripping defaults and precedents away from users

pyproject.toml
```diff
@@ -131,3 +131,16 @@ section-order = [
...
+server = [
+    # dependencies via fastapi and pydantic
+    "annotated-types>=0.7.0,<1.0.0",
+    "anyio>=4.8.0,<5.0.0",
+    "fastapi>=0.115.11,<1.0.0",
+    "pydantic>=2.10.6,<3.0.0",
+    "pydantic-core>=2.27.2,<3.0.0",
+    "sniffio>=1.3.1,<2.0.0",
+    "starlette>=0.46.1,<1.0.0",
+]
```
anthonywu, line 146: sorry about this super nit, please use a newline at end of files

image


> ## The UNIX Standard
> In the UNIX world, a text file is defined as a sequence of lines, each ending with a newline character ( \n ). Without a newline, the last line is considered incomplete. Many tools and scripts expect this convention, and omitting it can lead to unexpected behavior or errors.

you can configure any IDE or text editor to auto-insert the final newline

this will ensure others will visit the repo/file and not unnecessarily generate this `git diff` to heal the missing `\n`, and for other tools' expectations

pyproject.toml
```diff
@@ -131,3 +131,16 @@ section-order = [
     "first-party",
     "local-folder",
 ]
+
+
+[project.optional-dependencies]
```
anthonywu, line 136: this section header already exists at L47:

```toml


dev = [
    "pytest>=8.3.0,<9.0",
    "pytest-timer>=1.0,<2.0",
]
```

Just move it there. In fact I think the test of `uv pip install -e .` would error on the dupe section's existence.

anthonywu (changes requested):
- leaving some more implementation detail comments for now
- still need to zoom out and consider project structure and future direction of alike-server implementations

README.md
```diff
@@ -612,6 +613,65 @@ In the examples above the following LoRAs are used [Sketching](https://civitai.c
...
+# Print the image URL
+print(response.data[0].url)
+```
+
+### 🛠️ Advanced Configuration
+
+The server supports additional configuration options that can be passed as command-line arguments. For a complete list, run:
+
+```sh
+mflux-server --help
```
anthonywu, line 670: mflux-server needs to be declared in the `[project.scripts]` section of `pyproject.toml` (see L57-61) after the module has been properly relocated to a package/module path (TBD awaiting @filipstrand's review).

anthonywu (changes requested):

Goekdeniz-Guelmez:
Your suggestions are implemented, however I haven't got a way of unloading the model after use properly, Is there a method or function in mflux do unload it correctly?

filipstrand:
@Goekdeniz-Guelmez Sorry for being a bit late to review this, been very busy last couple of days but will try to find some time this weekend to try this out and review. Appreciate the ongoing discussions between you and Anthony so far! We have had server contributions in the past that we've chosen not to include for various reasons, but at the same time this feels like a highly requested feature to have in the core library so I am open to have this be part of mflux. 

@anthonywu @Goekdeniz-Guelmez I've also played around with the MCP protocol at work last couple of days and got inspired by including server support that way too. If we had support of that kind, it would be easy to integrate local mflux functionality into any MCP client like Claude Desktop, Cursor etc.

src/mflux/server/common.py
```diff
@@ -0,0 +1,131 @@
...
+        model_key = f"{model_name}_{base_model}_{quantize}_{local_path}_{lora_paths}"
+        
+        # Check if we have this model in cache
+        if model_key in self.model_cache:
+            logger.info(f"Using cached model: {model_key}")
+            flux = self.model_cache[model_key]
+            
+            # Cancel any existing timer for this model
+            if model_key in self.model_timers and self.model_timers[model_key]:
+                self.model_timers[model_key].cancel()
```
akx, line 61:
```suggestion
            if timer := self.model_timers.get(model_key):
                timer.cancel()
```

src/mflux/server/common.py
```diff
@@ -0,0 +1,131 @@
...
+        
+        # Handle the keep_alive parameter
+        self._setup_model_timer(model_key, keep_alive)
+        
+        return flux, model_key
+    
+    def _setup_model_timer(self, model_key: str, keep_alive: int):
+        """Set up a timer to unload the model after the specified time"""
+        if keep_alive != MODEL_KEEP_ALIVE_FOREVER:  # Replace -1 with constant
+            if keep_alive == MODEL_UNLOAD_IMMEDIATELY:  # Replace 0 with constant
```
akx, line 87:
```suggestion
        if keep_alive != MODEL_KEEP_ALIVE_FOREVER:
            if keep_alive == MODEL_UNLOAD_IMMEDIATELY:
```

src/mflux/server/common.py
```diff
@@ -0,0 +1,131 @@
...
+        self.model_timers: Dict[str, threading.Timer] = {}
+    
+    def get_model(self, 
+                  model_name: str, 
+                  base_model: Optional[str] = None,
+                  quantize: int = 8,
+                  local_path: Optional[str] = None,
+                  lora_paths: Optional[list] = None,
+                  lora_scales: Optional[list] = None,
+                  keep_alive: int = 0) -> Flux1:
```
akx, line 34: Python has `datetime.timedelta()` for durations – it'd be more "native" than having to look up the docstring to know this is in minutes :)

src/mflux/server/common.py
```diff
@@ -0,0 +1,131 @@
...
+    return tempfile.mkdtemp()
+
+
+def cleanup_temp_file(file_path):
+    """Clean up a temporary file and its parent directory if empty"""
+    if file_path and os.path.exists(file_path):
+        os.remove(file_path)
+        parent_dir = os.path.dirname(file_path)
+        if os.path.exists(parent_dir) and not os.listdir(parent_dir):
+            os.rmdir(parent_dir)
```
akx, line 131: The call sites for these could probably use `tempfile.TemporaryDirectory` and `tempfile.NamedTemporaryFile` respectively, so you wouldn't need to think about cleanup at all.

src/mflux/server/cli.py
```diff
@@ -0,0 +1,26 @@
...
+    # Add other common arguments here
+    
+    args = parser.parse_args()
+    
+    if args.api == "openai":
+        openai_server.run_server(host=args.host, port=args.port)
+    elif args.api == "automatic1111":
+        automatic1111.run_server(host=args.host, port=args.port)
+    elif args.api == "comfyui":
+        comfyui.run_server(host=args.host, port=args.port)
```
akx, line 20: Maybe the unimplemented interfaces shouldn't be available here at all? I'd find it confusing that I'd ask for comfyui and get... an AttributeError? :)

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,243 @@
...
+from mflux.callbacks.instances.memory_saver import MemorySaver
+from mflux.callbacks.instances.stepwise_handler import StepwiseHandler
+from common import ModelManager, create_temp_directory, cleanup_temp_file, MODEL_KEEP_ALIVE_FOREVER, MODEL_UNLOAD_IMMEDIATELY
+
+# Configure the logger
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+console_handler = logging.StreamHandler()
+console_handler.setFormatter(logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s'))
+logger.addHandler(console_handler)
```
akx, line 25: Same here about logging configuration.

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,243 @@
...
+class Image(BaseModel):
+    b64_json: str
+    
+class GenerationResponse(BaseModel):
+    created: int
+    data: List[Image]
+
+@app.post("/v1/images/generations", response_model=GenerationResponse)
+async def generate_images(request: GenerationRequest = Body(...)):
+    try:
```
akx, line 64: Wrapping the entire body of a function in one gigantic `try:except:` is bad form...

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,243 @@
...
+        
+        # Handle size parameter
+        if request.size is not None:
+            # Check if size is valid for the selected model
+            if valid_sizes and request.size not in valid_sizes:
+                raise HTTPException(
+                    status_code=http.client.BAD_REQUEST, 
+                    detail=f"Invalid size '{request.size}' for model '{model}'. Valid sizes: {valid_sizes}"
+                )
+            width_str, height_str = request.size.lower().split('x')
```
akx, line 97: `.partition` is faster than `.split()` since it can only ever do up to a single split (which is what you want here, anyway):
```suggestion
            width_str, _, height_str = request.size.lower().partition('x')
```

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,243 @@
...
+            CallbackRegistry.register_before_loop(handler)
+            CallbackRegistry.register_in_loop(handler)
+            CallbackRegistry.register_interrupt(handler)
+
+        memory_saver = None
+        if request.low_ram:
+            memory_saver = MemorySaver(flux)
+            CallbackRegistry.register_before_loop(memory_saver)
+            CallbackRegistry.register_in_loop(memory_saver)
+            CallbackRegistry.register_after_loop(memory_saver)
```
akx, line 150: These callbacks don't seem to be cleaned up after a request, which means after 10 requests the global singleton callback registry would have a bunch of "dead" callbacks...

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,243 @@
...
+        
+        images = []
+        
+        try:
+            for i in range(request.n or 1):
+                if request.seed is not None:
+                    current_seed = request.seed + i
+                else:
+                    current_seed = int.from_bytes(os.urandom(4), byteorder='big')
```
akx, line 159: Is cryptographic randomness truly needed here?

```suggestion
                    current_seed = random.getrandbits(32)
```
should be enough to get a 32-bit randomness...

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,243 @@
...
+                    prompt=request.prompt,
+                    config=Config(
+                        num_inference_steps=request.n or 2,
+                        height=height,
+                        width=width,
+                        guidance=request.guidance or 4.0,
+                        image_path=image_path,
+                        image_strength=request.image_strength,
+                        controlnet_strength=request.controlnet_strength
+                    )
```
akx, line 173: This config should probably be created outside the generation loop?

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,243 @@
...
+    quantize: Optional[int] = 8
+    n: Optional[int] = 1
+    lora_repo_id: Optional[str] = None
+    lora_scale: Optional[float] = 0.7
+    low_ram: Optional[bool] = False
+    stepwise_output: Optional[bool] = False
+    metadata: Optional[bool] = False
+    local_path: Optional[str] = None
+    response_format: Optional[str] = None
+    keep_alive: Optional[int] = MODEL_UNLOAD_IMMEDIATELY
```
akx, line 53: I don't think most of these should have `Optional` in them – that means Pydantic would also consider `None` a valid value for them. It will read the defaults if the user doesn't pass them in explicitly. 


```suggestion
    model: str = "schnell"
    base_model: Optional[str] = None
    height: Optional[int] = HEIGHT
    width: Optional[int] = WIDTH
    size: Optional[str] = None
    quality: Optional[str] = None
    guidance: float = 4.0
    image_strength: Optional[float] = None
    controlnet_strength: Optional[float] = None
    seed: Optional[int] = None
    quantize: int = 8
    n: int = 1
    lora_repo_id: Optional[str] = None
    lora_scale: float = 0.7
    low_ram: bool = False
    stepwise_output: bool = False
    metadata: bool = False
    local_path: Optional[str] = None
    response_format: Optional[str] = None
    keep_alive: int = MODEL_UNLOAD_IMMEDIATELY
```

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,243 @@
...
+
+                # Clean up temp file
+                cleanup_temp_file(temp_img_path)
+
+                # Encode image to base64
+                buffer.seek(0)
+                b64_img = base64.b64encode(buffer.getvalue()).decode('utf-8')
+                
+                # Add the image to our response list
+                images.append(Image(b64_json=b64_img))
```
akx, line 198: Using `NamedTemporaryFile` something like this should be a bit easier.
```suggestion
                with tempfile.NamedTemporaryFile(prefix="mflux", suffix="png") as tf:
                    generated_image.save(
                        path=tf.name, 
                        export_json_metadata=request.metadata,
                    )
                    tf.seek(0)
                    b64_img = base64.b64encode(tf.read()).decode('utf-8')
                
                # Add the image to our response list
                images.append(Image(b64_json=b64_img))
```

src/mflux/server/server.py
```diff
@@ -0,0 +1,258 @@
...
+    
+class GenerationResponse(BaseModel):
+    created: int
+    data: List[Image]
+
+@app.post("/v1/images/generations", response_model=GenerationResponse)
+async def generate_images(request: GenerationRequest = Body(...)):
+    try:
+        # Determine the image path (if an image upload feature is needed, it would require a different approach)
+        image_path = None
```
akx, line 81: This local is never set to anything but `None`?

src/mflux/server/server.py
```diff
@@ -0,0 +1,258 @@
...
+        raise HTTPException(status_code=500, detail=f"Generation failed: {str(e)}")
+    
+    return GenerationResponse(
+        created=int(time.time()),
+        data=images
+    )
+
+if __name__ == "__main__":
+    import uvicorn
+    uvicorn.run(app, host="0.0.0.0", port=8800)
```
akx, line 258: This probably shouldn't be here, since you'd want to use the CLI module to run this.
```suggestion
```

akx (commented):
Some humble suggestions :) I came here to see if there was some sort of server here already, so figured I'd not write another but contribute by way of review.

filipstrand:
I've quickly tried the server in OpenWebUI, it works there, but I feel the UX is a bit suboptimal - often times, when I try to generate the image, I get back a text saying it cannot be generated, even though it works: 

Screenshot 2025-03-31 at 07 27 27

Do you get the same behaviour?

filipstrand:
Tried setting up a MCP solution quickly (with Claude Desktop as the client) was able to have it generate (and display) the image there:
Screenshot 2025-03-31 at 07 22 11

I guess a benefit of this approach also is that MCP seems to be getting more wide support across clients (read that OpenAI Desktop client will soon support it, there seems to be [ongoing discussions](https://github.com/open-webui/open-webui/discussions/7363) for having it in OpenWebUI etc. 

Tried Cursor as a client too, but it does not support rendering the generated image, so this could limit the options for clients to use for image generation. 

Some pros I can see for having an official MCP server:
- Seems to be gaining momentum across the community
- Very natural interface for working with the image generation tool - inputs are often very transparent, the agent understands the API (e.g user can ask in natural text about things such as the prompt, resolution, steps, seed etc)
- You can iterate on a previous generate_image() call etc very natural in a conversation, and we don't need to hardcode much on the server  

@Goekdeniz-Guelmez @anthonywu what are your thoughts on this? Given an official MCP client, would there be need of also supporting specific one like the openai compatible one?

anthonywu:
If we choose to invest in one, MCP is my vote.

filipstrand:
Let's go with both options then! Once added to the project, these servers may be hard to remove in the future, but hopefully they won't need too much maintenance once they are in place. @Goekdeniz-Guelmez once you feel happy with the functionality and state of the code, feel free to ping me for a review and I'll try to take a look at this in the coming days

Goekdeniz-Guelmez:
@filipstrand Thanks for the quick reply!!! I do have a question, the model is not being unloaded form the RAM, correctly. I believe it's either lingering references or the Timer manager s not correct. Would you mind taking a look at the common.py file where the model handling is?

pyproject.toml
```diff
@@ -49,6 +49,17 @@ dev = [
...
 ]
+server = [
+    # dependencies via fastapi and pydantic
+    "annotated-types>=0.7.0,<1.0.0",
+    "anyio>=4.8.0,<5.0.0",
+    "fastapi>=0.115.11,<1.0.0",
+    "pydantic>=2.10.6,<3.0.0",
+    "pydantic-core>=2.27.2,<3.0.0",
+    "sniffio>=1.3.1,<2.0.0",
+    "starlette>=0.46.1,<1.0.0",
```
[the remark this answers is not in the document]
akx: This doesn't seem resolved? I think you could just keep `"fastapi[standard]>=0.115.11,<1.0.0",` in here and let it pull in the rest. (See the [FastAPI docs](https://fastapi.tiangolo.com/#standard-dependencies) for what the "standard" extra means.)

src/mflux/server/cli.py
```diff
@@ -0,0 +1,26 @@
...
+    
+    if args.api == "openai":
+        from . import openai_server
+        openai_server.run_server(host=args.host, port=args.port)
+    elif args.api == "automatic1111":
+        from . import automatic1111
+        automatic1111.run_server(host=args.host, port=args.port)
+    elif args.api == "comfyui":
+        from . import comfyui
+        comfyui.run_server(host=args.host, port=args.port)
```
akx, line 21: As said before, I don't think there's a point in having the stub implementations that wouldn't ever work in this PR?

src/mflux/server/common.py
```diff
@@ -0,0 +1,131 @@
...
+
+MODEL_KEEP_ALIVE_FOREVER = -1
+MODEL_UNLOAD_IMMEDIATELY = 0
+
+# Configure the logger
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+console_handler = logging.StreamHandler()
+console_handler.setFormatter(logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s'))
+logger.addHandler(console_handler)
```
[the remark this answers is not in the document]
akx: This doesn't seem resolved?

src/mflux/server/common.py
akx: This is still a valid suggestion IMO – no one knows what those comments mean.

src/mflux/server/common.py
```diff
@@ -0,0 +1,136 @@
...
+                timer.start()
+                logger.info(f"Model {model_key} will be unloaded in {keep_alive} minutes if not used again")
+        else:
+            logger.info(f"Model {model_key} will remain loaded indefinitely")
+    
+    def unload_model(self, model_key: str):
+        """Unload a model from cache when its timer expires"""
+        if model_key in self.model_cache:
+            logger.info(f"Unloading model: {model_key}")
+            del self.model_cache[model_key]
```
akx, line 113: Does deleting the model from the dict actually reliably trigger moving it off the device it's on?

src/mflux/server/common.py
```diff
@@ -0,0 +1,136 @@
...
+            if model_key in self.model_timers:
+                del self.model_timers[model_key]
```
akx, line 115:
```suggestion
            self.model_timers.pop(model_key, None)
```

src/mflux/server/common.py
```diff
@@ -0,0 +1,136 @@
...
+    
+    def unload_model_after_use(self, model_key: str):
+        """Unload a model immediately after use"""
+        if model_key in self.model_cache:
+            logger.info(f"Unloading model immediately as requested: {model_key}")
+            del self.model_cache[model_key]
+            if model_key in self.model_timers:
+                del self.model_timers[model_key]
```
akx, line 123:
```suggestion
        """Unload a model immediately after use"""
        logger.info(f"Unloading model immediately as requested: {model_key}")
        self.unload_model(model_key)
```

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,277 @@
...
+
+            # Clean up temp file
+            cleanup_temp_file(temp_img_path)
+
+            # Encode image to base64
+            buffer.seek(0)
+            b64_img = base64.b64encode(buffer.getvalue()).decode('utf-8')
+            
+            # Add the image to our response list
+            images.append(Image(b64_json=b64_img))
```
akx, line 224: I still think this should use `NamedTemporaryFile` – was there a reason not to?

src/mflux/server/openai_server.py
akx: The separated try-excepts that do the same thing don't make this much better IMO.

If you want the same effect, you could refactor the guarded part into a separate function:

```python
@app.post("/v1/images/generations", response_model=GenerationResponse)
def generate_images(request: GenerationRequest = Body(...)):
    try:
        return _generate_images(request)
    except Exception as e:
        logger.error(f"Generation failed: {e}", exc_info=True)
        raise HTTPException(status_code=http.client.INTERNAL_SERVER_ERROR, detail=f"Generation failed: {e}")


def _generate_images(request: GenerationRequest):
    ...  # no need for try-except here
```

Btw, note that this endpoint shouldn't be `async`, since marking it async when it's actually blocking will just block the server further.

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,277 @@
...
+
+        memory_saver = None
+        if request.low_ram:
+            memory_saver = MemorySaver(flux)
+            CallbackRegistry.register_before_loop(memory_saver)
+            registered_callbacks.append(("before_loop", memory_saver))
+            CallbackRegistry.register_in_loop(memory_saver)
+            registered_callbacks.append(("in_loop", memory_saver))
+            CallbackRegistry.register_after_loop(memory_saver)
+            registered_callbacks.append(("after_loop", memory_saver))
```
akx, line 170: It would be nicer to improve `CallbackRegistry` to support temporary callbacks with a context manager :)

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,277 @@
...
+        
+        images = []
+    
+    except Exception as e:
+        logger.error(f"Generation failed: {e}", exc_info=True)
+        raise HTTPException(status_code=http.client.INTERNAL_SERVER_ERROR, detail=f"Generation failed: {e}")
+        
+    try:
+        for i in range(request.n or 1):
```
akx, line 179: No need for the `or 1`.
```suggestion
        for i in range(request.n):
```

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,277 @@
...
+            if request.seed is not None:
+                current_seed = request.seed + i
+            else:
+                current_seed = random.getrandbits(32)
+            
+            config = Config(
+                num_inference_steps=request.n or 2,
```
akx, line 186:
```suggestion
                num_inference_steps=request.n,
```

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,277 @@
...
+                height=height,
+                width=width,
+                guidance=request.guidance or 4.0,
```
akx, line 189:
```suggestion
                guidance=request.guidance,
```

src/mflux/server/openai_server.py
```diff
@@ -0,0 +1,277 @@
...
+            model_manager.unload_model_after_use(model_key)
+        
+        # Clean up temporary files
+        if image_path:
+            cleanup_temp_file(image_path)
+
+    return GenerationResponse(
+        created=int(time.time()),
+        data=images
+    )
```
akx, line 257: This `return` could be inside the `try:` block of the `try:finally:`.

src/mflux/server/server.py
```diff
@@ -0,0 +1,265 @@
+from fastapi import FastAPI, HTTPException, Body
```
akx, line 1: This file is now a partial duplicate of `openai_server.py`? 🤔

src/mflux/server/openai_server.py
filipstrand: Agree, I think it can be improved in a few ways. 

More generally: I introduced the `CallbackRegistry` quite recently and did not put a lot of effort into it other than supporting the few use cases we had at the time (such as saving the stepwise image generation etc). Back then I opted for a global  registry, as it was simple and I when the "entry-point" of the application was the image generation itself and nothing more. But with a server running and managing the image generations, I could imagine that we would want fancier solutions over time, where the callbacks are model-specific, for example, and not fully global (for example, if we wanted to support model switching with different behaviours etc etc). This does not need to be done in this PR of course, but it feels like a few things could be re-architected on this front.

src/mflux/server/server.py
filipstrand: `automatic1111.py` is also missing an implementation at the moment

src/mflux/server/testing.py
```diff
@@ -0,0 +1,74 @@
+# Modified test_api.py using openai package
+import base64
+from PIL import Image
+import io
+import json
+from openai import OpenAI
+
+def test_api():
```
filipstrand, line 8: For testing, I would instead place this kind of file in the `tests` package. Probably under a `server` section.

src/mflux/server/testing.py
```diff
@@ -0,0 +1,74 @@
...
+    # Configure the client to use your local server
+    client = OpenAI(
```
filipstrand, line 10: Like the idea that you wanted to test the server at a high level, one thing I am a bit worried about is is that it is now tied to having a api key from open ai or other providers (I don't have one personally from OpenAI)

I can see some solutions/workarounds for these kinds of problems: 

- Somehow injecting the prompt externally for test purposes (I assume that is the only reason why we need the key?)
-  Have these kind of tests annotated by a special test decorator which would be skipped by default when running the standard test suite

and maybe other ways...

filipstrand:
Saw this recently also https://github.com/open-webui/openapi-servers#-bridge-to-mcp-optional but have not tried it yet, but seems like OpenWebUI has started to support MCP

src/mflux/server/common.py
Goekdeniz-Guelmez: Actually it's not guaranteed, hence it probably just removing the reference to the model. I'll add a garbage collector, because in python objects are garbage collected when there are no more references to them. I'll push the update tonight.

src/mflux/server/openai_server.py
Goekdeniz-Guelmez: The reason is that when using NamedTemporaryFile, the file is not send back properly, I believe that the file is opened immediately by Python, which can create conflicts when other processes (image library) try to write to it. I tried it at first too but dint work, hence me using this manual approach, because you create a simple string path for the temporary file, then the image library can freely open, write, and close this file, then the code opens it separately to read the data to send it and manually clean up the file.

src/mflux/server/openai_server.py
Goekdeniz-Guelmez: Yes, you're right. Especially when this grows more complex, especially with a server managing multiple image generations, the global registry could become problematic.

src/mflux/server/testing.py
Goekdeniz-Guelmez: I total agree, I also have to add better test handling and more tests.

src/mflux/server/testing.py
Goekdeniz-Guelmez: Yes some native tests too, the thing is that the OpenAi package decodes the images a little different and hence most people use and have the code running, this would mean minimal code changes, but yes l'll also add a direct API call with requests.

Goekdeniz-Guelmez:
Yes, that's really cool seeing this become a global standard. I looked into MCP servers and they should definitely be added int here too.

filipstrand:
> Yes, that's really cool seeing this become a global standard. I looked into MCP servers and they should definitely be added into here too.

You probably already know the official library https://github.com/modelcontextprotocol/python-sdk. I've played around with it a bit and found their examples quite intuitive.

anthonywu:
After doing some hello world implementation of MCP Python servers, I vote for MCP implementation as its own PR and module path as the optional dependencies should be isolated.

A bit busy with USA tax season here but I might be able to help out soon.

filipstrand:
I think there are some smaller cleanups to do (e.g removing mcp related things since we'll address that in upcoming PRs). 
I would also suggest removing `test_api_call.py` if that requires specific openai related keys - I think anyone should be able to run the test suite without any special setup.  

After some thought on this, I think we can go one of two routes with this PR: 
a) Merge this PR to main soon (after some of the lighter cleanup job mentioned above) and refactor some bits once we have a MCP server contribution. 
b) Keep this PR open until we have a MCP server PR to see what is similar and try to join the two efforts then.  
 
Also, when launching proper server support in the project, I think it would be best for it to be "feature-complete" and to support all Flux1 versions (Flux1, Flux1InContextLoRA, Flux1Fill etc... I can imagine followup feature-requests for the missing ones pretty soon after releasing this). I guess most of the work is the actual model-management part and my hope would be that a lot could be unified across the different servers we will support (looking at your implementation here of the Model Manager, this looks quite general so maybe it could be used mostly as is.)

In the short term, I'll be finishing up the remaining parts for the Redux and Depth functionality and have them be included in the upcoming v.0.7.0 release. If we go with a) and it lands in main before the 0.7.0 release, I think it is best to not promote it as a feature until it is fully complete (supports all the Flux variants), but instead aim for the release after that.    

What are your thoughts on a) and b) @Goekdeniz-Guelmez @anthonywu? I would prefer b), but on the other hand I am not the target user for the server functionally (yet, hehe :) ) and would not mind if this feature was delayed a bit. At the same time, I could also see that option a) might be preferable when developing the MCP solution as a 'warm start'. 

In other news, starting in 2 weeks, I'll be reducing my regular working hours to 50% for a few months so I'll have more time to focus on this project!

anthonywu:
This is now a stale PR, circling back to the concern that deep coupling with OpenAI DALLE's conventions and verbiage is a form of technical debt that this project should not take on.

Unless the backend implementation is very generalizable across many frontend tools (i.e. is a standard or a widely adopted de facto standard), it does not belong in the official project.

After thinking about this for a while, I think the implementation as is needs to be a separate python project (easy enough to setup using `uv init` and then creating a separate GitHub project for it).

PR #159, opened 2025-03-29
FLUX.1 Tools | Redux & Depth

filipstrand:
Adds the Redux and Depth functionality 

Also adds the following new models to support both features: 

- For Redux: SiglipVisionTransformer
- For Depth: Depth-map generation via [depth-pro](https://github.com/apple/ml-depth-pro) (not done yet) **Update** Now working!


azrahello:
Hi! I’ve been testing the recent updates — I was already impressed by how well it performs 😄 amazing work!

While experimenting, I noticed it currently supports only a single image at a time. In ComfyUI, it’s possible to chain multiple Redux calls, which really opens up creative possibilities. Do you think something like that could be implemented here as well?

Also, is there a way to adjust the balance between the prompt and SigLIP during generation? In ComfyUI, there’s a node that allows setting five influence levels — ultra low, low, medium, high, and ultra high — which can be quite useful. Would that be difficult to add?

Lastly, I think it might help if the image-path for Redux were separate from the one used for inpainting. That way, both techniques could be used simultaneously for more creative flexibility. That said, I don’t fully understand how everything works under the hood, so maybe there’s a good reason for the current setup — just thought it was worth mentioning!

azrahello:
Hi! Thanks for this wonderful Easter gift — and best wishes!
So, in ComfyUI, what connects the nodes is the “conditioning”, which is an input and the output of the prompt node and "style mode apply" as in figure. 
Screenshot 2025-04-21 alle 12 11 18 PM

In ComfyUI, the portion of code that handles this is here:
https://github.com/comfyanonymous/ComfyUI/blob/11b72c9c55d469c6f256eb0a8598e251ce504120/nodes.py#L1025
Through the “conditioning” node, it’s actually possible to chain together the entire block that manages the Redux technique.
(Sorry if I’m not using the most precise technical terms.)

Anyway, I tried tweaking the code a bit using Cursor and claude to allow it to accept multiple images. And I must say — it kind of works!
That said, I’m a bit worried I may have overlooked something that’s making the execution heavier than it should be. In ComfyUI, this operation doesn’t seem to add noticeable load. Basically, what I did was modify the files so that the tool accepts a list of image paths.
Then, I edited the redux.py file with the following changes:
```
class ReduxUtil:
    @staticmethod
    def embed_image(
        image_path: Union[str, Path, List[str], List[Path]],
        image_encoder: SiglipVisionTransformer,
        image_embedder: ReduxEncoder,
        max_embeddings: int = 4,  # Maximum number of embeddings to keep
        pooling_size: int = 2     # Pooling size to reduce dimensionality
    ) -> mx.array:
        if isinstance(image_path, (str, Path)):
            image_paths = [image_path]
        else:
            image_paths = image_path
            
        all_image_embeds = []
        
        for path in image_paths:
            # Basic process for each image
            image = ImageUtil.load_image(path).convert("RGB")
            image = ImageUtil.preprocess_for_model(image=image)
            image_latents, pooler_output = image_encoder(image)
            image_embeds = image_embedder(image_latents)
            
            # Dimensionality reduction via average pooling
            image_embeds = ReduxUtil._reduce_embedding_dimension(image_embeds, pooling_size)
            all_image_embeds.append(image_embeds)

        # If we have too many images, perform smart merging
        if len(all_image_embeds) > max_embeddings:
            all_image_embeds = ReduxUtil._merge_embeddings(all_image_embeds, max_embeddings)
            
        # Final concatenation
        if len(all_image_embeds) == 1:
            return all_image_embeds[0]
            
        return mx.concatenate(all_image_embeds, axis=1)

    @staticmethod
    def _reduce_embedding_dimension(embedding: mx.array, pooling_size: int) -> mx.array:
        """Reduces the dimensionality of the embedding using average pooling."""
        # Assumes embedding has shape (batch, sequence_length, hidden_size)
        seq_len = embedding.shape[1]
        if seq_len < pooling_size:
            return embedding
            
        # Reshape for pooling
        new_seq_len = seq_len // pooling_size
        embedding = embedding.reshape((1, new_seq_len, pooling_size, -1))
        # Average pooling
        embedding = mx.mean(embedding, axis=2)
        return embedding

    @staticmethod
    def _merge_embeddings(embeddings: List[mx.array], max_count: int) -> List[mx.array]:
        """Merges embeddings when there are too many."""
        if len(embeddings) <= max_count:
            return embeddings
            
        # Calculate how many embeddings to merge together
        merge_size = len(embeddings) // max_count + (1 if len(embeddings) % max_count else 0)
        merged = []
        
        for i in range(0, len(embeddings), merge_size):
            group = embeddings[i:i + merge_size]
            # Weighted average of embeddings in the group
            weights = mx.array([1.0 / len(group)] * len(group))
            merged_embed = mx.sum(mx.stack([w * e for w, e in zip(weights, group)]), axis=0)
            merged.append(merged_embed)
            
        return merged
```
After generating an image, I noticed that it was able to take information from both images, but there was no trace of the prompt. Since I didn’t know what to adjust anymore, I stopped there.

filipstrand:
Thanks for sharing! I tried your implementation quickly and it does seem to blend the images, kind of as expected. I also tried an even more naive approach by simply concatenating the two embeddings together (hardcoded image paths in the following example)

```python
    @staticmethod
    def _get_prompt_embeddings(
        prompt: str,
        prompt_cache: dict[str, tuple[mx.array, mx.array]],
        t5_tokenizer: TokenizerT5,
        clip_tokenizer: TokenizerCLIP,
        t5_text_encoder: T5Encoder,
        clip_text_encoder: CLIPEncoder,
        image_path: str,
        image_encoder: SiglipVisionTransformer,
        image_embedder: ReduxEncoder,
    ) -> (mx.array, mx.array):
        # 1. Encode the prompt
        prompt_embeds_txt, pooled_prompt_embeds_txt = PromptEncoder.encode_prompt(
            prompt=prompt,
            prompt_cache=prompt_cache,
            t5_tokenizer=t5_tokenizer,
            clip_tokenizer=clip_tokenizer,
            t5_text_encoder=t5_text_encoder,
            clip_text_encoder=clip_text_encoder,
        )

        # 2. Encode the image using the Siglip and Redux encoder
        image_embeds1 = ReduxUtil.embed_image(
            image_path="robot.png",
            image_encoder=image_encoder,
            image_embedder=image_embedder,
        )  # fmt:off

        image_embeds2 = ReduxUtil.embed_image(
            image_path="robot2.png",
            image_encoder=image_encoder,
            image_embedder=image_embedder,
        )  # fmt:off

        # 3. Join the text and image embeddings
        prompt_embeds = mx.concatenate([prompt_embeds_txt, image_embeds1, image_embeds2], axis=1)
        pooled_prompt_embeds = pooled_prompt_embeds_txt
        return prompt_embeds, pooled_prompt_embeds
```

Robot:
robot
Robot2:
robot2
Result
naive

More testing required, but  would be interesting to hear how does this approach work for you :)  Better/Worse/Equal to the pooling above?  

Since my simple approach here only concatenates the two embeddings, the shape of the resulting `prompt_embeds` are bigger so generation takes more time. I did not think of this before, but if we only consider the shapes, we can theoretically concatenate as many image embeddings as we want.

azrahello:
Hi, I’ve run some tests and now I’m even more confused 😅 — I couldn’t find any substantial differences.

While at it, I tried varying the guidance (cfg) values to observe how results change. What puzzles me is that, in the official Redux documentation on Hugging Face, the cfg is set to 2.5, whereas in ComfyUI the tendency is to raise it, often from 3.5 up to 6, or even 10.

Still, even after tweaking this value, I couldn’t detect any trace of prompt conditioning — the prompt doesn’t seem to have much influence on the generation.

It feels like the behavior and implementation of Redux within Diffusers is more rigid compared to the more flexible, dynamic approach in ComfyUI.
Maybe I should’ve noticed this earlier, but I wanted to share it in case it helps clarify something.

filipstrand:
> Hi, I’ve run some tests and now I’m even more confused 😅 — I couldn’t find any substantial differences.
> 
> While at it, I tried varying the guidance (cfg) values to observe how results change. What puzzles me is that, in the official Redux documentation on Hugging Face, the cfg is set to 2.5, whereas in ComfyUI the tendency is to raise it, often from 3.5 up to 6, or even 10.
> 
> Still, even after tweaking this value, I couldn’t detect any trace of prompt conditioning — the prompt doesn’t seem to have much influence on the generation.
> 
> It feels like the behavior and implementation of Redux within Diffusers is more rigid compared to the more flexible, dynamic approach in ComfyUI. Maybe I should’ve noticed this earlier, but I wanted to share it in case it helps clarify something.

Ok interesting... then I think we could go with the simpler approach to begin with if it gives roughly the same results. 

I'm also not sure to what extent the prompt conditioning changes as we concatenate more images embeddings. At least on a high level, the `pooled_prompt_embeds` that we get from the `clip_text_encoder` are unchanged in the current implementation (since they only depend strictly on the prompt itself)

filipstrand:
Things seems to be working as expected with the new depth- and redux additions. I'll merge this one now and start prepping for the 0.7.0 release.

README.md
```diff
@@ -869,6 +870,119 @@ mflux-generate-fill \
...
+*Image credit: [Paris Bilal on Unsplash](https://unsplash.com/photos/a-cat-statue-sitting-on-top-of-a-white-base-Bedd4qHGWCg)*
+
+##### Example
+
+The following examples show how to use the Redux tool to generate variations:
+
+```bash
+mflux-generate-redux \
+  --prompt "a grey statue of a cat on a white platform in front of a blue background" \
+  --redux-image-paths "original.png" \
```
anthonywu, line 938: Late comment: I noticed the plural wording of `paths` in the flag name, and we can totally support passing in a series of files, or a glob expression, and using that as an inner loop inside of N-seeds loop.

Would accepting N file paths be something you think we should support?

PR #160, opened 2025-03-30
Default inference steps to 4

akx:
Not passing in `--steps` on the command line would otherwise fail with "unsupported operand type(s) for /: 'int' and 'NoneType'" (as `args.steps` would be None).


akx:
@anthonywu Well, it clearly was an issue since @filipstrand ended up fixing it in https://github.com/filipstrand/mflux/pull/159/files#diff-0e3c48fa34274db5380f82da3bd235585b0069be68b10a067cbcd461824ba454R224 to default to 14 😉 

Closing this.

PR #161, opened 2025-03-30
Clean up fmt: off comments

akx:
* Use [magic trailing commas](https://docs.astral.sh/ruff/settings/#format_skip-magic-trailing-comma) instead of disabling formatting altogether
  * Black, and by heritage, Ruff Format, will keep arguments on separate lines if the last argument has a trailing comma. (I also prefer Ruff's COM812 rule to _enforce_ trailing commas like that, but that's not my call to make here.)
* Scope `fmt: off`s better otherwise

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.6.9
+    rev: v0.11.2
     hooks:
       # Run the linter.
       - id: ruff
@@ -11,6 +11,6 @@ repos:
       - id: ruff-format
         types_or: [python, pyi]
   - repo: https://github.com/crate-ci/typos
-    rev: v1.26.0
+    rev: v1.31.0
     hooks:
       - id: typos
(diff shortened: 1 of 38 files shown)


filipstrand:
@akx Much cleaner this way, thanks for the suggestion! I had the feeling that something like this could be solved easily but never looked into it.

filipstrand:
@akx I also like the idea of forcing trailing commas, however adding `COM812` it to the [tool.ruff.lint] section resulted in the following warning: 

```
warning: The following rule may cause conflicts when used with the formatter: `COM812`. To avoid unexpected behavior, we recommend disabling this rule, either by removing it from the `select` or `extend-select` configuration, or adding it to the `ignore` configuration.
```

If you have any suggestions for the best way of enforcing this, let me know

akx:
> I also like the idea of forcing trailing commas, however adding `COM812` it to the [tool.ruff.lint] section resulted in the following warning 

Yeah, the warning is unfortunate, but I've never seen it cause any actual issues. What they mean with "conflicts" apparently means that COM812 may make edits that cause the code to be needed to be reformatted again: https://github.com/astral-sh/ruff/issues/9216#issuecomment-1874384099

There doesn't seem to be a way to disable the warning, though. 😩

filipstrand:
Ah okay, a bit unfortunate, but then it sounds like it should probably be safe to add it

PR #171, opened 2025-04-25
MCP: initial proof of concept, minimal features, WIP

anthonywu:
This is a quick proof of concept, with minimal scoping to try out the popular `FastMCP` framework which is recommended by [MCP python SDK Examples](https://github.com/modelcontextprotocol/python-sdk?tab=readme-ov-file#quickstart)

# Modularization

Introducing a `mflux.ui.mcp` package adjacent to `mflux.ui.cli` which represents two different "user interface" entry points.

The system info stuff can possibly move to an util module, for now housing in same file for simplicity.

# Integration Success Criteria

- [x] Tools recognized by [Claude Desktop](https://modelcontextprotocol.io/quickstart/user)
- [x] Tools recognized by third party chat UI: [Groq Desktop](https://github.com/groq/groq-desktop-beta)
- [ ] Tools recognized by CLI MCP Client (i.e. a integration test script would do)
- [x] A reasonable prompt triggers tool call in Claude
- [x] A reasonable prompt triggers tool call in Groq Desktop
- [x] Images generate and are displayed on the chat UIs

# WIP

- I think the desktop clients may be wrapping the Python executable in some sort of sandbox for the Stdio server that is disagreeable with MLX and GPU usage

First demo of it working inside Claude Desktop:

image

image

the UX in Claude Desktop is not ideal, you have to scroll down in the little frame to see the image

image


pyproject.toml
```diff
@@ -64,6 +67,7 @@ mflux-generate-redux = "mflux.generate_redux:main"
 mflux-save = "mflux.save:main"
 mflux-save-depth = "mflux.save_depth:main"
 mflux-train = "mflux.train:main"
+mflux-mcp-tools = "mflux.ui.mcp.tools:main"
```
anthonywu, line 70: Claude Desktop config:

image

the literal json config on disk is:

```json
{
  "mcpServers": {
    "mflux": {
      "command": "/Users/anthonywu/workspace/mflux-ops/.venv/bin/mflux-mcp-tools",
      "args": []
    }
  }
}
```

config file is at `~/Library/Application Support/Claude/claude_desktop_config.json`

Groq Desktop:

image

src/mflux/ui/mcp/tools.py
```diff
@@ -0,0 +1,90 @@
...
+    prompt: str, steps: int, width: int, height: int, guidance: float, seed: int = None
+) -> MCPImage:
+    seed = seed or int(time.time())
+    args = argparse.Namespace(
+        model="schnell", steps=steps, prompt=prompt, width=width, height=height, guidance=guidance, seed=seed
+    )
+    return generate_image(args)
+
+
+@mcp.tool()
```
anthonywu, line 49: tools are recognized by both desktop clients:

image

image

image

image

src/mflux/ui/mcp/tools.py
```diff
@@ -0,0 +1,90 @@
...
+            "SPDisplaysDataType"
+        ][0]
+    except (IndexError, KeyError) as e:
+        return f"Error parsing system profile: {e}"
+    return SystemInfo(chipset_name=info["_name"], cpu_count=os.cpu_count(), gpu_count=info["sppci_cores"])
+
+
+# Add a dynamic greeting resource
+@mcp.resource("greeting://{name}")
+def get_greeting(name: str) -> str:
```
anthonywu, line 80: copied this from example, not sure how to reveal it in UIs yet, so this function is no-op for now

src/mflux/ui/mcp/tools.py
```diff
@@ -0,0 +1,90 @@
...
+
+@dataclass
+class SystemInfo:
+    chipset_name: str
+    cpu_count: int
+    gpu_count: int
+
+
+# Get number of logical CPUs
+def get_system_info() -> SystemInfo:
```
anthonywu, line 68: the idea is it'd be nice to greet the user in the UI with the system specs

anthonywu (commented):
WIP, the desktop chat -> python server is "talking" already, but the image generation causes my Mac to hard reboot.

For now, just showing you the WIP so you have an idea of where this could go.

src/mflux/ui/mcp/tools.py
filipstrand: Nice! I have not personally tried the Groq desktop app yet, cool that they have support too!

filipstrand:
On the topic of tools in general, I have found myself often reaching for the ones you added earlier for the fill functionality 
(`inpaint_mask_tool.py` and `create_outpaint_image_canvas_and_mask.py`). In a subsequent PR, after everything is up and running, I think we could expose these as well.

anthonywu:
Ah yeah great idea, let the LLMs build those annoying calls

src/mflux/ui/mcp/tools.py
```diff
@@ -0,0 +1,151 @@
...
+        bytes: The compressed image data as bytes
+    """
+    image.save(buffer := io.BytesIO(), compress_level=9, format=format, optimize=True)
+    client_image: MCPImage = None
+    client_image_data: str = None
+    while client_image is None or len(client_image_data) > client_max_length:
+        image = image.resize((int(image.width * (1 - size_down_step)), int(image.height * (1 - size_down_step))))
+        image.save(buffer := io.BytesIO(), compress_level=9, format=format, optimize=True)
+        client_image = MCPImage(data=buffer.getvalue(), format=format)
+        client_image_data = client_image.to_image_content().data
```
anthonywu, line 71: there's a size limit of 1 mebibyte (1024^2) allowable on the Claude Desktop client. At first this doesn't make sense, but a quick search reveals it's the message limit of stdio interface

anthonywu:
need to find/wait for a diff third party MCP client that shows images. Groq desktop as of today cannot display inline images: 
image

anthonywu:
Because of the long generation times that can run over five minutes for many cases, I'm not sure this is a good user experience on the client. Has anyone seen a MCP implementation that makes the user wait a few minutes to respond? I think the default has to be the schnell model with few steps, target a demo that can return an image in under one minute on recent hardware.

filipstrand:
> Because of the long generation times that can run over five minutes for many cases, I'm not sure this is a good user experience on the client...

You are right that this is indeed not optimal... It also feels like this ties back a bit to the discussion we had way back about adding a yield-generator to return partial progress from the denoise-loop. In terms of UI for the end user, that would of course be the optimal thing to do, but it would require some refactoring. My guess is also that there is no support to stream/continuously output information from a tool call to a client (but I haven't looked into this)?

It feels like the big win with MCP is currently on the input-side of things: The very natural tool-use and context/api understanding (and that it is plug-and-play with any compatible client). But when it comes to output, with more advanced things like showing images, output "streaming" etc., it feels like support is still lacking (and suddenly it is not as much plug-and-play across different clients). 

Still, I think this is cool to have and support will probably evolve over time

anthonywu:
@filipstrand the review bot may not work on incoming PR branches from forked repos:

```
This is an open PR, checking out PR branch...
fatal: couldn't find remote ref mcp-starter
Error: Prepare step failed with error: ShellError: Failed with exit code 128
Error: Process completed with exit code 1.
```

anthonywu:
Found 2 new Mac desktop GUI to test against:

- https://github.com/indragiek/Context/releases/tag/v1.0.3
- https://lmstudio.ai/blog/lmstudio-v0.3.17

anthonywu:
This implementation will introduce extra requirements that should not be imposed on CLI/Library-only users.

I've decided to continue development in a separate repo and pypi project, which I'll add @filipstrand as co-owners of.

They will be at:

1. https://github.com/anthonywu/mflux-mcp (project skeleton set up)
2. https://pypi.org/project/mflux-mcp/ (already published placeholder release)

I also considered making a mflux org to house affiliated projects, but I'll let @filipstrand decide if he wants that admin overhead.

PR #173, opened 2025-04-25
Fix up some path-related types

akx:
Mypy was (still is) lighting up like a christmas tree where strs are being passed as paths and the other way around, so this is a gentle pass to fix some of those issues.


filipstrand:
I know the feeling hehe (even though my current PyCharm setup did not complain about these particular things). Thanks for the updates!

PR #175, opened 2025-04-25
Fix type annotations for tuples

akx:
The correct syntax for tuple types is `tuple[a, b, c]` (or `tuple[a, ...]` for a homogenous n-ple), not a tuple literal.


filipstrand:
Oh, thanks for pointing this out! I mistakenly thought `()` syntax was a valid type annotation. Weird that my editor did not complain about this...

akx:
PyCharm isn't very fussy about that in particular, yeah :)

PR #176, opened 2025-04-25
Remove requirements.txt

filipstrand:
@anthonywu Do you think it is safe to remove this?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,0 @@
-# direct habitual `pip install -r requirements.txt`
-# users to `pyproject.toml` defined req list
--e .


anthonywu:
Safe to remove, no real loss, with caveats.

The `requirements.txt` serves a slightly different purpose. Whereas the `pyproject.toml` constrains allowable ranges of dependencies, usually `requirements.txt` serves as a snapshot in time of a set of completely resolved dependencies including the libraries not explicitly listed in `pyproject.toml`. It's useful in cases where we send a python app to production and you want that production app to pin literally every library. Or to have a `requirements-test.txt` that represents the exact versions for reproducible CI testing.

In this case, `mflux` is a library and not the application at the top layer, so we can defer requirements locking to whoever is deploying or wrapping mflux.

filipstrand:
make sense, thanks for clarifying 👍

PR #183, opened 2025-05-01
In-context: allow lora composition, add option to save full image

anthonywu:
This appears to be an oversight that was merged, but you probably were planning to come back to this.

The in-context generator can and should allow for mix/matching the usual LoRAs with the named in-context special LoRAs.

Also proposing a opt-in `--save-full-image` arg that stores additional copy of the output side by side with the reference image. Basically, no crop before output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 .env.production.local
 
 .venv
+*.lock  # created by uv sync
 *.png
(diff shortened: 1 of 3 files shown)


filipstrand:
Great! I'll include make a new release with this included later today

PR #184, opened 2025-05-08
Add upscaling support

filipstrand:
Under the hood, this is simply the controlnet pipeline with [jasperai/Flux.1-dev-Controlnet-Upscaler](https://huggingface.co/spaces/jasperai/Flux.1-dev-Controlnet-Upscaler), with a small modification that we don't process the with canny edge detection

Screenshot 2025-05-08 at 19 18 00
Currently, there is something a bit off with `ControlnetUtil.encode_image` (it is more apparent on some images than others),  however, when using an encoded image from the diffusers codebase in MFLUX, the results look good. To be updated...**Update** Fixed

---

On my machine (32GB) I have noticed that the image generation can progress as normal only to fail at the very end when decoding the latent in the VAE. Some quick vibe coding with Cursor suggested chunking up the latent into parts - this actually works quite well, and with some trial and error I found that `UpBlock3` was the bottleneck (causing out of memory errors before) so there is now a new flag `--vae-chunking` which the latents into 2 (top/bottom) and it works for bigger resolutions (maybe a more scalable way is to actually let the user define the chunking). This can cause small visual artifacts as a horizontal line (typically very noticeable in my testing, but would be interesting if there was a better way)

Another thing I noticed is that doing the upscaling for high resolutions (e.g 2400x2400 which now works on my machine) is that the regular shifted time-steps of dev become very heavily skewed towards early time-steps - for example, 

Not adjusted (linear spacing, like with the schnell model)
`array([1, 0.95, 0.9, ..., 0.1, 0.05, 0], dtype=float32)`

Adjusted with the current logic that is standard for Flux dev, for resolution 2400x2400:
`array([1, 0.999261, 0.998441, ..., 0.887757, 0.789318, 0], dtype=float32)`

In the the last 2 steps we go from 0.789318 -> 0 which results in poor quality. For this reason I have now defaulted to using the (non-adjusted) linear spacing, but not sure this is the correct way to do it. 

As a comparison, this is the sigmas for the adjusted ones, but for 512x512 

```
array([1, 0.972733, 0.944129, ..., 0.172612, 0.0899342, 0], dtype=float32)
```

At first I was a bit surprised, but as the interface to the function suggests, `RuntimeConfig._shift_sigmas(sigmas=sigmas, width=config.width, height=config.height)` the resolution actually enters here, so it might be reasonable. Could be worth exploring different schedulers for this functionality also!  

Here is Claude drawing the shifted sigma-series for different square resolutions: 
Screenshot 2025-05-08 at 17 07 12


src/mflux/config/model_config.py
```diff
@@ -148,10 +167,33 @@ def from_name(
...
         model_name="black-forest-labs/FLUX.1-Redux-dev",
         base_model=None,
+        controlnet_model=None,
         num_train_steps=1000,
         max_sequence_length=512,
         supports_guidance=True,
         requires_sigma_shift=True,
         priority=3,
     ),
+    "dev-controlnet-canny": ModelConfig(
```
filipstrand, line 177: This refactor might couple the base-model and controlnet model too much (e.g this current setup would not work with a custom dev model now - but I am not sure if a given controlnet trained model is expected to work with an arbitrary dev-like model anyways...). I could see an argument for keeping the controlnet and base model more separate to allow users to mix and match more freely also...

azrahello:
Happy to see you’re considering this possibility! In ComfyUI, the VAE can be set to ‘tiled’ and it works more or less like what you’ve done. It could also be useful for encoding in inpaint scenarios.”
https://github.com/comfyanonymous/ComfyUI/blob/924d771e18000f4cb223575189daa6d2c6c5a9c1/comfy/sd.py#L500 
I actually wanted to propose a ‘tiled VAE’ earlier, especially when testing with visions around versions 0.5 or 0.6, where towards the end there was a tendency to swap.
Another possible approach could be to split the image into smaller sections, pass each portion to the upscaler individually, and then reassemble everything afterwards.”

filipstrand:
Thanks for the resource!  Yes that would be another possibility. One benefit of that would be that the latents themselves should be smaller for each individual section, but I am not sure how it would work in practise and if there would be inconsistencies in terms of colors when finally joining the results etc... (I could imagine that using global information, i.e seeing the whole image/latent at once, might help the model). 

Also, I noticed that prompting helps a lot for the upscaling quality - before I just went with the "null prompt" `""` as in the diffusers example for the jasper upscaler, but specific prompting enhances the clarity *a lot*. I'll put this into the documentation for this feature once it is ready.

Very excited to see your redux PR! I spent about a day trying to play around with how to better control the influence of the multiple images in redux (tried adding specific attention masks so different reference images could attend to different parts of the prompt), but it ended up not working that well so I abandoned that idea for now.

src/mflux/controlnet/controlnet_util.py
```diff
@@ -18,15 +18,18 @@ def encode_image(
...
 
         control_image = ImageUtil.load_image(controlnet_image_path)
         control_image = ControlnetUtil._scale_image(height=height, width=width, img=control_image)
-        control_image = ControlnetUtil._preprocess_canny(control_image)
+        if is_canny:
+            control_image = ControlnetUtil._preprocess_canny(control_image)
         controlnet_cond = ImageUtil.to_array(control_image)
         controlnet_cond = vae.encode(controlnet_cond)
-        controlnet_cond = (controlnet_cond / vae.scaling_factor) + vae.shift_factor
+        if is_canny:
```
filipstrand, line 31: Note to self: Must investigate this... (current tests for canny functionality assumes this)

PR #185, opened 2025-05-09
Add support for --prompt-file for dynamic prompt updates in large batches

anthonywu:
Add support for `--prompt-file` for dynamic prompt updates.

This is something that solves my own use case. I am betting others will find this helpful in batch workflows too.

This PR adds support for reading prompts from files during image generation, allowing prompts to be updated between generations without restarting.

## Changes

- Added `--prompt-file` option as an alternative to `--prompt` for providing generation prompts
- Made `--prompt` and `--prompt-file` mutually exclusive options in the command line parser
- Added `mflux.ui.prompt_utils` helper to read prompt file content with appropriate error handling
- Implemented dynamic prompt reading before each seed iteration in all `generate_*` scripts
- Added two new tests to verify prompt-file functionality

## Use Cases

This feature is particularly useful when running batch generations with many seeds where you want to fine-tune the prompt between iterations without interrupting or restarting the process.

For example, when generating a series of images with `--auto-seeds 100`, you can edit the prompt file between generations to make minor adjustments based on observing incoming results, while keeping all other generation parameters constant. A trivial example is changing a color name in the prompt.

## Testing

Added tests to verify:
- Prompt files can be correctly specified and parsed
- Using both `--prompt` and `--prompt-file` simultaneously raises an appropriate error

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,11 +228,50 @@ The `mflux-generate-controlnet` command supports most of the same arguments as `
 See the [Controlnet](#%EF%B8%8F-controlnet) section for more details on how to use this feature effectively.
 
 
-#### 📜 Batch Image Generation Arguments
+#### Dynamic Prompts with `--prompt-file`
 
-- **`--prompts-file`** (required, `str`): Local path for a file that holds a batch of prompts.
+MFlux supports dynamic prompt updates through the `--prompt-file` option. Instead of providing a fixed prompt with `--prompt`, you can specify a plain text file containing your prompt. The file is re-read before each generation, allowing you to modify prompts between iterations without restarting.
 
-- **`--global-seed`** (optional, `int`): Entropy Seed used for all prompts in the batch.
+##### Key Benefits
+
+- **Real-time Prompt Editing**: Edit prompts between generations while the program continues running
+- **Iterative Refinement**: Fine-tune prompts based on previous results in a batch
+- **Editor Integration**: Work with complex prompts in your preferred text editor
+- **Consistent Parameters**: Experiment with prompt variations while keeping all other settings constant
+
+##### Example Usage
+
+```bash
+# Generate 10 images using a prompt from a file
+mflux-generate --prompt-file my_prompt.txt --auto-seeds 10
+```
+
+##### Workflow Example
+
+1. Create a prompt file:
+   ```
+   echo "a surreal landscape with floating islands" > my_prompt.txt
+   ```
+
+2. Start a batch generation:
+   ```
+   mflux-generate --prompt-file my_prompt.txt --auto-seeds 10
+   ```
+
+3. After reviewing initial results, edit the prompt file while generation continues:
+   ```
+   echo "a surreal landscape with floating islands and waterfalls" > my_prompt.txt
+   ```
+
+4. Subsequent generations will automatically use your updated prompt
+
+5. Continue refining as needed between generations
+
+##### Notes
+
+- `--prompt` and `--prompt-file` are mutually exclusive options
+- Empty prompt files or non-existent files will raise appropriate errors
+- Each generated image's metadata will contain the actual prompt used for that specific generation
 
 #### 📜 Training Arguments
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -228,11 +228,50 @@ The `mflux-generate-controlnet` command supports most of the same arguments as `
 See the [Controlnet](#%EF%B8%8F-controlnet) section for more details on how to use this feature effectively.
 
 
-#### 📜 Batch Image Generation Arguments
+#### Dynamic Prompts with `--prompt-file`
 
-- **`--prompts-file`** (required, `str`): Local path for a file that holds a batch of prompts.
+MFlux supports dynamic prompt updates through the `--prompt-file` option. Instead of providing a fixed prompt with `--prompt`, you can specify a plain text file containing your prompt. The file is re-read before each generation, allowing you to modify prompts between iterations without restarting.
 
-- **`--global-seed`** (optional, `int`): Entropy Seed used for all prompts in the batch.
```
anthonywu, line 235: these were never implemented, just a parser placeholder `add_batch_image_generator_arguments` referenced these, removing for now to avoid support questions

src/mflux/ui/cli/parsers.py
```diff
@@ -84,7 +86,7 @@ def add_image_to_image_arguments(self, required=False) -> None:
         self.add_argument("--image-strength", type=float, required=False, default=ui_defaults.IMAGE_STRENGTH, help=f"Controls how strongly the init image influences the output image. A value of 0.0 means no influence. (Default is {ui_defaults.IMAGE_STRENGTH})")
 
     def add_batch_image_generator_arguments(self) -> None:
-        self.add_argument("--prompts-file", type=Path, required=True, default=argparse.SUPPRESS, help="Local path for a file that holds a batch of prompts.")
+        self.add_argument("--batch-prompts-file", type=Path, required=True, default=argparse.SUPPRESS, help="Local path for a file that holds a batch of prompts.")
```
anthonywu, line 89: never implemented, okay to update/reserve for future

anthonywu (commented):
This PR done with help from Sonnet 3.7 Thinking with very explicit directions.

src/mflux/generate.py
```diff
@@ -41,10 +43,15 @@ def main():
 
     try:
         for seed in args.seed:
+            # Get prompt from file if prompt_file is provided, otherwise use the static prompt
```
filipstrand, line 46: The other files use

`prompt = get_effective_prompt(args)`

should it be the same for `generate.py` too?

filipstrand:
Like this feature!

src/mflux/ui/prompt_utils.py
```diff
@@ -0,0 +1,57 @@
...
+            raise PromptFileReadError(f"Prompt file is empty: {prompt_file_path}")
+
+        print(f"Using prompt from file: {prompt_file_path}")
+        return content
+    except (IOError, OSError) as e:
+        raise PromptFileReadError(f"Error reading prompt file '{prompt_file_path}': {e}")
+
+
+def get_effective_prompt(args: t.Any) -> str:
+    """
```
filipstrand, line 39: Regarding comments - would be nice to hear your opinions about this in general and compared with your previous experiences and typical python conventions: Personally, I feel like a lot of vibe coding output really prefer extensive docstrings for every method they generate, even though the implementation is small and the method-name mostly explains it. 

Maybe this is very common in a lot of python packages? To me, it often feels a bit redundant/unnecessary, and I typically default to not reading them unless I get stuck.  

I'm thinking of putting in some efforts into defining cursor rules etc that could help guide vibe-coding approaches but if my personal preference here goes against conventional wisdom too much (and, who knows, maybe the docstrings helps Agents understand the codebase better) then we probably shouldn't enforce such rules too hard.

anthonywu: I've gone back-and-forth on this throughout my career. I think this comes down to what our goals are and who the audience is. For 
comments on a tight knit team where everyone is expected to be at a minimum coding level, we can agree to only document beyond a particular difficulty. For open source projects where you want to be very welcoming to outside contributors, you may have to become more inclusive with comments. And in this age of AI assistant coding, we have to consider that AI assistants are part of this audience.

I agree that when method names and the comment strings largely overlap, we can go easy on the comment strings. Feel free to delete unnecessary text and merge this branch at your leisure.

PR #188, opened 2025-05-10
Add support for depth-guided training

filipstrand:
Have not tried this yet so this will have to be evaluated before merged as a proper feature, but especially for subject-related LoRAs it feels like a lot of training could be wasted on modelling unnecessary details often present in the background. With DepthPro now natively supported, I thought scaling the loss according to the depth-map could potentially lead to faster convergence.

### DIFF
--- a/src/mflux/dreambooth/_example/train.json
+++ b/src/mflux/dreambooth/_example/train.json
@@ -44,23 +44,28 @@
       "images": [
         {
           "image": "01.jpeg",
-          "prompt": "photo of sks dog"
+          "prompt": "photo of sks dog",
+          "use_depth": true
         },
         {
           "image": "02.jpeg",
-          "prompt": "photo of sks dog"
+          "prompt": "photo of sks dog",
+          "use_depth": false
         },
         {
           "image": "03.jpeg",
-          "prompt": "photo of sks dog"
+          "prompt": "photo of sks dog",
+          "use_depth": false
         },
         {
           "image": "04.jpeg",
-          "prompt": "photo of sks dog"
+          "prompt": "photo of sks dog",
+          "use_depth": false
         },
         {
           "image": "05.jpeg",
-          "prompt": "photo of sks dog"
+          "prompt": "photo of sks dog",
+          "use_depth": false
         }
       ]
   }
--- a/src/mflux/dreambooth/dataset/dataset.py
+++ b/src/mflux/dreambooth/dataset/dataset.py
@@ -9,6 +9,7 @@ from mflux import Flux1, ImageUtil
 from mflux.dreambooth.dataset.batch import Example
 from mflux.dreambooth.dataset.dreambooth_preprocessing import DreamBoothPreProcessing
 from mflux.dreambooth.state.training_spec import ExampleSpec
+from mflux.models.depth_pro.depth_pro import DepthPro
 from mflux.post_processing.array_util import ArrayUtil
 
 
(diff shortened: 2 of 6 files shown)


src/mflux/dreambooth/_example/train.json
```diff
@@ -44,23 +44,28 @@
       "images": [
         {
           "image": "01.jpeg",
-          "prompt": "photo of sks dog"
+          "prompt": "photo of sks dog",
+          "use_depth": true
```
filipstrand, line 48: The idea is to specify `use_depth` on a per-example basis

src/mflux/dreambooth/dataset/dataset.py
```diff
@@ -61,19 +73,42 @@ def _create_examples(
...
-            mx.eval(pooled_prompt_embeds)
+            mx.eval(encoded_image, prompt_embeds, pooled_prompt_embeds)
+            if depth_map is not None:
+                mx.eval(depth_map)
 
         return examples
 
     @staticmethod
-    def _encode_image(vae: nn.Module, image_path: Path, width: int, height: int) -> mx.array:
+    def _create_and_encode_depth_map(
```
filipstrand, line 88: Depth map is created when the examples are generated before training

src/mflux/dreambooth/optimization/dreambooth_loss.py
```diff
@@ -62,4 +62,18 @@ def _single_example_loss(flux: Flux1, config: RuntimeConfig, example: Example, r
         )
 
         # Construct the loss (derivation in src/mflux/dreambooth/optimization/_loss_derivation)
-        return (clean_image + predicted_noise - pure_noise).square().mean()
+        pixel_losses = (clean_image + predicted_noise - pure_noise).square()
```
filipstrand, line 65: This loss kind of assumes that it make sense to simply encode the depth image and that the spatial information is more or less maintained in the lower resolution latent. (i.e this would not work if the latents were more abstract and did not retain any spatial layout)

src/mflux/dreambooth/optimization/dreambooth_loss.py
```diff
@@ -62,4 +62,18 @@ def _single_example_loss(flux: Flux1, config: RuntimeConfig, example: Example, r
...
+
+    @staticmethod
+    def scale_loss_with_depth_map(depth_map: mx.array, pixel_losses: mx.float16) -> mx.float16:
+        # 1. Normalize the depth map to [0, 1]
+        depth_map_normalized = (depth_map - mx.min(depth_map)) / (mx.max(depth_map) - mx.min(depth_map) + 1e-8)
+
+        # 2. Normalize weights to sum to 1
+        weights = depth_map_normalized / (mx.sum(depth_map_normalized) + 1e-8)
+        weighted_loss = mx.sum(pixel_losses * weights)
+        return weighted_loss
```
filipstrand, line 79: This loss kind of assumes that it make sense to simply encode the depth image and that the spatial information is more or less maintained in the lower resolution latent. (i.e this would not work if the latents were more abstract and did not retain any spatial layout)

anthonywu:
Looking forward to this feature, I definitely have observed unwanted artifacts that are unmistakably from the training set.

Related, the Mac Finder has the right-click remove background Service, which I think is exactly what you integrating as part of the workflow here. Really good addition.

filipstrand:
> Looking forward to this feature, I definitely have observed unwanted artifacts that are unmistakably from the training set.
> 
> Related, the Mac Finder has the right-click remove background Service, which I think is exactly what you integrating as part of the workflow here. Really good addition.

Thanks! Yes, now that we have the model in place I kinda want to find more use-cases for it :)  The model itself seem to perform really well in my testing (and is also really fast), handling hair and edges really well! I tried the depthAnything model before committing to this one (via the standard Pytorch implementation), but typically got a softer depth map with less detail.

Just got around to doing a training run today and my initial results were not that different from training without the depth map... Will do some more testing of this feature over the coming week. Maybe it is worth having some kind of scale that says how much of the depth area to reward/penalise in the loss (e.g to express that the "near-white" is a lot more important than grey etc. in a non-linear way)

filipstrand:
@claude Can you summarise what was done in d232f91b3612acc4df03358efa2abc7d79db54e9?

PR #189, opened 2025-05-12
🪫 Battery Saver - manage power over long batch operations

anthonywu:
When using `--auto-seeds` or any batch operation that can deplete your MacBook {Air,Pro} battery over time, add some reasonable defaults and user flags to boundary the image generation to prevent the laptop from undesirable shut downs. This is a useful boundary if you rely on your Mac being remotely accessible by ssh or Screen Sharing.

# reference

This is the battery charging history for a M4 Max running mflux continuously while plugged into a 95W charger.

image


tests/callbacks/test_battery_saver.py
```diff
@@ -0,0 +1,100 @@
...
+        # Assert the function properly extracted the battery percentage
+        assert percentage == 42
+
+
+def test_get_battery_percentage_no_match():
+    """Test that the function returns None when the output doesn't match the expected pattern."""
+    with patch("subprocess.run") as mock_run:
+        # Set up mock to return an output that doesn't match the expected pattern
+        mock_result = MagicMock()
+        mock_result.stdout = "Now drawing from 'AC Power'\n-InternalBattery-0 (id=1234567)    No battery installed"
```
anthonywu, line 32: may need someone with a Mac mini / studio to confirm - I am pretty sure "No battery installed" is in the string but the rest of the return status string is my guess.

src/mflux/ui/defaults.py
```diff
@@ -1,3 +1,4 @@
+BATTERY_PERCENTAGE_STOP_LIMIT = 5
```
anthonywu, line 1: I am open to suggestions to lower this to like 1 or 2, given we are adding the callback by default.

anthonywu (commented):
My observations on my M1 Max and M4 Max laptops, having run batches > 24 and > 48 hours.

1. on 100% GPU utilization and screen usage over a long batch run, a 85W or 96W battery charger will not keep up with energy consumption and slowly deplete the battery.
2. the biggest charger shipped with MBPs is 96W
3. most monitors provide power to the laptop around 90W (I've owned multiple Thunderbolt monitors (LG, Dell) popular with Mac users)

tests/callbacks/test_battery_saver.py
azrahello: i'have a Mac studio, (and a Macbook pro m1) ,  how can help?

anthonywu: Show me the output of terminal command "pmset -g batt" and paste me the exact strings. I'll add to unit test. In your case this feature should never activate for you.

azrahello: > Show me the output of terminal command "pmset -g batt" and paste me the exact strings. I'll add to unit test. In your case this feature should never activate for you.

(base) @Alessandros-Mac-Studio ~ % pmset -g batt 
Now drawing from 'AC Power'
(base)  @Alessandros-Mac-Studio ~ %

src/mflux/callbacks/instances/battery_saver.py
```diff
@@ -0,0 +1,63 @@
...
+PMSET_BATT_STATUS_PATTERN = r"InternalBattery-.+?(\d+)%"
+
+logger = logging.getLogger(__name__)
+
+
+def _get_machine_model() -> str:
+    """Get the Mac machine model using system_profiler."""
+    try:
+        result = subprocess.run(
+            ["system_profiler", "-json", "SPHardwareDataType"], capture_output=True, text=True, check=True
```
anthonywu, line 20: profiler data may be interesting going forward for creating a benchmarking script so that we can all share run stats:

Here's mine with relevant fields:

```json
{
  "SPHardwareDataType" : [
    {
      "chip_type" : "Apple M1 Max",
      "machine_model" : "MacBookPro18,4",
      "machine_name" : "MacBook Pro",
      "number_processors" : "proc 10:8:2",
      "physical_memory" : "64 GB",
    }
  ]
}
```

src/mflux/generate.py
```diff
@@ -25,7 +26,15 @@ def main():
         lora_scales=args.lora_scales,
     )
 
-    # 2. Register the optional callbacks
+    # 2. Register callbacks
+
+    # Add BatterySaver callback unconditionally
```
anthonywu, line 31: "unconditionally" but should be no-cost to desktops, see latest commit on excluding AC powered machines from the check.

filipstrand:
Interesting! Never thought of adding this as a feature but it makes much sense to have (I am running MacBooks exclusivley and running out of juice has definitely happened hehe).

src/mflux/ui/defaults.py
filipstrand: I think this works fine! On older machines (e.g my older pre apple-silicon Macs) I noticed that as they get older they might report a 7% battery status, then suddenly die within the next 2 minutes or so if under any kind of load, so it might actually be good to have some extra margin here.

src/mflux/generate.py
filipstrand: Nice!

filipstrand:
Now that we have an increasing number of callbacks being registered at startup, I moved these to a separate method (still duplicated across the different entry-point files for now).

src/mflux/callbacks/instances/battery_saver.py
filipstrand: Good idea! Right now I think we only mention the command `system_profiler SPHardwareDataType SPDisplaysDataType` in the readme for the benchmark section, but it would be convenient to have a separate script that does all of this at once (including the generation itself)

filipstrand (approved):
Looks good!

PR #190, opened 2025-05-15
Possibility to balance the Redux function

azrahello:
What I’m trying to do is find a compromise between the prompt and the image weight in the Redux function. For values greater than 0.1, the prompt seems to have little to no influence, while below 0.1, it starts to have a noticeable effect. Ideally, we should find a formula such that at a value of 0.5, both the prompt and the image are weighted equally. (Suggestions are welcome!)”
image3
image
The parameters I used were the default ones — the only thing I changed was the redux-image-strength factor.

```
{
    "mflux_version": "0.7.0",
    "model": "black-forest-labs/FLUX.1-dev",
    "base_model": "None",
    "seed": 1010,
    "steps": 13,
    "guidance": 2.5,
    "precision": "mlx.core.bfloat16",
    "quantize": null,
    "generation_time_seconds": 113.77,
    "lora_paths": [
        "/Volumes/NewHome/comfy_new/ComfyUI/models/loras/FLUX/diffusion_pytorch_model.safetensors"
    ],
    "lora_scales": [
        1.0
    ],
    "image_path": null,
    "image_strength": null,
    "controlnet_image_path": null,
    "controlnet_strength": null,
    "masked_image_path": null,
    "depth_image_path": null,
    "redux_image_paths": "[PosixPath('robot.webp')]",
    "redux_image_strengths": [
        0.01
    ],
    "prompt": "yellow robot"
}
```

The method I implemented is a bit rough (and probably not very elegant). Also, the way ComfyUI and MFLUX interpret and weigh prompts seems to differ — but I’m not sure where to intervene in the code to better align them. It’s as if they give different importance to certain image features. For example, ComfyUI seems to recognize the subject more clearly, while in some tests I ran on MFLUX, the background — which should have been secondary — had a much stronger influence on the result. It’s a possible direction, but it still needs refinement.

### DIFF
--- a/src/mflux/generate_redux.py
+++ b/src/mflux/generate_redux.py
@@ -47,7 +47,10 @@ def main():
             args.redux_image_strengths.extend([1.0] * (len(args.redux_image_paths) - len(args.redux_image_strengths)))
         # If too many strengths provided, truncate to match image count
         elif len(args.redux_image_strengths) > len(args.redux_image_paths):
-            args.redux_image_strengths = args.redux_image_strengths[: len(args.redux_image_paths)]
+            raise ValueError(
+                f"Too many strengths provided ({len(args.redux_image_strengths)}), "
+                f"expted at most {len(args.redux_image_paths)}."
+            )
 
     try:
         for seed in args.seed:
--- a/src/mflux/generate_redux.py
+++ b/src/mflux/generate_redux.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 from mflux import Config, ModelConfig, StopImageGenerationException
 from mflux.callbacks.callback_registry import CallbackRegistry
 from mflux.callbacks.instances.memory_saver import MemorySaver
@@ -16,6 +18,12 @@ def main():
     parser.add_redux_arguments()
     parser.add_output_arguments()
     args = parser.parse_args()
+
+    # Validate and normalize redux image strengths
+    redux_image_strengths = _validate_redux_image_strengths(
+        redux_image_paths=args.redux_image_paths,
+        redux_image_strengths=args.redux_image_strengths,
+    )
 
     # 1. Load the model
     flux = Flux1Redux(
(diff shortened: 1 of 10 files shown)


src/mflux/generate_redux.py
```diff
@@ -40,6 +40,15 @@ def main():
...
         CallbackRegistry.register_after_loop(memory_saver)
 
+    # Validate redux_image_strengths if provided
+    if args.redux_image_strengths and len(args.redux_image_strengths) > 0:
+        # If strengths provided but not enough for all images, fill with default (1.0)
+        if len(args.redux_image_strengths) < len(args.redux_image_paths):
+            args.redux_image_strengths.extend([1.0] * (len(args.redux_image_paths) - len(args.redux_image_strengths)))
+        # If too many strengths provided, truncate to match image count
+        elif len(args.redux_image_strengths) > len(args.redux_image_paths):
+            args.redux_image_strengths = args.redux_image_strengths[: len(args.redux_image_paths)]
```
anthonywu, line 50: I would vote to raise an error for "too many strengths provided" and let the user fix the length mismatch.

The Zen of Python advocates not doing magical behaviors.

anthonywu (approved):
I think the implementation follows existing repo style and should work. Only minor suggestions.

src/mflux/generate_redux.py
azrahello: I believe I followed the suggestion you gave me. Sorry, I’m not a programmer and I haven’t touched code in years — Python is quite new to me. Thank you for the advice and for pointing it out!

anthonywu (commented):
@filipstrand  subject to manual verification of the workflow (which I have not gotten to yet), I think this is something worth merging.

filipstrand:
Great to see this! I've been quite busy this previous week so did not have time to review it sooner. I added a few minor refactors as well as some updates to the readme (saw that I had forgotten to add a section in the parameters for the flux tools)

anthonywu:
I did a few quick runs of `--redux-image-strength` variations, the change seems to have the desired effect from the change.

PR #203, opened 2025-06-05
Generalise in-context functionality 

filipstrand:


filipstrand:
@claude can you look through the commits on this branch back to release 0.7.1 and see what features we added since then and by which contributors. After that please make sure the release notes updated accordingly in the Changelog file. Follow the format from previous releases

src/mflux/generate_in_context_catvton.py
```diff
@@ -20,6 +20,10 @@ def main():
     parser.add_output_arguments()
     args = parser.parse_args()
 
+    # 0. Default to a higher guidance value for fill related tasks.
+    if args.guidance is None:
+        args.guidance = 30
```
anthonywu: recommend: put `30` in `defaults.py` as `DEFAULT_DEV_FILL_GUIDANCE`

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+          python -m pip install requests
+          
+      - name: Run release script
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ github.event.inputs.version }}
+          TEST_PYPI_API_TOKEN: ${{ secrets.TEST_PYPI_API_TOKEN }}
+          PYPI_API_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
+        run: |
+          python scripts/release.py --version ${{ github.event.inputs.version }}
```
anthonywu: nit: missing newline at EOF would invite formatters to update this later

scripts/release.py
```diff
@@ -0,0 +1,288 @@
...
+
+        # Set environment variables for subprocess
+        cmd_env = os.environ.copy()
+        cmd_env.update(env)
+
+        print("🔄 Publishing to PyPI...")
+        print(f"   Running: {sys.executable} -m twine upload dist/* --verbose")
+
+        result = subprocess.run(
+            [sys.executable, "-m", "twine", "upload", "dist/*", "--verbose"],
```
anthonywu, line 159: tool opinion: PyPA's `hatch` is really great: https://hatch.pypa.io/latest/

It won't be too much work to convert mflux to Hatch project type later, but just wanted to note Hatch is my own personal recommendation since 2023.

src/mflux/generate_in_context_catvton.py
filipstrand: This was actually a big thing I forgot during this refactor at first - For for flux fill, the guidance should be set to something like 30 otherwise results can become pretty bad, will fix this later today!

scripts/release.py
filipstrand: Interesting, never heard of it. Will look into it when I have the time

filipstrand:
@anthonywu Thanks for the review, I'll fix some things soon when I get the time. This would be the last thing before the 0.8.0 release. The idea was also to prepare a bit for Flux Kontext[dev] which looks like it is using a similar technique (though I have not seen any such source code yet). I think I wrote this somewhere in the readme/release notes, but it could very well be the case that the official Kontext feature outperforms both Catvton and ICEdit, so it might not be worth keeping those around later if that is the case). Also tried contacting BFL regarding early beta-testing of Kontext[dev] but heard nothing so far...

.github/workflows/release.yml
filipstrand: Lets see if this works :) 

@claude can you address this comment?

PR #204, opened 2025-06-09
Support Prompt via stdin. e.g. LLMs can generate prompts that directly feed into mflux.

anthonywu:
# Add stdin support for prompt input

## Summary

This PR adds support for reading prompts from stdin when `--prompt -` is specified, following Unix/Linux conventions. This enables piping prompts from other commands, making mflux more composable in shell scripts and workflows.

The big win is: this allows integration with any LLM text generator, without committing the mflux project to have an opinion or any dependencies on the wide variety of LLM generator CLI tools available. IMO this is the best way to be widely compatible without dependency creep. (e.g. I considered adding `ollama` and `mlx_lm` and `llm` as dependencies and handling it in the library, but when do we stop adding (N+1)th generator?)

### Feature Implementation
- **Added stdin support in `prompt_utils.py`**: When `--prompt -` is provided, the prompt is read from stdin
- **Follows Unix conventions**: Uses `-` as the special marker for stdin input, consistent with tools like `cat`, `grep`, etc.
- **Proper error handling**: Raises `PromptFileReadError` with clear message when stdin is empty
- **Preserves existing behavior**: Regular prompts and `--prompt-file` continue to work as before

### Code Improvements
- **Converted print statements to logging**: Changed `print()` calls to `logger.info()` for better control over output
- **Improved type annotations**: Changed `get_effective_prompt(args: t.Any)` to use proper `argparse.Namespace` type
- **Added comprehensive tests**: Both unit tests and integration tests verify the functionality

## Usage Examples

```bash
# Pipe prompt from echo
echo "A beautiful sunset over mountains" | \
    mflux-generate --prompt - --model dev --steps 20

# Pipe prompt from a file
cat prompt.txt | \
    mflux-generate --prompt - --model dev --output result.png

# Use in a pipeline
generate-prompt-script.py | \
    mflux-generate --prompt - --model dev --metadata

# MLX LM as a good MLX partner in the same venv
mlx_lm.generate --prompt "A tired hiker on Mt Everest" | \
    mflux-generate --prompt - --model dev

# Ollama as the prompt feeder
ollama run gemma3:12b "describe the photo composition of a perfect sunset, be concise just respond with the scene and camera/lens equipment used" | \
    mflux-generate --prompt - --model schnell

# the popular python 'llm' library as a feeder
llm -m gpt-4o-mini "describe the perfect breakfast for a track and field champion" | \
    mflux-generate --prompt - --model dev
```

### Unit Tests (`tests/arg_parser/test_stdin_prompt.py`)
- ✅ Basic stdin parsing (`--prompt -` returns "-")
- ✅ Regular prompt still works unchanged
- ✅ Whitespace trimming from stdin
- ✅ `--prompt-file` takes precedence over stdin

### Integration Tests (`tests/image_generation/test_stdin_prompt_integration.py`)
- ✅ Single-line stdin prompt generates image with correct metadata
- ✅ Multi-line stdin prompt is preserved correctly
- ✅ Empty stdin shows appropriate error message
- ✅ Shell piping with echo command works correctly

All tests verify that the stdin prompt is correctly captured in the output metadata when `--metadata` flag is used.

## Technical Details

- Stdin reading happens in `get_effective_prompt()` when `args.prompt == "-"`
- Uses `sys.stdin.read().strip()` to read and clean the input
- Logging messages indicate when prompt is read from stdin vs file
- Error handling for `IOError`, `OSError`, and `KeyboardInterrupt`

## Breaking Changes

None. This is a purely additive feature that doesn't affect existing functionality.

## Future Considerations

- we can support accepting stdin of image data as well for the various image args, but we would need to enforce xor usage of the `-` marker in only one of the args, that is we should not accept stdin for the prompt and image at the same time

### DIFF
--- a/ignore_pr.md
+++ b/ignore_pr.md
@@ -1,0 +1,62 @@
+# Add stdin support for prompt input
+
+## Summary
+
+This PR adds support for reading prompts from stdin when `--prompt -` is specified, following Unix/Linux conventions. This enables piping prompts from other commands, making mflux more composable in shell scripts and workflows.
+
+## Changes
+
+### Feature Implementation
+- **Added stdin support in `prompt_utils.py`**: When `--prompt -` is provided, the prompt is read from stdin
+- **Follows Unix conventions**: Uses `-` as the special marker for stdin input, consistent with tools like `cat`, `grep`, etc.
+- **Proper error handling**: Raises `PromptFileReadError` with clear message when stdin is empty
+- **Preserves existing behavior**: Regular prompts and `--prompt-file` continue to work as before
+
+### Code Improvements
+- **Converted print statements to logging**: Changed `print()` calls to `logger.info()` for better control over output
+- **Improved type annotations**: Changed `get_effective_prompt(args: t.Any)` to use proper `argparse.Namespace` type
+- **Added comprehensive tests**: Both unit tests and integration tests verify the functionality
+
+## Usage Examples
+
+```bash
+# Pipe prompt from echo
+echo "A beautiful sunset over mountains" | mflux-generate --prompt - --model dev --steps 20
+
+# Pipe prompt from a file
+cat prompt.txt | mflux-generate --prompt - --model dev --output result.png
+
+# Use in a pipeline
+generate-prompt-script.py | mflux-generate --prompt - --model dev --metadata
+```
+
+## Testing
+
+### Unit Tests (`tests/arg_parser/test_stdin_prompt.py`)
+- ✅ Basic stdin parsing (`--prompt -` returns "-")
+- ✅ Regular prompt still works unchanged
+- ✅ Whitespace trimming from stdin
+- ✅ `--prompt-file` takes precedence over stdin
+
+### Integration Tests (`tests/image_generation/test_stdin_prompt_integration.py`)
+- ✅ Single-line stdin prompt generates image with correct metadata
+- ✅ Multi-line stdin prompt is preserved correctly
+- ✅ Empty stdin shows appropriate error message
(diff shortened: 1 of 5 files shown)


tests/image_generation/test_stdin_prompt_integration.py
```diff
@@ -0,0 +1,148 @@
...
+
+import pytest
+
+
+@pytest.fixture
+def temp_output_dir(tmp_path) -> Path:
+    return tmp_path
+
+
+def test_stdin_prompt_with_actual_generation(temp_output_dir):
```
filipstrand, line 14: I really like this idea: very clean and general as you described! In terms of these tests, do we need the full image generation tests here ? I like that it is tested end-to-end, but at the same time the current image generation test-suite starts to build up in terms of how long it takes to run. If we keep these tests, maybe we should scale down the resolution to minimise time spent here? (and since they are computing 1 step, maybe it is not so bad)

anthonywu: I set the steps to 1 to run the whole flow but minimize how much work is done. Because the image output doesn't actually matter, my tests here focus on `--metadata` receiving the effective prompt so that images can be reproduced later from the data alone. I think the `--metadata` guarantee is important for compact storage/sharing of worthwhile generations.

ignore_pr.md
```diff
@@ -0,0 +1,62 @@
+# Add stdin support for prompt input
```
anthonywu, line 1: this was the AI agent local staging file for the PR summary, should not have been included! See #207 for quick fix.

PR #208, opened 2025-06-11
➡️➡️ Tab Tab for zsh: CLI completion generator

anthonywu:
# Status

- [x] Rebased and updated after merge of #203 which adds `in-context-{edit,catvton}` CLIs
- [x] No hurry to merge this for 0.8.* release, can be next release.

# Add ZSH shell completion support for mflux CLI commands

## Summary

- Adds `mflux-completions` command to generate and install ZSH completions for all mflux CLI tools
- Supports ZSH as it's the default shell on macOS, improving the developer experience for the primary target platform
- Implements dynamic completion generation by introspecting argparse parsers, ensuring completions stay in sync with code changes
- Provides intelligent completions for models, quantization levels, LoRA styles, file paths, and other command-specific options

## Details

This PR introduces comprehensive ZSH shell completion support for the mflux library's 13 CLI commands. The implementation uses a Python-based approach that dynamically generates completion scripts by introspecting the existing argparse parsers.

### Key Features:
- **New command**: `mflux-completions` - generates and installs ZSH completions
- **Smart completions**: Custom completion functions for models, quantization levels, and LoRA styles
- **Auto-sync**: Completions automatically reflect changes to CLI arguments without manual updates
- **User-friendly installation**: Auto-detects best completion directory or allows manual specification
- **Update support**: `--update` flag for safely updating existing completions
- **Verification**: `--check` flag to verify installation and troubleshoot issues
- **Documentation**: Comprehensive docs and quick start guide in README

### Usage:
```bash
# Install completions
mflux-completions

# Update existing completions
mflux-completions --update

# Print completion script to stdout
mflux-completions --print

# Install to specific directory
mflux-completions --dir ~/.config/zsh/completions
```

### Implementation:
- `src/mflux/completions/generator.py`: Core logic for generating ZSH completion scripts
- `src/mflux/completions/install.py`: Installation utility with user-friendly CLI
- Proper handling of mutually exclusive options, multiple values, and special characters in descriptions

This enhancement significantly improves the developer experience when using mflux CLI tools by providing context-aware tab completion for all commands and their options.

## Cost

```
> /cost
  ⎿  Total cost:            $15.02
     Total duration (API):  27m 47.5s
     Total duration (wall): 46m 14.6s
     Total code changes:    754 lines added, 29 lines removed
     Token usage by model:
         claude-3-5-haiku:  107.8k input, 4.5k output, 0 cache read, 0 cache write
              claude-opus:  240 input, 29.3k output, 5.7m cache read, 218.7k cache write
```

# Usage

Tab to complete long worded flags:

```sh
mflux-generate --bat<tab>
mflux-generate --battery-percentage-stop-limit
```
-----

No need to remember flag names: `mflux-generate-fill` can now tab to completion of its many flags.

```sh
$ mflux-generate-fill --<tab>
mflux-generate-fill --
--auto-seeds                     -- Auto generate N Entropy Seeds (random ints between 0 and 1 billion
--base-model                     -- When using a third-party huggingface model, explicitly specify whether the base model is dev or schnell
--battery-percentage-stop-limit  -- On Macs powered by battery, stop image generation when battery reaches this percentage. Default: 5
--guidance                       -- Guidance Scale (Default is 3.5)
--height                         -- Image height (Default is 1024)
--image-path                     -- Local path to the source image
--lora-name                      -- Name of the LoRA to download from Hugging Face
--lora-paths                     -- Local safetensors for applying LORA from disk
--lora-repo-id                   -- Hugging Face repository ID for LoRAs (default: ali-vilab/In-Context-LoRA)
--lora-scales                    -- Scaling factor to adjust the impact of LoRA weights on the model. A value of 1.0 applies the LoRA weights as they are.
--lora-style                     -- Style of the LoRA to use (e.g., storyboard for film storyboard style)
--low-ram                        -- Enable low-RAM mode to reduce memory usage (may impact performance).
--masked-image-path              -- Local path to the mask image
--metadata                       -- Export image metadata as a JSON file.
--model                          -- The model to use (dev or dev-fill or schnell or a compatible huggingface repo_id org/model).
--output                         -- The filename for the output image. Default is "image.png".
--path                           -- Local path for loading a model from disk
--prompt                         -- The textual description of the image to generate.
--prompt-file                    -- Path to a file containing the prompt text. The file will be re-read before each generation, allowing you to edit the pro
--quantize                       -- Quantize the model (3 or 4 or 6 or 8, Default is None)
--seed                           -- Specify 1+ Entropy Seeds (Default is 1 time-based random-seed)
--steps                          -- Inference Steps
--stepwise-image-output-dir      -- [EXPERIMENTAL] Output dir to write step-wise images and their final composite image to. This feature may change in futur
--vae-tiling                     -- Enable VAE tiling to reduce memory usage at the cost of a potential seam
--vae-tiling-split               -- Direction to split the latents when using VAE tiling (horizontal = top/bottom, vertical = left/right). Default is horizo
--width                          -- Image width (Default is 1024)
```

-----

New to a command, quick lookup!

```sh
mflux-train -<tab>
--base-model            -- When using a third-party huggingface model, explicitly specify whether the base model is dev or schnell
--model             -m  -- The model to use (dev or dev-fill or schnell or a compatible huggingface repo_id org/model).
--path                  -- Local path for loading a model from disk
--quantize          -q  -- Quantize the model (3 or 4 or 6 or 8, Default is None)
--train-checkpoint      -- Local path of the checkpoint file which specifies how to continue the training process
--train-config          -- Local path of the training configuration file
```

### DIFF
--- a/src/mflux/completions/generator.py
+++ b/src/mflux/completions/generator.py
@@ -3,7 +3,7 @@
 import argparse
 from pathlib import Path
 
-from mflux.community.in_context_lora.in_context_loras import LORA_NAME_MAP
+from mflux.community.in_context.utils.in_context_loras import LORA_NAME_MAP
 from mflux.ui import defaults as ui_defaults
 from mflux.ui.cli.parsers import CommandLineParser
 
@@ -16,6 +16,8 @@ class CompletionGenerator:
             "mflux-generate",
             "mflux-generate-controlnet",
             "mflux-generate-in-context",
+            "mflux-generate-in-context-edit",
+            "mflux-generate-in-context-catvton",
             "mflux-generate-fill",
             "mflux-generate-depth",
             "mflux-generate-redux",
@@ -57,6 +59,24 @@ class CompletionGenerator:
             parser.add_output_arguments()
             # Add save-full-image manually since it's not in parsers.py
             parser.add_argument("--save-full-image", action="store_true")
+
+        elif command == "mflux-generate-in-context-edit":
+            parser.add_general_arguments()
+            parser.add_model_arguments(require_model_arg=False)
+            parser.add_lora_arguments()
+            parser.add_image_generator_arguments(supports_metadata_config=False, require_prompt=False)
+            parser.add_in_context_edit_arguments()
+            parser.add_in_context_arguments()
+            parser.add_output_arguments()
+
+        elif command == "mflux-generate-in-context-catvton":
+            parser.add_general_arguments()
+            parser.add_model_arguments(require_model_arg=False)
+            parser.add_lora_arguments()
+            parser.add_image_generator_arguments(supports_metadata_config=False, require_prompt=False)
+            parser.add_catvton_arguments()
+            parser.add_in_context_arguments()
+            parser.add_output_arguments()
 
         elif command == "mflux-generate-fill":
             parser.add_general_arguments()
--- a/src/mflux/completions/generator.py
+++ b/src/mflux/completions/generator.py
@@ -1,0 +1,308 @@
+"""Generate ZSH completion scripts for mflux commands."""
+
+import argparse
+from pathlib import Path
+
+from mflux.community.in_context_lora.in_context_loras import LORA_NAME_MAP
+from mflux.ui import defaults as ui_defaults
+from mflux.ui.cli.parsers import CommandLineParser
+
+
+class CompletionGenerator:
+    """Generate ZSH completion scripts by introspecting argparse parsers."""
+
+    def __init__(self):
+        self.commands = [
+            "mflux-generate",
+            "mflux-generate-controlnet",
+            "mflux-generate-in-context",
+            "mflux-generate-fill",
+            "mflux-generate-depth",
+            "mflux-generate-redux",
+            "mflux-concept",
+            "mflux-concept-from-image",
+            "mflux-save",
+            "mflux-save-depth",
+            "mflux-train",
+            "mflux-upscale",
+            "mflux-lora-library",
+        ]
+
+    def create_parser_for_command(self, command: str) -> CommandLineParser:
+        """Create the appropriate parser for a given command."""
+        parser = CommandLineParser(prog=command, add_help=False)
+
+        if command == "mflux-generate":
+            parser.add_general_arguments()
+            parser.add_model_arguments()
+            parser.add_lora_arguments()
+            parser.add_image_generator_arguments(supports_metadata_config=True)
+            parser.add_image_to_image_arguments()
+            parser.add_image_outpaint_arguments()
+            parser.add_output_arguments()
+
+        elif command == "mflux-generate-controlnet":
+            parser.add_general_arguments()
+            parser.add_model_arguments()
+            parser.add_lora_arguments()
+            parser.add_controlnet_arguments()
+            parser.add_image_generator_arguments()
+            parser.add_output_arguments()
+
+        elif command == "mflux-generate-in-context":
+            parser.add_general_arguments()
+            parser.add_model_arguments()
+            parser.add_image_generator_arguments()
+            parser.add_image_to_image_arguments(required=True)
+            parser.add_output_arguments()
+            # Add save-full-image manually since it's not in parsers.py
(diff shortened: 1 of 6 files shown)


src/mflux/completions/install.py
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env python3
+"""Install ZSH completions for mflux commands."""
```
anthonywu, line 2: This file is entirely vibe coded, shell completions are a black art to me, I have to rely on eyeballing this code to seem reasonable, and actually using the result of the change to verify that tab-completion works as expected.

src/mflux/completions/install.py
```diff
@@ -0,0 +1,203 @@
...
+        print(f"✓ Installed ZSH completions to: {completion_file}")
+        print()
+        print("To activate completions:")
+        print("1. Ensure this directory is in your $fpath:")
+        print(f"   echo 'fpath=({install_dir} $fpath)' >> ~/.zshrc")
+        print("2. Reload your shell:")
+        print("   exec zsh")
+        print()
+        print("If completions don't work immediately, you may need to:")
+        print("   rm -f ~/.zcompdump && compinit")
```
anthonywu, line 195: verified this fixed a broken "can't complete" state for me

anthonywu (commented):
There is a downside to releasing this - some users will not be able to get their shell envs to work correctly and may file issues. Shell problems are so individualized and I think we should not support snowflakey issues that we cannot produce on "clean enough" zsh envs.

src/mflux/completions/generator.py
```diff
@@ -0,0 +1,328 @@
+"""Generate ZSH completion scripts for mflux commands."""
```
anthonywu, line 1: the `mflux.completions` scope is downstream of the actual generators and I would consider it "P2" in terms of importance to the library. It's a nice to have for terminal warriors like myself who mostly tab-tab through commands.

Also had considered hosting in a separate library, but decided to propose coupling here because:

1. no additional dependencies
2. module isolation (nothing else in the library should `import mflux.completions`
3. we update CLIs often, in-repo coupling would keep features and completions sync'ed every release

With AI agents, giving them a hint like this should help:

> when updating `mflux-` scripts and args during a change, make sure to update the corresponding `mflux/completions/generator.py` configuration to match

filipstrand:
> There is a downside to releasing this - some users will not be able to get their shell envs to work correctly and may file issues. Shell problems are so individualized and I think we should not support snowflakey issues that we cannot produce on "clean enough" zsh envs.

True, I can image people opening issues related to people not getting this to work on their systems. I had some trouble in the beginning myself, but letting cursor fix it for me (complementing my .zshrc with some fpath thing) after a couple of tries. After that it worked fine. But this is still a neat feature to have so I think we should include it! On the topic of the cli-args, I know there are some improvements that could be made on this front (e.g better errors - when not including the `--model` right now it gives a really ugly error message) but those have been lower on the priority list so far...

src/mflux/completions/generator.py
filipstrand: Agree with this decision. Even though I personally don't use the terminal that much and mostly use the python api myself, I think it's nice to have support for this (and my guess is that it could be improved further with some more effort spent on the cli-arg handling, e.g better error messages etc). 

On the slightly realted topic of prompting coding agents, I started to briefly look at cursor rules / claude.md / llm.txt again. Pretty soon it would be nice to have proper support for these as we've talked about before, and including things like what you mentioned here for example.  I briefly tried an experiment: To maintain a single "source of truth" description and then "compiling" it down to the various formats, like cursor rules, (by asking a LLM to do that step), but that worked OK at best. 

I feel like it would be doable (but not optimal) to manually maintain a few such documents (the most common ones) by having cursor/claude update the instructions across the various formats we support.   

 I'll try to get a PR up for this in the upcoming week after the 0.9 release and we could iterate on it.

filipstrand:
Don't have much to add for this PR (only added a small warning in the readme), it looks good and the feature is really nice to have. I'll probably use the CLI more after this, thanks for the initiative!

filipstrand:
Reminder to self to include kontext support after that is merged

PR #209, opened 2025-06-13
Small additions before 0.8.0

filipstrand:


anthonywu:
scope looks good to me, should not creep any more into 0.8.0

filipstrand:
> scope looks good to me, should not creep any more into 0.8.0

Agree! Will generate the latest pictures and descriptions for the readme, then everything should be good to go! Will try to wrap up the release today after work or tomorrow :)

PR #201, opened 2025-06-04
Cache Refactor - use platformdirs helper

anthonywu:
# Summary

- Implement platform-idiomatic cache directory management using `platformdirs` library (successor to the also-popular `appdirs` library, which is deprecated)
- Support custom cache locations via `MFLUX_CACHE_DIR` environment variable
- Add automatic migration from legacy (barely) `~/.cache/mflux` to new platform-specific locations with symlink for backward compatibility

Future scope:

- save photos to `platformdirs.user_pictures_dir()`
- version isolation with e.g. `PlatformDirs("mflux", version="1.0")
- see links to docs for any other system dirs we may want to use going forward

# Tests

1. I ran the `generate-depth` CLI and monitored the migration of old to new path
2. I put the old path back again and retried several times

Because we're using system level utilities to move the contents, IMO adding testing around that is superfluous. The affected users are the `mflux` users who just recently upgraded to the Depth/InContext features, and does not include lazy users who will jump several releases to this one.

# Links

1. https://platformdirs.readthedocs.io/en/latest/
2. https://pypi.org/project/platformdirs/

PR #215, opened 2025-06-20
🖼️↗️ Add scale factor support for image upscaling dimensions

anthonywu:
- [X] This is a WIP PR that follows the commit history of the `platformdirs` branch in unmerged #201 - for now the base branch for the PR is `platformdirs` but will rebase/switch to `main` for final review regardless of the outcome of #201

-----

# Add ScaleFactor Support to mflux-upscale

This PR adds flexible dimension scaling options to the `mflux-upscale` command, allowing users to specify output dimensions using scale factors (e.g., "2x", "3.5x"), "auto", or explicit pixel values. Additionally, it implements safety warnings for large image generations that may exceed system capabilities.

Introduces a ScaleFactor class and CLI parsing logic. Updates the upscale pipeline and argument parser to handle scale factors, including validation, pixel rounding, and user warnings for large outputs. Adds comprehensive tests for scale factor parsing, CLI integration, and correct dimension calculation.

Motivated by my usage where I was upscaling images while leaving `--height` and `--width` at default (resolved to 1024), but my images were often rectangular. IMO, software should do the thing that is most obvious and intuitive, which is to keep the resolution but upscale it as-is. This led me to consider the default should be "1x", but while doing this, I had the opportunity to also allow for intuitive up/downscaling via simple arg values.

### New Features

- **ScaleFactor Input Support**: Users can now specify dimensions using:
  - Scale factors: `--height 2x --width 2x` (doubles both dimensions)
  - Decimal scale factors: `--height 1.5x` (scales by 1.5x)
  - Auto mode: `--height auto` (preserves original dimension)
  - Plain integers: `--height 2048` (explicit pixel value)
  - Mixed modes: `--height 2x --width 1920` (scale height, fixed width)

- **Large Image Warning**: Added safety check when output dimensions exceed 2048×2048 pixels (4,194,304 total pixels)
  - Displays warning about potential memory/performance issues
  - Prompts user for confirmation before proceeding
  - User can press Enter to continue or 'n' to cancel

#### New Files

- `src/mflux/ui/scale_factor.py`: Core ScaleFactor class implementation
  - Parses scale factor strings (e.g., "2.5x")
  - Calculates scaled dimensions with pixel step quantization
  - Ensures dimensions are multiples of `DIMENSION_STEP_PIXELS` (16)

#### Modified Files

- `src/mflux/ui/cli/parsers.py`:
  - Added `int_or_special_value()` type converter
  - Prioritizes integer parsing over scale factor parsing
  - Supports "auto" keyword for maintaining original dimensions

- `src/mflux/upscale.py`:
  - Integrated ScaleFactor dimension calculation
  - Added dimension validation with user warning
  - Uses `sys.exit(1)` for idiomatic Python exits

- `src/mflux/ui/defaults.py`:
  - Added `MAX_PIXELS_WARNING_THRESHOLD = 2048 * 2048`

### Technical Decisions

1. **Quantization Formula**: 
   ```python
   int(self.value * orig_value - (self.value * orig_value) % pixel_steps)
   ```
   Ensures scaled dimensions are always multiples of 16 pixels for model compatibility.

2. **Parser Priority**: Integer parsing takes precedence over scale factor parsing to handle edge cases like "2" vs "2x".

3. **Regex Pattern**: Only accepts positive numbers: `r"^(\d+(?:\.\d+)?)x$"`

4. **Exit Strategy**: Uses `sys.exit(1)` instead of `exit()` for proper Python idioms.

## Testing

Comprehensive test coverage added:

- `tests/arg_parser/test_upscale_argparser.py`: Tests argument parsing
- `tests/ui/test_scale_factor.py`: Tests ScaleFactor class functionality
- `tests/image_generation/test_upscale_dimensions.py`: Tests dimension calculation
- `tests/image_generation/test_upscale_dimension_limits.py`: Tests warning system

All tests verify behavior without reimplementing logic, using mocks to check correct values are passed to underlying functions.

## Examples

```bash
# Scale image by 2x in both dimensions
mflux-upscale --prompt "beautiful landscape" --controlnet-image-path input.jpg --height 2x --width 2x

# Scale height by 1.5x, set width to 1920
mflux-upscale --prompt "portrait" --controlnet-image-path photo.jpg --height 1.5x --width 1920

# Keep original dimensions
mflux-upscale --prompt "enhanced" --controlnet-image-path original.png --height auto --width auto

# Warning example (4096x4096 = 16,777,216 pixels)
mflux-upscale --prompt "huge image" --controlnet-image-path small.jpg --height 4096 --width 4096
# ⚠️ WARNING: The requested dimensions 4096x4096 (16,777,216 pixels) exceed max recommended (4,194,304 pixels).
```

## Breaking Changes

None - the feature is backwards compatible. Users can still use integer values for dimensions as before.

## Future Considerations

- Could extend scale factor support to other image generation commands
- Might add preset scale factors (e.g., "hd", "4k", "8k")
- Could make the warning threshold configurable via environment variable

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,53 +12,37 @@ Run the powerful [FLUX](https://blackforestlabs.ai/#get-flux) models from [Black
 
 - [Philosophy](#philosophy)
 - [💿 Installation](#-installation)
-- [🚀 Shell Completions (Quick Start)](#-shell-completions-quick-start)
 - [🖼️ Generating an image](#%EF%B8%8F-generating-an-image)
   * [📜 Full list of Command-Line Arguments](#-full-list-of-command-line-arguments)
 - [⏱️ Image generation speed (updated)](#%EF%B8%8F-image-generation-speed-updated)
 - [↔️ Equivalent to Diffusers implementation](#%EF%B8%8F-equivalent-to-diffusers-implementation)
 - [🗜️ Quantization](#%EF%B8%8F-quantization)
-  * [📊 Size comparisons for quantized models](#-size-comparisons-for-quantized-models)
-  * [💾 Saving a quantized version to disk](#-saving-a-quantized-version-to-disk)
-  * [💽 Loading and running a quantized version from disk](#-loading-and-running-a-quantized-version-from-disk)
 - [💽 Running a non-quantized model directly from disk](#-running-a-non-quantized-model-directly-from-disk)
 - [🌐 Third-Party HuggingFace Model Support](#-third-party-huggingface-model-support)
 - [🎨 Image-to-Image](#-image-to-image)
 - [🔌 LoRA](#-lora)
-  * [Multi-LoRA](#multi-lora)
-  * [LoRA Library Path](#lora-library-path)
-  * [Supported LoRA formats (updated)](#supported-lora-formats-updated)
 - [🎭 In-Context Generation](#-in-context-generation)
   * [🎨 In-Context LoRA](#-in-context-lora)
-    + [Available Styles](#available-styles)
-    + [How It Works](#how-it-works)
-    + [Tips for Best Results](#tips-for-best-results)
   * [👕 CatVTON (Virtual Try-On)](#-catvton-virtual-try-on)
   * [✏️ IC-Edit (In-Context Editing)](#%EF%B8%8F-ic-edit-in-context-editing)
 - [🛠️ Flux Tools](#%EF%B8%8F-flux-tools)
   * [🖌️ Fill](#%EF%B8%8F-fill)
-    + [Inpainting](#inpainting)
-    + [Outpainting](#outpainting)
   * [🔍 Depth](#-depth)
   * [🔄 Redux](#-redux)
 - [🕹️ Controlnet](#%EF%B8%8F-controlnet)
 - [🔎 Upscale](#-upscale)
-- [🎛️ Dreambooth fine-tuning](#-dreambooth-fine-tuning)
-  * [Training configuration](#training-configuration)
-  * [Training example](#training-example)
-  * [Resuming a training run](#resuming-a-training-run)
-  * [Configuration details](#configuration-details)
-  * [Memory issues](#memory-issues)
-  * [Misc](#misc)
+- [🎛️ Dreambooth fine-tuning](#%EF%B8%8F-dreambooth-fine-tuning)
 - [🧠 Concept Attention](#-concept-attention)
 - [🚧 Current limitations](#-current-limitations)
 - [💡Workflow tips](#workflow-tips)
-- [🔬 Cool research / features to support](#-cool-research--features-to-support-)
+- [🔬 Cool research](#-cool-research)
 - [🌱‍ Related projects](#-related-projects)
 - [🙏 Acknowledgements](#-acknowledgements)
-- [⚖️ License](#-license)
+- [⚖️ License](#%EF%B8%8F-license)
 
 <!-- TOC end -->
+
+---
 
 ### Philosophy
 
@@ -71,6 +55,7 @@ All models are implemented from scratch in MLX and only the tokenizers are used 
 [Huggingface Transformers](https://github.com/huggingface/transformers) library. Other than that, there are only minimal dependencies
 like [Numpy](https://numpy.org) and [Pillow](https://pypi.org/project/pillow/) for simple image post-processing.
 
+---
 
 ### 💿 Installation
 For users, the easiest way to install MFLUX is to use `uv tool`: If you have [installed `uv`](https://github.com/astral-sh/uv?tab=readme-ov-file#installation), simply:
@@ -140,9 +125,10 @@ pip install -U mflux
 
 *If you have trouble installing MFLUX, please see the [installation related issues section](https://github.com/filipstrand/mflux/issues?q=is%3Aissue+install+).* 
 
-### ⌨️ Shell Completions (Quick Start)
+<details>
+<summary>⌨️ <strong>Shell Completions (Optional)</strong></summary>
 
-MFLUX supports ZSH (default on macOS) shell completions for all CLI commands.
+MFLUX supports ZSH (default on macOS) shell completions for all CLI commands. This provides tab completion for all [command-line arguments](#-full-list-of-command-line-arguments) and options.
 
 To enable completions:
 
@@ -164,6 +150,10 @@ mflux-completions --check
 ```
 
 For more details and troubleshooting, see the [completions documentation](src/mflux/completions/README.md).
+
+</details>
+
+---
 
 ### 🖼️ Generating an image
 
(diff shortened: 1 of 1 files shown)


anthonywu:
FYI: This PR was vibe coded with a very short leash. The agent made many mistakes that I had to reverse/undo.

```
> /cost
  ⎿  Total cost:            $31.99
     Total duration (API):  36m 37.0s
     Total duration (wall): 12h 33m 47.5s
     Total code changes:    1121 lines added, 474 lines removed
     Token usage by model:
         claude-3-5-haiku:  161.1k input, 7.4k output, 0 cache read, 0 cache write
              claude-opus:  275 input, 46.2k output, 10.8m cache read, 651.3k cache write
```

src/mflux/controlnet/controlnet_util.py
```diff
@@ -17,7 +17,7 @@ def encode_image(
         vae: VAE,
         height: int,
         width: int,
-        controlnet_image_path: str,
+        controlnet_image_path: PIL.Image.StrOrBytesPath,
```
anthonywu: TIL: `PIL.Image.StrOrBytesPath` is the built-in type hint we can borrow from the `PIL` library

src/mflux/controlnet/controlnet_util.py
```diff
@@ -34,15 +34,15 @@ def encode_image(
         return controlnet_cond, control_image
 
     @staticmethod
-    def _preprocess_canny(img: PIL.Image) -> PIL.Image:
+    def _preprocess_canny(img: PIL.Image.Image) -> PIL.Image.Image:
```
anthonywu, line 38: bug fix: `PIL.Image` is a module, `PIL.Image.Image` is the actual type/class we want for hinting

src/mflux/generate_controlnet.py
```diff
@@ -13,7 +13,7 @@ def main():
     parser.add_model_arguments(require_model_arg=True)
     parser.add_lora_arguments()
     parser.add_image_generator_arguments(supports_metadata_config=False)
-    parser.add_controlnet_arguments()
+    parser.add_controlnet_arguments(mode="canny")
```
anthonywu, line 16: dev fix: the canny options aren't applicable to all controlnet workflows anymore

src/mflux/post_processing/image_util.py
```diff
@@ -117,8 +117,11 @@ def to_array(image: PIL.Image.Image, is_mask: bool = False) -> mx.array:
         return array
 
     @staticmethod
-    def load_image(path: str | Path) -> PIL.Image.Image:
-        return PIL.Image.open(path).convert("RGB")
+    def load_image(image_or_path: PIL.Image.Image | PIL.Image.StrOrBytesPath) -> PIL.Image.Image:
+        if isinstance(image_or_path, PIL.Image.Image):
+            return image_or_path.convert("RGB")
```
anthonywu: This is a look forward setup for me to pipe in image data via stdin so in the near future not all images "loaded" will be a plain path/str.

src/mflux/ui/defaults.py
```diff
@@ -11,8 +11,10 @@
 CONTROLNET_STRENGTH = 0.4
 DEFAULT_DEV_FILL_GUIDANCE = 30
 DEFAULT_DEPTH_GUIDANCE = 10
+DIMENSION_STEP_PIXELS = 16
 GUIDANCE_SCALE = 3.5
 HEIGHT, WIDTH = 1024, 1024
+MAX_PIXELS_WARNING_THRESHOLD = 2048 * 2048
```
anthonywu, line 17: Flux.1 ultra/pro stops here, pretty sure this is the capability limit of 2025 Macs (without using distributed compute)

In any case, the workflows allows users to push past this value at their own risk.

tests/image_generation/test_upscale_dimensions.py
```diff
@@ -0,0 +1,70 @@
...
+    mock_image.height = orig_height
+    mock_image.width = orig_width
+
+    # Mock the flux object
+    mock_flux = Mock()
+
+    # Import and patch the actual upscale module
+    with patch("PIL.Image.open", return_value=mock_image):
+        with patch("mflux.upscale.Flux1Controlnet", return_value=mock_flux):
+            with patch("mflux.upscale.ModelConfig"):
```
anthonywu, line 37: the mocking of the flux object makes this test run quickly without actually evaluating any tensors, we just wanted to check arg passing is entirely handled at the controller layer (CLI) and not something the Flux-implementation has to worry about

src/mflux/ui/defaults.py
azrahello: to push this limit, it might be helpful to divide the image to be upscaled into smaller portions and iterate the process for each smaller portion. A more or less elegant solution would then need to be found to blend the intersection points and make them invisible. This way, I believe the sampling phase would be less demanding, and at the same time, resolutions suitable for potential printing could be achieved.

filipstrand: @azrahello I know you mentioned this before, I would also like to try this sometime and see how this works in practice. I gave this a short try before (very manually) by cropping the image and applying upscaling to each individual section (this experiment was for scaling up a catvton image with a special garment. The problem I got there was that the different sections that covered the garment ended up having slightly different hues/saturations, even though the seed/prompt was kept fixed for the different sections. This totally broke the illusion of having one big image after stitching it all together .

src/mflux/controlnet/controlnet_util.py
filipstrand: nice, was not aware of this either

src/mflux/controlnet/controlnet_util.py
filipstrand: Thanks for noting and fixing this! Reminds me when I used to have the parenthesis notation `(mx.array, mx.array)` as a return type thinking it was equivalent to the `tuple[mx.array, mx.array]`. I don't think about this much when coding unless my IDE gives a warning (which, I guess, it should have done in this kind of case?), so good that you point it out as it must look weird. I would really like to get all help I can get from tools to avoid constructing meaningless type hints like this without having to worry too much about it.

filipstrand:
FYI: Rebased this onto main and changed the base to main

filipstrand:
This kind of command is really convenient!

```
# Scale height by 1.5x, set width to 1920
mflux-upscale --prompt "portrait" --controlnet-image-path photo.jpg --height 1.5x --width 1920
```

filipstrand:
I have not had the time to try it yet, but maybe the kontext feature actually has the ability to upscale images too (simply by instructing it to do so, would be very cool if it worked - I just remembered seeing some nice examples of image restoration, so maybe it would actually work...). If that would work reliably, then maybe we could think about adding similar command line options there too

src/mflux/ui/defaults.py
azrahello: > @azrahello I know you mentioned this before, I would also like to try this sometime and see how this works in practice. I gave this a short try before (very manually) by cropping the image and applying upscaling to each individual section (this experiment was for scaling up a catvton image with a special garment. The problem I got there was that the different sections that covered the garment ended up having slightly different hues/saturations, even though the seed/prompt was kept fixed for the different sections. This totally broke the illusion of having one big image after stitching it all together .

This is the real issue to address. It can also happen in ComfyUI, but there are two possible solutions there.
The first is ensuring that ControlNet Upscale works throughout the entire process (even with low strength values).
The second, which I also found in ComfyUI, involves changing the shape of the tiles — using rectangular shapes instead of square ones — and splitting them properly based on resolution. I found a node that does exactly this, and with Flux it works well even without applying any upscaling.

In ComfyUI, when I use SDXL for upscaling, the image quality often degrades. With Flux, on the other hand, the quality is preserved — but in the meantime, I grow old. 😄


PS: To be fair, in addition to the upscale process, I’m also using something available here:
👉 https://openmodeldb.info/
The node I’m referring to is:
👉 https://github.com/ssitu/ComfyUI_UltimateSDUpscale

PR #219, opened 2025-06-26
Add Flux1.Kontext[dev] support

filipstrand:
Almost completely vibe coded but seems to work with a first initial test :)

`--image-path "cat.png" --guidance 2.5 --prompt "Add a hat to the cat" -q 8 --height 352 --width 512 --steps 20`

cat
image


src/mflux/flux_tools/kontext/flux_kontext.py
```diff
@@ -0,0 +1,158 @@
...
+from mflux.models.text_encoder.prompt_encoder import PromptEncoder
+from mflux.models.text_encoder.t5_encoder.t5_encoder import T5Encoder
+from mflux.models.transformer.transformer import Transformer
+from mflux.models.vae.vae import VAE
+from mflux.post_processing.array_util import ArrayUtil
+from mflux.post_processing.generated_image import GeneratedImage
+from mflux.post_processing.image_util import ImageUtil
+
+
+class FluxKontext(nn.Module):
```
anthonywu: do we need another one of these top level classes? Wondering at what point we do some sub-classing either with is-a-Flux or has-a-Flux-Thing patterns.

Or going forward trust code assistants to update everywhere applicable because it's just busy work after all.

anthonywu (approved):

src/mflux/generate_kontext.py
```diff
@@ -0,0 +1,65 @@
...
+    parser.add_model_arguments(require_model_arg=False)
+    parser.add_lora_arguments()
+    parser.add_image_generator_arguments(supports_metadata_config=True)
+    parser.add_image_to_image_arguments(required=True)
+    parser.add_output_arguments()
+    args = parser.parse_args()
+
+    # 0. Set default guidance value if not provided by user
+    if args.guidance is None:
+        args.guidance = ui_defaults.GUIDANCE_SCALE
```
anthonywu: is there a rec'd guidance default for Kontext that's diff than plain Dev?

filipstrand: Good question, was wondering the same thing...

In the example here I saw they used a value of 2.5 (that is actually the only thing I have tried so far myself)
https://huggingface.co/black-forest-labs/FLUX.1-Kontext-dev

src/mflux/flux_tools/kontext/flux_kontext.py
filipstrand: Indeed, it start to feel like around now would be a good time to refactor/abstract a bit. I've tried to avoiding doing that too early to see what common patterns emerge from all features, but now it feels like we know roughly the commonalities/differences between the features and their implementations. 

Maybe less important, but same thing applies to the test helpers - a lot of them share almost the same structure so a lot can be done there too.

But as you say, it also suddenly feels a lot easier to deal with duplication when the agents are good at spotting this kind of similarities (but in my testing they sometimes fail with this when there is too much to change at once, e.g covering 5 out of 7 cases etc.)  

Might be good to invest some time into this after the next release. I've also been wanting to rearrange some things in the package structure. Part of the weight-handling could also be improved and simplified a bit.

filipstrand:
> what's your performance? On my M4 Max 128GB it's looking pretty meh
> 
> ```
> 3/14 [02:40<09:40, 52.78s/it]
> ```
> 
> this is the best MacBook Pro chip available today and it's ~1 minute per iteration. This smells wrong but I can also accept that maybe this is a taxing model.

Yeah mine is also quite slow, for a  640 x 384 image, 20 steps 8bit qant takes me more than 5 minutes on my 2021 M1 Pro :( , but my guess is that this is mostly due to having to work with latents that are twice the input resolution, (so similar to the other in-context approaches).

filipstrand:
@akx I typically quantise the model "on-the-fly" myself whenever starting a generation, it is very fast and convenient even on slower machines (and I don't mind storing the full precision weights locally). But I will upload the quantized mflux weights to HF (https://github.com/filipstrand/mflux/issues/227), unless someone beats me to it, just have not gotten around to that yet :) 

In the meantime, If you are willing to download the full weights temporarily, you can then use those to save a qauntaized copy locally, and after that only use that for subsequent generations https://github.com/filipstrand/mflux?tab=readme-ov-file#-saving-a-quantized-version-to-disk

Issue #220, opened 2025-06-27
Planning: Type cleanup


anthonywu:
I want to help do a cleanup of the type hinting debt.

Using the classic `mypy` and the new/upcoming Astral [ty](https://github.com/astral-sh/ty) here's what the two tools flag.

[mypy_snapshot.txt](https://github.com/user-attachments/files/20955825/mypy_snapshot.txt)
[ty_snapshot.txt](https://github.com/user-attachments/files/20955826/ty_snapshot.txt)

Because `ty` is in beta for now, using it as the source of correctness is probably not a good idea, so my proposal is to get to a checkpoint where `mypy` reports no warnings/errors.

Due to the expected bloat of the fixer changes, I think the cleanup should be done in multiple PRs, each focusing on a specific pattern or a specific collection of files, in order to make reviews easier.

Checking with @filipstrand to see if he agrees it's worthwhile to get `mypy` (and later `ty check` errors down to 0, then we can introduce `pre-commit` hooks to enforce.

PR #221, opened 2025-06-27
🧹 Cleanup/bump pre-commits; intro mypy config, quick type ignores/fixes

anthonywu:
First of N PRs for #220 - the fixes are what's flagged in #220's attachments.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,16 +1,21 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.11.13
+    rev: v0.12.1
     hooks:
       # Run the linter.
-      - id: ruff
+      - id: ruff-check
         types_or: [python, pyi]
         args: [--fix]
       # Run the formatter.
       - id: ruff-format
         types_or: [python, pyi]
   - repo: https://github.com/crate-ci/typos
-    rev: v1.31.0
+    rev: v1.33.1
     hooks:
       - id: typos
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.16.1
+    hooks:
+      - id: mypy
+        files: ^src/
--- a/src/mflux/callbacks/instances/battery_saver.py
+++ b/src/mflux/callbacks/instances/battery_saver.py
@@ -56,7 +56,7 @@ class BatterySaver(BeforeLoopCallback):
     def __init__(self, battery_percentage_stop_limit=10):
         self.limit = battery_percentage_stop_limit
 
-    def call_before_loop(self, **kwargs) -> None:
+    def call_before_loop(self, **kwargs) -> None:  # type: ignore
         current_pct: int | None = get_battery_percentage()
         if current_pct is not None and current_pct <= self.limit:
             raise StopImageGenerationException(f"Battery below {self.limit}% threshold: {current_pct}%")
--- a/src/mflux/config/runtime_config.py
+++ b/src/mflux/config/runtime_config.py
@@ -33,7 +33,7 @@ class RuntimeConfig:
         self.config.width = value
 
     @property
-    def guidance(self) -> float | None:
+    def guidance(self) -> float:
         return self.config.guidance
 
     @property
@@ -82,10 +82,10 @@ class RuntimeConfig:
 
         if is_img2img:
             # 1. Clamp strength to [0, 1]
-            strength = max(0.0, min(1.0, self.config.image_strength))
+            strength = max(0.0, min(1.0, self.config.image_strength))  # type: ignore
 
             # 2. Return start time in [1, floor(num_steps * strength)]
-            return max(1, int(self.num_inference_steps * strength))
+            return max(1, int(self.num_inference_steps * strength))  # type: ignore
         else:
             return 0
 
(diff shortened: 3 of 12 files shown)


.pre-commit-config.yaml
```diff
@@ -1,16 +1,21 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.11.13
+    rev: v0.12.1
     hooks:
       # Run the linter.
-      - id: ruff
+      - id: ruff-check
```
anthonywu, line 7: `ruff` id was deprecated for newer `ruff-check`, no-op change

src/mflux/callbacks/instances/battery_saver.py
```diff
@@ -56,7 +56,7 @@ class BatterySaver(BeforeLoopCallback):
     def __init__(self, battery_percentage_stop_limit=10):
         self.limit = battery_percentage_stop_limit
 
-    def call_before_loop(self, **kwargs) -> None:
+    def call_before_loop(self, **kwargs) -> None:  # type: ignore
```
anthonywu, line 59: mypy didn't like that I accepted `**kwargs` because I don't want to list out the before-loop kwargs I definitely don't need to use

src/mflux/config/runtime_config.py
```diff
@@ -33,7 +33,7 @@ def width(self, value):
         self.config.width = value
 
     @property
-    def guidance(self) -> float | None:
+    def guidance(self) -> float:
```
anthonywu, line 36: detected that if it can be `None`, then it should not be used in calculations. I think our `ui.defaults` handles this so should not be None by the time we reach here

src/mflux/config/runtime_config.py
```diff
@@ -82,10 +82,10 @@ def init_time_step(self) -> int:
 
         if is_img2img:
             # 1. Clamp strength to [0, 1]
-            strength = max(0.0, min(1.0, self.config.image_strength))
+            strength = max(0.0, min(1.0, self.config.image_strength))  # type: ignore
 
             # 2. Return start time in [1, floor(num_steps * strength)]
-            return max(1, int(self.num_inference_steps * strength))
+            return max(1, int(self.num_inference_steps * strength))  # type: ignore
```
anthonywu, line 88: complains that the config attributes can be None, which we know, but there's plenty of gating before this point

src/mflux/dreambooth/dataset/dataset.py
```diff
@@ -27,10 +27,9 @@ def prepare_dataset(
         examples = Dataset._create_examples(flux, raw_data, width=width, height=height)
 
         # Expend the original dataset to get more training data with variations
-        augmented_examples = []
-        for example in examples:
-            [augmented_examples.append(variation) for variation in DreamBoothPreProcessing.augment(example)]
```
anthonywu, line 32: mypy complained this just built an array of `None` outputs because `.append(...)` returns None

src/mflux/dreambooth/dataset/dataset.py
```diff
@@ -27,10 +27,9 @@ def prepare_dataset(
...
-
+        augmented_examples = [
+            variation for example in examples for variation in DreamBoothPreProcessing.augment(example)
+        ]
```
anthonywu, line 32:
```python
# List comprehension with double loop
result = [x * y for x in [1, 2, 3] for y in [10, 20]]
print(result)
# Output: [10, 20, 30, 40, 60, 120]

# same as
result = []
for x in [1, 2, 3]:      # Outer loop (first in comprehension)
    for y in [10, 20]:   # Inner loop (second in comprehension)
        result.append(x * y)
```

src/mflux/flux_tools/redux/flux_redux.py
```diff
@@ -55,17 +55,17 @@ def generate_image(
         self,
         seed: int,
         prompt: str,
-        config: Config,
+        runtime_config: Config,
     ) -> GeneratedImage:
         # 0. Create a new runtime config based on the model type and input parameters
-        config = RuntimeConfig(config, self.model_config)
```
anthonywu, line 61: mypy didn't like `config` name was re-used between the arg list and the local scope var

src/mflux/models/vae/common/resnet_block_2d.py
```diff
@@ -66,7 +66,7 @@ def __call__(self, input_array: mx.array) -> mx.array:
         hidden_states = self.norm2(hidden_states.astype(mx.float32)).astype(Config.precision)
         hidden_states = nn.silu(hidden_states)
         hidden_states = self.conv2(hidden_states)
-        if self.is_conv_shortcut:
+        if self.conv_shortcut is not None:
             input_array = self.conv_shortcut(input_array)
```
anthonywu, line 70: complained that `self.conv_shortcut` could be `None` - so just check on that which is the same logic

src/mflux/callbacks/instances/battery_saver.py
filipstrand: Yeah, this is also a little bit of weird design from me (again bringing in some "java-like" like pattern of having the interface/protocol state every dependency). This worked well in the beginning, but as we got more uses of the callbacks, it now starts to look a bit weird... I would also be happy to change the signature to `def call_before_loop(self, **kwargs)`, or something similar, elsewhere too to keep things similar across all implementations.

src/mflux/config/runtime_config.py
filipstrand: makes sense

src/mflux/flux_tools/redux/flux_redux.py
filipstrand: I can see that it might bring confusion, in this case I think we should rename `config -> runtime_config` across all Flux implementations (can be done in a separate PR)

Issue #216, opened 2025-06-24
HuggingFace downtime blocks local generation due to assumed HF uptime


anthonywu:
As I write this issue, HF is down for the first time I can recall, which is blocking local generation.

I think the todos are:

1. don't check for tokenizers if the model is already cached
2. `_download_or_get_cached_tokenizers` should have reasonable timeout handling instead of blocking

```
$ date && mflux-generate -m dev --prompt "beautiful landscape"
Mon Jun 23 22:22:07 PDT 2025

Traceback (most recent call last):
    File "/workspace/mflux/.venv/bin/mflux-generate", line 10, in <module>
    sys.exit(main())
             ~~~~^^
    File "/workspace/mflux/src/mflux/generate.py", line 25, in main
    flux = Flux1(
        model_config=ModelConfig.from_name(model_name=args.model, base_model=args.base_model),
    ...<3 lines>...
        lora_scales=args.lora_scales,
    )
  File "/workspace/mflux/src/mflux/flux/flux.py", line 38, in __init__
    FluxInitializer.init(
    ~~~~~~~~~~~~~~~~~~~~^
        flux_model=self,
        ^^^^^^^^^^^^^^^^
    ...<4 lines>...
        lora_scales=lora_scales,
        ^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/workspace/mflux/src/mflux/flux/flux_initializer.py", line 40, in init
    tokenizers = TokenizerHandler(
        repo_id=model_config.model_name,
        max_t5_length=model_config.max_sequence_length,
        local_path=local_path,
    )
  File "/workspace/mflux/src/mflux/tokenizer/tokenizer_handler.py", line 16, in __init__
    root_path = Path(local_path) if local_path else TokenizerHandler._download_or_get_cached_tokenizers(repo_id)
                                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/workspace/mflux/src/mflux/tokenizer/tokenizer_handler.py", line 32, in _download_or_get_cached_tokenizers
    snapshot_download(
    ~~~~~~~~~~~~~~~~~^
        repo_id=repo_id,
        ^^^^^^^^^^^^^^^^
        allow_patterns=["tokenizer/**", "tokenizer_2/**"],
[...]

PR #226, opened 2025-07-02
🗃️ Implement Local-first behavior to be resilient when HF/network is down

anthonywu:
This is a fix for #216 

# 🗃️ Implement Cache-First Model Loading to Handle Hugging Face Downtime

## Summary

This PR implements a cache-first approach for model downloads to ensure mflux can continue working even when Hugging Face is experiencing downtime. Previously, when HF was down, model loading would fail even if the models were already cached locally. Not exactly great for a product that is supposed to be local and network/cloud-optional!

## Problem

`snapshot_download` default behavior checks the HF site anyways, there are relevant flags we did not yet.

## Solution

Created a new `mflux.weights.download` module with a wrapper around `huggingface_hub.snapshot_download` that:

  1. First attempts to load from local cache using `local_files_only=True`
  2. Only attempts to download from Hugging Face if the cache doesn't exist
  3. Preserves the original behavior when `force_download` or `local_files_only` are explicitly set
  4. Isolates the `try/except` wrapper to just one place in the project, so all existing usages of `force_download` is just a simple modification.

##  Changes

  - New module: `src/mflux/weights/download.py`
  - Updated imports: All modules now use mflux.weights.download.snapshot_download instead of the direct HF import
    - `src/mflux/weights/weight_handler.py`
    - `src/mflux/weights/weight_handler_lora_huggingface.py`
    - `src/mflux/controlnet/weight_handler_controlnet.py`
    - `src/mflux/flux_tools/redux/weight_handler_redux.py`

## Small behavior change

If you didn't previously suppress HF progress bars, you would see them every generation. Now you should not see the progress bar after the model has been downloaded the first time (because no attempt to reach out to HF site!)

## Testing

Decided not to add test maintenance burden because this is a binary work/break chokepoint that the test generation tests will exercise.

### DIFF
--- a/src/mflux/tokenizer/tokenizer_handler.py
+++ b/src/mflux/tokenizer/tokenizer_handler.py
@@ -3,6 +3,7 @@ from pathlib import Path
 import transformers
 
 from mflux.tokenizer.clip_tokenizer import TokenizerCLIP
+from mflux.weights.download import snapshot_download
 
 
 class TokenizerHandler:


src/mflux/weights/download.py
```diff
@@ -0,0 +1,41 @@
...
+    Otherwise, it will try local cache first, then download if cache is not found.
+
+    Args:
+        repo_id: A user or an organization name and a repo name separated by a `/`.
+        **kwargs: Additional arguments passed to huggingface_hub.snapshot_download
+
+    Returns:
+        str: Path to the downloaded/cached repository
+
+    Docs: https://huggingface.co/docs/huggingface_hub/v0.33.1/en/package_reference/file_download#huggingface_hub.snapshot_download
```
anthonywu, line 21: I confirmed the flags in this upstream docs. Nothing  hallucinated.

Vibe coding report: cost $10, probably better off having done it manually. Claude Code started out with duplicating the try/except in every file, causing a big/dangerous diff. Then I guided it to create the proxy helper and the diff scope is now tolerable.

filipstrand: Nice. I like how this is just an import change in all other files!

filipstrand (approved):
Great addition! It should clearly work like how you describe it here. I didn't get the chance to see the failure myself during the HF downtime, but I have been able to run image generation completely offline before, it feels like it would be similar to not being able to reach HF?

Issue #206, opened 2025-06-10
Expected folder paths


exdysa:
Hi! I was able to generate from Shuttle in less than 14 seconds! Needless to say I am a fan of this project.

I encountered an issue where the system fails when quantization request is received.

<details><summary>That looks like this</summary>

```
The tokenizer class you load from this checkpoint is not the same type as the class this function is called from. It may result in unexpected tokenization. 
The tokenizer class you load from this checkpoint is 'T5Tokenizer'. 
The class this function is called from is 'CLIPTokenizer'.
Traceback (most recent call last):
  File "/Users/{user__venv_folderfolder_etc}/.venv/bin/mflux-save", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/{user__venv_folderfolder_etc}/.venv/lib/python3.13/site-packages/mflux/save.py", line 13, in main
    flux = Flux1(
        model_config=ModelConfig.from_name(args.model, base_model=args.base_model),
    ...<2 lines>...
        lora_scales=args.lora_scales,
    )
  File "/Users/{user__venv_folderfolder_etc}/.venv/lib/python3.13/site-packages/mflux/flux/flux.py", line 38, in __init__
    FluxInitializer.init(
    ~~~~~~~~~~~~~~~~~~~~^
        flux_model=self,
        ^^^^^^^^^^^^^^^^
    ...<4 lines>...
        lora_scales=lora_scales,
        ^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/{user__venv_folderfolder_etc}/.venv/lib/python3.13/site-packages/mflux/flux/flux_initializer.py", line 39, in init
    tokenizers = TokenizerHandler(
        repo_id=model_config.model_name,
        max_t5_length=model_config.max_sequence_length,
        local_path=local_path,
    )
  File "/Users/{user__venv_folderfolder_etc}/.venv/lib/python3.13/site-packages/mflux/tokenizer/tokenizer_handler.py", line 18, in __init__
    self.clip = transformers.CLIPTokenizer.from_pretrained(
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        pretrained_model_name_or_path=root_path / "tokenizer",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        local_files_only=True,
[...]

PR #228, opened 2025-07-03
Proposal: if third party model does not provide resources like tokeni…

anthonywu:
This is a discussion/proposal for how to gracefully support the ask in #206 by @exdysa

- [ ] WIP: PR diffs against in-review `handle-hf-downtime` branch in #226. Reset to `main` and rebase

# Problem

1. There exists some third party model e.g. `XLabs-AI/flux-dev-fp8`
2. The provider does not provide the expected folder paths e.g. `tokenizers`, etc [model files](https://huggingface.co/XLabs-AI/flux-dev-fp8/tree/main)
3. Can we assume that these models just want us to re-use the upstream dev/schnell models
4. Use the `snapshot_download` wrapper/proxy to gracefully fallback to default model resources

I think we always suspected this was a problem, just didn't prioritize the workaround.

### DIFF
--- a/src/mflux/weights/download.py
+++ b/src/mflux/weights/download.py
@@ -4,7 +4,7 @@ from huggingface_hub import snapshot_download as hf_snapshot_download
 from huggingface_hub.utils import LocalEntryNotFoundError
 
 
-def snapshot_download(repo_id: str, **kwargs) -> str:
+def snapshot_download(repo_id: str, fallback_repo_id="black-forest-labs/FLUX.1-dev", **kwargs) -> str:
     """Download repo files with cache-first behavior.
 
     This wrapper function attempts to use cached files first before downloading.
@@ -33,9 +33,21 @@ def snapshot_download(repo_id: str, **kwargs) -> str:
         # Override local_files_only to True for cache check
         cache_kwargs = kwargs.copy()
         cache_kwargs["local_files_only"] = True
-        return hf_snapshot_download(repo_id=repo_id, **cache_kwargs)
+        cache_path = hf_snapshot_download(repo_id=repo_id, **cache_kwargs)
+        return cache_path
     except LocalEntryNotFoundError:
         # Cache doesn't exist, download from Hugging Face
+        # optimistically assuming provider has the target files
         download_kwargs = kwargs.copy()
         download_kwargs["local_files_only"] = False
-        return hf_snapshot_download(repo_id=repo_id, **download_kwargs)
+        try:
+            cache_path = hf_snapshot_download(repo_id=repo_id, **cache_kwargs)
+            print(f"{repo_id} cache found for {kwargs}")
+            return cache_path
+        except LocalEntryNotFoundError:
+            # most likely: third party derivative model re-uses official resources
+            # TODO: does dev/schnell official resources differ? if so need a repo_id -> fallback map
+            #       similar to model/base_model setup
+            cache_path = snapshot_download(fallback_repo_id, **kwargs)
+            print(f"{repo_id} serving as fallback repo_id for resources {kwargs.get('allow_patterns', '')}")
+            return cache_path


src/mflux/weights/download.py
```diff
@@ -33,9 +33,21 @@ def snapshot_download(repo_id: str, **kwargs) -> str:
...
+        try:
+            cache_path = hf_snapshot_download(repo_id=repo_id, **cache_kwargs)
+            print(f"{repo_id} cache found for {kwargs}")
+            return cache_path
+        except LocalEntryNotFoundError:
+            # most likely: third party derivative model re-uses official resources
+            # TODO: does dev/schnell official resources differ? if so need a repo_id -> fallback map
+            #       similar to model/base_model setup
+            cache_path = snapshot_download(fallback_repo_id, **kwargs)
+            print(f"{repo_id} serving as fallback repo_id for resources {kwargs.get('allow_patterns', '')}")
```
anthonywu:
```
mflux-generate -m XLabs-AI/flux-dev-fp8 --base-model dev --steps 4 --prompt "a shiny diamond, a banner that reads xlabs"
XLabs-AI/flux-dev-fp8 serving as fallback repo_id for resources ['tokenizer/**', 'tokenizer_2/**']
XLabs-AI/flux-dev-fp8 serving as fallback repo_id for resources ['text_encoder/*.safetensors', 'text_encoder_2/*.safetensors', 'transformer/*.safetensors', 'vae/*.safetensors']
  0%|                                                                                                                                                  | 0/4 [00:00<?, ?it/s]
```

will remove `print` during PR cleanup

src/mflux/weights/download.py
```diff
@@ -33,9 +33,21 @@ def snapshot_download(repo_id: str, **kwargs) -> str:
...
+        try:
+            cache_path = hf_snapshot_download(repo_id=repo_id, **cache_kwargs)
+            print(f"{repo_id} cache found for {kwargs}")
+            return cache_path
+        except LocalEntryNotFoundError:
+            # most likely: third party derivative model re-uses official resources
+            # TODO: does dev/schnell official resources differ? if so need a repo_id -> fallback map
+            #       similar to model/base_model setup
+            cache_path = snapshot_download(fallback_repo_id, **kwargs)
+            print(f"{fallback_repo_id} serving as fallback repo_id for resources {kwargs.get('allow_patterns', '')}")
```
anthonywu:
```
$ mflux-generate -m XLabs-AI/flux-dev-fp8 --base-model dev --steps 4 --prompt "a shiny diamond"
black-forest-labs/FLUX.1-dev serving as fallback repo_id for resources ['tokenizer/**', 'tokenizer_2/**']
black-forest-labs/FLUX.1-dev serving as fallback repo_id for resources ['text_encoder/*.safetensors', 'text_encoder_2/*.safetensors', 'transformer/*.safetensors', 'vae/*.safetensors']
100%|█████████████████████████████████████████████████████████████████████████████████████| 4/4 [01:31<00:00, 22.93s/it]
```

src/mflux/flux/flux_initializer.py
```diff
@@ -37,6 +37,7 @@ def init(
         flux_model.model_config = model_config
 
         # 1. Initialize tokenizers
+        # TODO: tokenizers needs to also refer to base model tokenizer assets
```
anthonywu, line 40: the earlier 3rd party model run "success" was a false win, all assets led to fallback official repo so it was not really using the third party model.

The latest commit makes further progress, but I will have to leave it WIP here, and maybe @filipstrand will find a quick suggestion/win based on the update so far.

exdysa:
Apologies, neglected to link the initial attempt that got through a few checks https://github.com/exdysa/filipstrand-mflux/tree/x/conditional_tokenizers

filipstrand:
I'll beginning by saying that I have not looked into other models much and personally use the standard ones myself, but it is interesting to hear findings things like `there has been growing evidence that CLIP tends to hold diffusion models back, rather than improve them`. 

Just to get up to speed a bit, are we essentially discussing these two approaches? :

a) @anthonywu 's approach of simply reusing the standard tokenizers & text_encoders if they are not present for the 3rd party model (e.g by using the ones from `black-forest-labs/FLUX.1-schnell` or similar).

b) Somehow make one (or the other?) tokeniser/text_encoder optional (sounds like you are referring to making the CLIP part optional @exdysa?)

I feel approach in a) feels quite natural, but it sounds like you are saying @exdysa that this assumption that Anthony mentions is not necessarily valid? 

```
Can we assume that these models just want us to re-use the upstream dev/schnell models
```

On the other hand, I guess approach b) would require a more structural change to the network architecture, since we use both outputs from the text_encoders `prompt_embeds` (from T5) and `pooled_prompt_embeds` (from CLIP) thought the network

```python
prompt_embeds, pooled_prompt_embeds = PromptEncoder.encode_prompt(
            prompt=prompt,
            prompt_cache=self.prompt_cache,
            t5_tokenizer=self.t5_tokenizer,
            clip_tokenizer=self.clip_tokenizer,
            t5_text_encoder=self.t5_text_encoder,
            clip_text_encoder=self.clip_text_encoder,
        )
```

(e.g in the Transformer class these are being processed further as `encoder_hidden_states` and `text_embeddings` etc  and it is not obvious to me right now what it would mean to leave one out). I guess it would be possible to do this, but it feels like a bigger task of generalising the architecture layout to support Flux alternatives where not only the weights are updated, but also the "wiring" of the network itself.

filipstrand:
> Eh. We're experimenting. Try this way first, maybe it will turn out alright?

Yes I think we can start with this and see how it works...

As you mentioned SD3, they are very similar architecture-wise indeed. Actually, I started this project working on SD3 and was maybe 70% finished when FLUX was released (I should have the implementation somewhere, but have not looked at it since I published this project). After I read the initial feedback on Flux vs SD3 I decided to pivot completely to Flux. @exdysa  How is SD3/3.5 doing nowadays? Does it have any pros compared to Flux?

filipstrand:
@anthonywu I've been away for a few days and forgot about this, but now that we reverted `snapshot_download` in main, I guess things have to be updated for this branch too? 

I like the idea of this feature and think we can include it in next release, but I didn't feel confident doing the rebase on main myself

filipstrand:
not urgent at all, just let me know once it starts to feel ready again :)

PR #232, opened 2025-07-05
🏗️ uv build backend update, trim 33MB from user pip install load

anthonywu:
- converge to the excellent universe of `uv`, `ruff`, `ty` by Astral
- opportunity: trim the examples/docs images from `pip install` distributable, saving users network load, disk space and wall time during `pip install`s
- optimistically update the GitHub Actions build step from `python -m build` to `uv build` (while keeping `twine`)

@filipstrand - if the GitHub Actions update has a mistake, you may have to bump a few nit changes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,33 +14,32 @@ jobs:
     permissions:
       contents: write
       pull-requests: read
-    
+
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-        
+
       - name: Validate confirmation
         run: |
           if [ "${{ github.event.inputs.confirm }}" != "publish" ]; then
             echo "❌ Must type 'publish' to confirm release"
             exit 1
           fi
-        
+
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
-          
+          python-version: "3.10"
+
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
-          python -m pip install requests build twine toml
-          
+          python -m pip install --upgrade uv
+
       - name: Configure Git
         run: |
           git config --global user.name "GitHub Actions"
           git config --global user.email "actions@github.com"
-          
+
       - name: Run release script
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,17 @@
 [build-system]
-requires = ["setuptools>=42", "wheel"]
-build-backend = "setuptools.build_meta"
+requires = ["uv_build>=0.7.19,<0.8.0"]
+build-backend = "uv_build"
+
+[tool.uv.build-backend]
+default-excludes = true  # __pycache__, *.pyc, and *.pyo
+source-exclude = [
+    # documentation assets were 27MB on 2025-07-05
+    "**/assets/**",
+    # # dreambooth examples ~=5MB on 2025-07-05
+    "**/dreambooth/_example/images/**",
+    # loss pdf/tex does not need to be distributed
+    "**/optimization/_loss_derivation/**",
+]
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -1,10 +1,21 @@
...
+
+[tool.uv.build-backend]
+default-excludes = true  # __pycache__, *.pyc, and *.pyo
+source-exclude = [
+    # documentation assets were 27MB on 2025-07-05
+    "**/assets/**",
+    # dreambooth examples ~=5MB on 2025-07-05
+    "**/dreambooth/_example/images/**",
+    # loss pdf/tex does not need to be distributed
+    "**/optimization/_loss_derivation/**",
```
anthonywu, line 13:
```
du -sh src/mflux/assets src/mflux/dreambooth/_example/
 26M	src/mflux/assets
5.3M	src/mflux/dreambooth/_example/
```

without these, the entire `dist` is back to < 1MB

scripts/release.py
```diff
@@ -190,7 +184,8 @@ def build_package(self):
             print("🧹 Cleaned dist/ directory")
 
         # Build the package
-        self.run_command([sys.executable, "-m", "build"], "Building package")
+        self.run_command([sys.executable, "-m", "uv", "--version"], "Verify 'uv build' version")
```
anthonywu, line 187:
```
python -m uv --version
uv 0.7.19 (38ee6ec80 2025-07-02)
```

scripts/release.py
```diff
@@ -190,7 +184,8 @@ def build_package(self):
...
+        self.run_command([sys.executable, "-m", "uv", "build"], "Building package with 'uv build'")
```
anthonywu, line 188:
```
python -m uv build
Building source distribution (uv build backend)...
Building wheel from source distribution (uv build backend)...
Successfully built dist/mflux-0.9.2.tar.gz
Successfully built dist/mflux-0.9.2-py3-none-any.whl
```

scripts/release.py
```diff
@@ -175,12 +175,6 @@ def run_command(cmd: list, description: str, check: bool = True, env: dict = Non
         return result
 
     def build_package(self):
-        # Install build dependencies
-        self.run_command(
-            [sys.executable, "-m", "pip", "install", "--upgrade", "pip", "build", "twine"],
-            "Installing build dependencies",
-        )
-
```
anthonywu, line 183: I suspect this was redundant because in the `release.yaml` we already had:

```
python -m pip install --upgrade uv
```

.github/workflows/release.yml
```diff
@@ -14,33 +14,34 @@ jobs:
...
       contents: write
       pull-requests: read
-    
+
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
-        
+        with:
+          fetch-depth: 0 # Need full history for git operations
```
anthonywu, line 22: Claude suggested this, seems harmless to enable for future actions

docs: https://github.com/actions/checkout

```
# Number of commits to fetch. 0 indicates all history for all branches and tags.
    # Default: 1
```

pyproject.toml
filipstrand: nice!

scripts/release.py
filipstrand: Ah yes. This whole script was actually vibe-coded quite quickly as a last thing before the 0.8 release, so to be honest I have not actually read the contents too closely 😅. Thanks for going through it (also feel free to completely restructure it/rewrite it if you feel the need to.) 

It might not be the prettiest, but it seems to work with the main 2 goals of: 
1) generating and publishing the release to PyPi
2) generating the release notes from the changelog

I also saw there is [some error](https://github.com/filipstrand/mflux/actions/runs/16077977422/job/45377359322) that I have not yet had the time to investigate yet. I will try to do so tomorrow and do a 0.9.2 release including this PR

PR #241, opened 2025-08-01
Add support for FLUX.1-Krea-dev model

filipstrand:
Downloading the weights now... no architecture change so this is mostly a config update. 

**Update**
Will also try to make this compatible with mlx>=0.27


anthonywu (approved):
I've failed the download the weights in the last 24 hours (many failures and timeouts), but pre-approving as this is expected to work.

filipstrand:
Same for me, tried a bunch of times and finally got it working. Seems to be working well with no additional tweaks! Simply swap `dev` for `dev-krea` 

I'm also lagging a bit behind on the test cases when upgrading to newer mlx version (slight image diffs as usual), so I'll try to address that today and bump to mlx==0.27.1 . Would be cool to see if mflux would work out of the box on a CUDA machine :)

filipstrand:
@claude Alongside adding support for the new image model, we also update the release to 0.10.0 - this includes adding new changeling entries, bumping versions etc etc, have we forgotten anything based on what has been done previously when doing a new release? (When reviewing the code, don't add any new commits, but instead describe things in plain text only)

filipstrand:
I guess some things should work out-of-the-box with this model, such as LoRA training. 

Upscaling is essentially dev + special controlnet, but right now it is hardcoded to work with the dev model:

```
"dev-controlnet-upscaler": ModelConfig(
        alias="dev-controlnet-upscaler",
        model_name="black-forest-labs/FLUX.1-dev",
        base_model=None,
        controlnet_model="jasperai/Flux.1-dev-Controlnet-Upscaler",
        custom_transformer_model=None,
        num_train_steps=1000,
        max_sequence_length=512,
        supports_guidance=False,
        requires_sigma_shift=False,
        priority=8,
    ),
```

If this (and other controlnet features) also works well with `dev-krea` then we should support that too, this could be added in a 0.10.x release

anthonywu:
CUDA testing can be done on Modal, a very python friendly gpu cloud host

anthonywu:
I would alias krea-dev and dev-krea to the same model config. People will type it wrong and it's better to help them out.

filipstrand:
> I would alias krea-dev and dev-krea to the same model config. People will type it wrong and it's better to help them out.

Good idea! I just released I had it backwards and wanted to stick to something close to the official naming. Will update it to support both

filipstrand:
> CUDA testing can be done on Modal, a very python friendly gpu cloud host

Heard good things about Modal (started by a fellow swede :) but have not tried it out yet, this would be a great opportunity! Have you tried it @anthonywu? Wonder if they might have images pre-loaded with FLUX huggingface weights already?

anthonywu:
> Have you tried it @anthonywu? Wonder if they might have images pre-loaded with FLUX huggingface weights already?

I have used Modal. The last time I used it, I had to preload the model cache on a volume, then mount the volume with every job to access the model files, as well as tweaking app and dependencies to expect the models at a different path.

filipstrand:
Okay. It will be interesting to see how seamless it would be to run this on a linux machine. I'm reminded of the platformdirs thing you did a while back, which makes me wonder what other Mac-specific things we currently might have in the project...

filipstrand:
Running the full test suite once again before going to bed, I'll aim for a release tomorrow. Might also include a fix for https://github.com/filipstrand/mflux/issues/224 if it is quick and easy

PR #246, opened 2025-08-03
Turbo mode

filipstrand:
Last feature to add before 0.10 release: 

This is essentially a just a LoRA https://huggingface.co/alimama-creative/FLUX.1-Turbo-Alpha which I have found quite good in my personal use (though have not had the time to try it yet with Krea). Allows dev like models to be run with fewer steps while maintaining reasonable quality. 

Note: Completely vibe coded so far (will do some cleanup later today) 

Was debating whether or not to have this "baked in" as a special feature in the respective Flux classes but opted for a simple command line addition that simply append this LoRA. So this is only a convenient shortcut


src/mflux/ui/turbo_util.py
```diff
@@ -0,0 +1,42 @@
+class TurboUtil:
+    # Turbo mode configuration
+    TURBO_LORA_REPO_ID = "alimama-creative/FLUX.1-Turbo-Alpha"
+    TURBO_LORA_FILENAME = "diffusion_pytorch_model.safetensors"
+    TURBO_GUIDANCE = 3.5
```
anthonywu, line 5: I'd name this `DEFAULT_TURBO_GUIDANCE` because it's overrideable by user

src/mflux/ui/turbo_util.py
```diff
@@ -0,0 +1,42 @@
...
+        if not args.turbo:
+            return
+
+        # Add turbo LoRA parameters
+        # Safely initialize lora_names, preserving any existing values
+        if not hasattr(args, "lora_names") or args.lora_names is None:
+            args.lora_names = []
+
+        # Convert CLI's singular lora_name to plural lora_names list if provided
+        if hasattr(args, "lora_name") and args.lora_name:
```
anthonywu, line 18: this smells like tech debt, we should just converge towards the plural CLI param name, but would trade-off some breakage

src/mflux/ui/turbo_util.py
```diff
@@ -0,0 +1,42 @@
...
+                args.lora_names = []
+            if args.lora_name not in args.lora_names:
+                args.lora_names.append(args.lora_name)
+
+        # Handle LoRA scales when turbo adds an additional LoRA
+        # If user provided lora_paths but no lora_scales, we need to add default scales
+        if hasattr(args, "lora_paths") and args.lora_paths and isinstance(args.lora_paths, list):
+            if not hasattr(args, "lora_scales") or args.lora_scales is None:
+                # Default to 1.0 for all existing user LoRAs
+                args.lora_scales = [1.0] * len(args.lora_paths)
```
anthonywu, line 30: this processing of arg values is in line with the logic in `parsers.py`, and I think would be more consistent if we continued to utilize `parsers.py` to pre-process arg values.

anthonywu (changes requested):
1. considering "what if another vendor comes out with a turbo-labeled" model, which one wins, feels like we're picking a winner here. I think users can save a shortcut of this problem by using the metadata json files with the turbo lora pre-populated?
1. alternatively, what if we offered pre-loaded 'recipes' where the alimama turbo workflow is something you can enable like `--recipe alimama-turbo` (open to flag name changes)
1. you may be able to get rid of `turbo_util.py` if you follow the pattern in `parsers.py` to pre-process flags as they come in before they touch the `generate*` controllers. I get the feeling the `parsers.py` may be mission creeping, but the `lora_name[s]` parsing definitely belongs in there.

filipstrand:
Very valid concerns.

> considering "what if another vendor comes out with a turbo-labeled" model, which one wins, feels like we're picking a winner here.

True. I like the `--turbo` keyword in general I agree that we could be claiming it a bit too early with this specific implementation. I can see other optimisation that would go under such label (another LoRA, as you say, or another method completely, like https://github.com/ali-vilab/TeaCache, etc. ).   

>  I think users can save a shortcut of this problem by using the metadata json files with the turbo lora pre-populated?

Did not think of this and honestly forgot about this feature. Given that we have support for this it might be a lot more general to use this feature instead for this purpose... Is it correct that it currently does not support `lora_names` and `lora_repo_id` (which would enable automatic download of the specified LoRAs)?  I'm guessing that for us to support this we would also need to add the `lora_names`, `lora_repo_id` inputs in `generate.py`, `generate_controlnet.py` etc, as was done in this PR. 

On a slightly related note here, I feel like there could be simplifications made to how we specify loras overall through the system: Now we have `lora_repo_id`, `lora_names`, `lora_paths`, `lora_scales` and it is not obvious how these work by just glancing at them, offline/online assumptions etc... I'm guessing we could have a simpler interface backed by a smarter lookup procedure that would look things offline first, then try to download them if not founds etc etc. But this is another topic. Curious if you've had similar thoughts on this?

> alternatively, what if we offered pre-loaded 'recipes' where the alimama turbo workflow is something you can enable like --recipe alimama-turbo (open to flag name changes)

Yes I like this idea. The good thing about this kind of solution would be that we could provide a few "curated" recipes that we've found works well in practice. It feels like there is a bit of a jungle out there with different LoRAs claiming to do certain things etc, and that the once we find very useful and general could be . Automatic download of the adapters would have to be supported. 

In conclusion, I guess I kinda like the idea of a general "turbo" mode (simple and without any additional setup), but agree that it probably deserves more thought into how that should work. Now that I've considered your points here, I feel this PR was a bit too rushed from my end and now I think it is better to leave this out for now, and take a more proper stab at this in a later release.

Issue #248, opened 2025-08-05
open discussion: mflux-extras, affiliated libraries


anthonywu:
One thing I like about libraries is the good ones are well scoped, the authors decide what's in and what's out of scope as a guideline to continual development.

One way `mflux` embodied this spirit from the beginning was to not rely on `torch` or `diffusers`, and lean into relying on MLX. Another way we've held the line is keeping some scripts in the tools dir but not officially part of the mflux package. This ensures the library does not mission creep.

However, this conflicts with the maintainers/users desire of ever expanding feature sets such as:

- Gradio GUI
- RESTful API
- DALLE-compatible endpoints
- ComfyUI integration points
- Upscale using a specific controlnet model (implemented)
- Turbo using a specific third party model (postponed)

My proposal is we could:

1. shove all of these into an additional package called `mflux-extras` and import as either `mflux_extras` or with some [namespaced packages setup](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) we can make `mflux.extras` but publish separately. (or multiple packages, `mflux-workflows`, `mflux-servers`, and `mflux-mcp` which I'm working on)
2. `mflux` will have dependency groups including a `extras` for `pip install mflux[extras]` for when users want all the bells and whistles (but acknowlege venv bloat and potential install conflicts)
3. all affiliated packages will publish CLI tools under the `mflux` top level namespace, and `mflux-` prefix as CLIs

I can volunteer to bootstrap all the new github repos and pypi artifacts, it's actually not that big of an effort.

azrahello:
t’s not a bad thing at all, especially if it could help make a porting to ComfyUI easier. Unfortunately, for ComfyUI, Rayer’s project is currently stalled — and in the end, it was more of a wrapper, but too closed off for proper ComfyUI integration. Development has been inactive for quite a while.

Back then, LoRAs, models, and images were loaded directly via the wrapper, but there was no way to connect the nodes with standard ComfyUI nodes — so practically, it was easier to just use the command line directly.

The only project that was showing some promising direction was [ComfyUI-MLX by thoddnn](https://github.com/thoddnn/ComfyUI-MLX), which was based on DiffusionKit. But even that is now stalled and lacks LoRA support.

[...]

PR #249, opened 2025-08-05
make mflux a namespace package (prepare for mflux.mcp)

anthonywu:
related to #248 

In this change, we (hope) harmlessly convert `mflux` into a namespaced project so that other repos can create `mflux.fooanything` subpackages.

The strategy is documented:

- https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages
- https://docs.astral.sh/uv/concepts/build-backend/#namespace-packages

^ basically we have to delete `mflux/__init__.py` which fortunately does nothing so I think we lose nothing here

And the integration checkpoint is here:

- https://github.com/anthonywu/mflux-mcp/commit/603ce45080963b5847e60e4d5de4097acfba9667#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R84

The series of work I am planning goes like this:

  - mflux repo prepares for namespaced sub-packages
  - mflux-mcp repo provides mflux.mcp subpackage and mflux-mcp-server executable
  - implement mflux-mcp-server with ONLY the standard image generation (control code review scope)
  - validate that mflux-mcp-server works with Claude Desktop and other must-test clients
  - prototype/build a SwiftUI frontend that uses MCP to speak to mflux-mcp-server
  - once this circle of success is locked in, I'll let loose the code assistant help to fill out the other generation modes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,9 @@ requires = ["uv_build>=0.7.19,<0.8.0"]
 build-backend = "uv_build"
 
 [tool.uv.build-backend]
-default-excludes = true  # __pycache__, *.pyc, and *.pyo
+default-excludes = true # __pycache__, *.pyc, and *.pyo
+module-name = "mflux"
+namespace = true
 source-exclude = [
     # documentation assets were 27MB on 2025-07-05
     "**/assets/**",


anthonywu:
@filipstrand I think this PR is safe to merge assuming the tests pass all the same

PR #254, opened 2025-08-20
install pytorch only for special workflows

anthonywu:
As one of my side quests, I'm trying to get MLX to run on cloud services - two in particular are:

1. Modal Functions (blocker: https://github.com/google/gvisor/issues/12034)
2. Fal Serverless (blocker: various, will share soon)

So to manage scope for now, one of the easier modes to get working is "CPU only": https://ml-explore.github.io/mlx/build/html/install.html#cpu-only-linux

So off I go to install mflux, and I notice it's a lot slower and the venv is a lot bigger on Linux!

> pip show torch # after the fact

```
Name: torch
Version: 2.8.0
Summary: Tensors and Dynamic neural networks in Python with strong GPU acceleration
Home-page: https://pytorch.org/
Author: PyTorch Team
Author-email: packages@pytorch.org
License: BSD-3-Clause
Location: /app/env/lib/python3.11/site-packages
Requires: filelock, fsspec, jinja2, networkx, nvidia-cublas-cu12, nvidia-cuda-cupti-cu12, nvidia-cuda-nvrtc-cu12, nvidia-cuda-runtime-cu12, nvidia-cudnn-cu12, nvidia-cufft-cu12, nvidia-cufile-cu12, nvidia-curand-cu12, nvidia-cusolver-cu12, nvidia-cusparse-cu12, nvidia-cusparselt-cu12, nvidia-nccl-cu12, nvidia-nvjitlink-cu12, nvidia-nvtx-cu12, sympy, triton, typing-extensions
Required-by: mflux
```

torch brings in these transitive dependencies only when resolved on Linux: `nvidia-cublas-cu12, nvidia-cuda-cupti-cu12, nvidia-cuda-nvrtc-cu12, nvidia-cuda-runtime-cu12, nvidia-cudnn-cu12, nvidia-cufft-cu12, nvidia-cufile-cu12, nvidia-curand-cu12, nvidia-cusolver-cu12, nvidia-cusparse-cu12, nvidia-cusparselt-cu12, nvidia-nccl-cu12, nvidia-nvjitlink-cu12, nvidia-nvtx-cu12`

This change would:

1. cause some backwards breakage for users of the two workfows (depth, lora converter), going forward they have to opt in to the `[torch]` extra - minimal pain, I think
2. enable most macOS users and cloud deployments to stay light weight

I feel would be a net improvement.

Impact to tests: maintainers would test with all extras installed, so no expectations change.

If we `uv sync --extra torch` and then `uv sync` (no extra this time) - the diff is -6 packages on macOS (and a lot more on Linux)

```
Resolved 92 packages in 0.78ms
Uninstalled 6 packages in 885ms
 - jinja2==3.1.6
 - markupsafe==3.0.2
 - mpmath==1.3.0
 - networkx==3.5
 - sympy==1.14.0
 - torch==2.7.1
 - ```

### DIFF
--- a/src/mflux/models/depth_pro/weight_handler_depth_pro.py
+++ b/src/mflux/models/depth_pro/weight_handler_depth_pro.py
@@ -7,7 +7,7 @@ import mlx.core as mx
 try:
     import torch
 except ImportError:
-    print("torch not found. Please install with `pip install mflux[torch]`")
+    print("Extra dependency:torch required this workflow. Please install with `pip install mflux[torch]`")
 from mlx.utils import tree_unflatten
 
 from mflux.ui.defaults import MFLUX_CACHE_DIR
--- a/src/mflux/weights/lora_converter.py
+++ b/src/mflux/weights/lora_converter.py
@@ -5,7 +5,7 @@ import mlx.core as mx
 try:
     import torch
 except ImportError:
-    print("torch not found. Please install with `pip install mflux[torch]`")
+    print("Extra dependency:torch required this workflow. Please install with `pip install mflux[torch]`")
 from mlx.utils import tree_unflatten
 from safetensors import safe_open
 
--- a/uv.lock
+++ b/uv.lock
@@ -729,7 +729,6 @@ dependencies = [
     { name = "sentencepiece", marker = "python_full_version < '3.13'" },
     { name = "tokenizers", marker = "python_full_version >= '3.13'" },
     { name = "toml" },
-    { name = "torch" },
     { name = "tqdm" },
     { name = "transformers" },
     { name = "twine" },
@@ -740,6 +739,9 @@ dev = [
     { name = "mlx" },
     { name = "pytest" },
     { name = "pytest-timer" },
+]
+torch = [
+    { name = "torch" },
 ]
 
 [package.metadata]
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by anthonywu** [comment]
update: got the mlx to startup in MLX + Linux + CPU mode, and received:

```
Extra dependency:torch required this workflow. Please install with `pip install mflux[torch]`

Extra dependency:torch required this workflow. Please install with `pip install mflux[torch]`
```

I think torch is ultimately called by one of the tokenizer loaders and we can't spare it.

filipstrand:
Very nice initiative! Ever since you recommended modal I've also been thinking of trying this out but never got the time to do it. 

> I think torch is ultimately called by one of the tokenizer loaders and we can't spare it.

Yes this is true. I think this is our only Torch dependency when it comes to image generation (for the Depth Pro model, I think we also import torch to convert the tensors to the appropriate format). Way back, I remember starting to implement the tokeniser themselves in MLX too, but decided to skip it due to complexity. Not sure if I remember this correctly, but I think the mlx-examples repo may have their own fully native implementations of the tokenizers.

filipstrand:
Maybe it could actually be worth having another look at implementing the tokenizers... If I remember correctly, I think one of them (maybe CLIP) was mostly straight forward to port, but T5 looked messy, or the other way around

PR #255, opened 2025-08-22
add qwen 2.5 vl text encoder

mzbac:


src/mflux/generate_qwen.py
```diff
@@ -39,13 +84,14 @@ def main():
         for seed in seeds:
             # 2. Generate an image for each seed value
             print(f"🎨 Generating {height}x{width} image with seed {seed}")
+            start_time = time.time()
             image = qwen.generate_image(
                 seed=seed,
                 prompt=prompt,
                 prompt_embeds=prompt_embeds,
-                negative_prompt_embeds=negative_prompt_embeds,
+                negative_prompt_embeds=None,
```
mzbac: It seems that without negative prompt embeddings, it still works for text-to-image generation, but causes degradation in following instructions when rendering text in images.

filipstrand: Yes saw your comment [here](https://github.com/filipstrand/mflux/pull/252#issuecomment-3216124430). This sounds important to always include then if it impacts quality a lot if omitted... Maybe we should have some extra logic around it if the user does not explicitly specify it in the command line interface? In the official example https://huggingface.co/Qwen/Qwen-Image , there they explicitly encode an empty string:

```
negative_prompt = " " # using an empty string if you do not have specific concept to remove
```

Sound like this could maybe be a candidate for the current `prompt_cache` functionality, if we always used a fixed string like this in some cases.

mzbac: I will dig a bit deeper for the negative promp embed, never used it before, using empty string's embedding kind of make no sense to me, but we never know 🤔

filipstrand: Agree. I think FLUX had support for negative prompt embeds too but I never used them and remember reading somewhere that they weren't really needed in practice, so I skipped including them in the original port. But it feels very different in this case I guess if it really messes up the image...

mzbac: I have tried a few different negative prompts, but didn't find much. However, without them, the text rendering tends to become bad quickly, so it seems safe for us to keep them, even if it reduces the generation speed. I'm not sure how difficult it would be to run the prompt and negative embeddings in batch, in that way we might be able to do one forward pass instead of two. But that can be a future optimization.

filipstrand: Im thinking the same - I think I will default to always use it, and fill in some default "empty" value if the user does not specify the negative prompt.

Yes that might be an option... I'm a bit worried about the memory requirements though, but it is something to look into in the future as you say (batch processing would be cool to support in general).

PR #257, opened 2025-08-27
Qwen Image Support

filipstrand:
Resuming work from https://github.com/filipstrand/mflux/pull/252 and https://github.com/filipstrand/mflux/pull/255

I've also started to do some cleanup / restructuring now that we support more models. Part of the functionality is model agnostic and could be shared, but to keep things simple for now, I have separated as much as possible away from Flux. A proper unification with the correct abstractions will take time and not worth the effort with this initial release

Structure is now the following: 

```
models/
├── depth_pro/
│   ├── model/
│   └── weights/
├── flux/
│   ├── model/
│   │   ├── flux_text_encoder/
│   │   ├── flux_transformer/
│   │   ├── flux_vae/
│   │   ├── redux_encoder/
│   │   └── siglip_vision_transformer/
│   ├── tokenizer/
│   ├── variants/
│   │   ├── concept_attention/
│   │   ├── controlnet/
│   │   ├── depth/
│   │   ├── dreambooth/
│   │   ├── fill/
│   │   ├── in_context/
│   │   ├── kontext/
│   │   ├── redux/
│   │   └── txt2img/
│   └── weights/
└── qwen/
    ├── model/
    │   ├── qwen_text_encoder/
    │   ├── qwen_transformer/
    │   └── qwen_vae/
    ├── tokenizer/
    ├── variants/
    │   └── txt2img/
    └── weights/
```

A lot of existing functionality is currently broken due to the refactor, but the image tests should cover all existing functionality, so will deal with that once Qwen feels ready


filipstrand:
@ivanfioravanti I've noticed that the images always turns out a bit soft (I think I read other people have noticed the same for this model). When you tested it out on MPS/Diffusers, have you had a similar experience?

ivanfioravanti:
Yes, I confirm model tends to use blur heavily and be soft. Even if you ask flor clear and sharp image you always get only a small portion in focus and the rest blurred.

filipstrand:
Thanks for confirming! Then it should not be related to the current implementation at least 👍

mzbac:
> Finally a lot nicer cleaner setup compared to before.
> 
> @mzbac By the way, just out of curiosity, have you had a chance to have a look at the edit model yet? (I have not had the time yet but will have a look once everything feels finished with the regular one). My hope is that most things are pretty similar.

I haven't looked into the details, but from a quick glance, it seems only the image and text go through the Qwen VL encoder and image VAE encoder, rest just difference in processing. it shouldn't require any changes to the models, but I will look at it in a bit details in the weekend..

filipstrand:
@mzbac I haven't looked into it, not even fully sure which is "correct" here (unlike when I had an obvious bug before, where I could tell the text rendering was broken) but could be worth investigating... For simplicity in this project, I have aimed to copy the diffusers implementation as close as possible. I guess there could be various kinds of optimisations and similar things that causes the difference

filipstrand:
Current VAE encoder is broken and I am to fix that now, once that is done I think the regular text2img / img2img is more or less done (+ some cleanup related things that has done be done). Have not tested many LoRAs with this model, but the 4bit/8bit mentioned above seems to be working. 

There are a few things we don't have automatic tests for, such as loading quantised models from disk and things, that should also be confirmed before we are completely done with this feature.

filipstrand:
VAE encoder looks good now. This is what a roundtrip gives:

```
encoded = vae.encode(img_original)
img_reconstructed = vae.decode(encoded)
compare(img_original, img_reconstructed)
```

Left: img_original, Right: img_reconstructed 
Screenshot 2025-08-31 at 10 43 21

Hard to tell but reconstructed image typically has a bit less color saturation (and maybe slightly softer). Same thing with FLUX. Otherwise pretty much identical.

anthonywu:
I just thought this would be a good first use case for a namespaced package called `mflux-qwen`. It could be a new repo that follows the structure you suggested here, but the entire qwen structure is considered an official plugin. What do you think? If you want to init a `mflux-qwen` repo under your username, I can help you bootstrap it.

The user experience will be `pip install mflux mflux-qwen` if it's not bundled by default, but of course you can just require the side package as a default which does what you intend here. Choices!

filipstrand:
> I just thought this would be a good first use case for a namespaced package called mflux-qwen. It could be a new repo that follows the structure you suggested here, but the entire qwen structure is considered an official plugin. What do you think? 

That could be an option, right now I'm leaning towards just having multiple model support directly in one and the same repo, but that comes with tradeoffs...  

Would be interesting to hear your take on the following:

- The best way to share lower-level functionality between models if having sepeate repos (e.g model components, AttentionUtils etc)? - Right now, a lot of lower level detals are isolated to each specific implmentation of Qwen and Flux, but with a little bit of work, some parts here can most likley be shared. Along the same line: Some features, e.g Dreambooth LoRA training, feels like they should be plug-and-play with any model (not really true at the moment) and one question would be where such functionality would live? 
- Given the pace of the field, it feels reasonable that more models will come along that we would want to support, would a new repo per model be ideal? 
- Since we started with FLUX, it feels natural to think of that as the "main" model family, but maybe our view-point should shift so that we see Flux as just another model family (with variants) that we support, and that qwen is just as equal in status and not a "extra/plugin" (but then what about the project name, see next point)? 
- If we did just continue with the existing repo, I guess it starts to become weirder now to refer to it as a Flux-based project, especially if we would support more (and possible different) models over time. Maybe it should have been more general, like what @Blaizzy has done with [mlx-vlm](https://github.com/Blaizzy/mlx-vlm). I also just googled and saw that [mlx-image](https://github.com/riccardomusmeci/mlx-image) already exists. In hindsight, a more general name like that, or what they had with [DiffusionKit](https://github.com/argmaxinc/DiffusionKit), would probably have been better. I picked the MFLUX name because I thought it sort of suggested "Mac-FLUX"  and also spelled MLX with every other letter :) 

A somewhat related discussion - I would like to get your opinion on the CLI:

For image generation,  I guess it feels quite natural to simply specify the model like so

```
mflux-generate --model schnell --prompt "Luxury food photograph" --steps 2 --seed 2 -q 8
mflux-generate --model qwen --prompt "Luxury food photograph" --steps 2 --seed 2 -q 8
```
(i.e we wouldn't have a `mflux-generate-qwen` command or similar)

but for the upcoming qwen edit model, how should that work? (I'm thinking especially compared to Flux Kontext - there we currently have the `mflux-generate-kontext` command). Instead of the kontext-specfic command, maybe a more natural command would be a `mflux-edit` command like so: 

```
mflux-edit \
  --model kontext <-- could also be just "flux" here, but maybe kontext is more direct?
  --image-path "house.png" \
  --prompt "Make the house red" \
  ...

mflux-edit \
  --model qwen 
  --image-path "house.png" \
  --prompt "Make the house red" \
  ...
```

We have some other specific editing mdoels too (CatVTON and the older In-Context LoRA and IC-Edit) but I personally consider these more experimental/ "lower status" compared to kontext and qwen edit, so maybe best to keep them separate...

src/mflux/generate.py
```diff
@@ -24,7 +25,8 @@ def main():
         args.guidance = ui_defaults.GUIDANCE_SCALE
 
     # 1. Load the model
-    flux = Flux1(
+    model_class = QwenImage if "qwen" in args.model.lower() else Flux1
```
filip-dema, line 28: Not sure how we would handle the general case if we decide to split up repos going forward, but I put this here for now to make `generate.py` compatible with Qwen too so we could remove the hardcoded file `generate_qwen.py` used during development

src/mflux/generate.py
```diff
@@ -33,14 +35,15 @@ def main():
...
+    memory_saver = CallbackManager.register_callbacks(args=args, model=model)
 
     try:
         for seed in args.seed:
             # 3. Generate an image for each seed value
-            image = flux.generate_image(
+            image = model.generate_image(
                 seed=seed,
                 prompt=get_effective_prompt(args),
+                negative_prompt=get_effective_negative_prompt(args),
```
filipstrand, line 46: Wasn't 100% sure how to best handle this, I guess it would not be compatible stdin as we have it now?

src/mflux/generate.py
filipstrand, line 28: Maybe we will update here regarding how we handle the general case of model selection if we decide to split up repos going forward, but I put this here for now to make generate.py compatible with Qwen too so we could remove the hardcoded file generate_qwen.py used during development

pyproject.toml
```diff
@@ -26,25 +26,31 @@ maintainers = [{ name = "Filip Strand", email = "strand.filip@gmail.com" }]
 license = { file = "LICENSE" }
 requires-python = ">=3.10"
 dependencies = [
+    "accelerate>=0.31.0",
```
filipstrand, line 29: Note to self: Cleanup these once done with the port

filipstrand:
Note to self: Once happy with the PR, run tests with a strict `MFLUX_IMAGE_MISMATCH_THRESHOLD` just to see everything works as expected (should not be any differences in existing FLUX tests)

filipstrand:
Having a little trouble with correct LoRA assignment at the moment so back to debugging... Right now it produces an oversaturated image:
image_13

I first thought was related to some scaling issue, but not sure at the moment...

Also the weight loading code for LoRA is horrible, but my first priority right now is to solve this, then set up a unit test and once everything looks good start to simplify / prune the actual implementation.

filipstrand:
Edit model seems to be working now based on a few simple tests (though there might still be bugs present - e.g. subtle errors in the tokeniser can still produce ok-ish looking images):

Screenshot 2025-09-14 at 09 40 21

From what I can tell, negative prompt is not needed in the same way for the edit model and skipping the negative noise calculation saves us much time at inference (This can always be added back if needed).

This will be the last feature before we can finally merge this big PR. Will do some of the remaining cleanup today and add some tests for this model too. 

Also, in order to not spend too much time on the CLI before release, I think I will go with the proposed updates [mentioned above](https://github.com/filipstrand/mflux/pull/257#issuecomment-3240376297): 

```
mflux-edit \
  --model kontext <-- could also be just "flux" here, but maybe kontext is more direct?
  --image-path "house.png" \
  --prompt "Make the house red" \
  ...

mflux-edit \
  --model qwen 
  --image-path "house.png" \
  --prompt "Make the house red" \
  ...
```

since the interface/api is similar to Kontext (and keep the old `mflux-generate-kontext` around for a little while).

filipstrand:
Some quick updates on this PR: 

Due to work + travel, I've not made any progress on this PR for the last 3 weeks, but I'm eager to resume work in the coming days after I get home. 

Since this PR has been open for a while, I'm thinking of breaking this up a bit and merge the non-edit model first (+ the overall refactor) and make an official release. After that, the edit model could follow in a minor release update (as I remembered, there were some non-obvious bugs there that caused some problems that are yet to be resolved).

PR #259, opened 2025-08-30
Python 3.13 sentencepiece and torch unblocked

anthonywu:
As of `sentencepiece` 0.2.1 from Aug 21 and `torch` 2.8.0 from Aug 6, Python 3.13 usage of `mflux` upstream requirements are unblocked.

Going forward, I think we can recommend Python 3.13 as a supported runtime version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,15 +36,13 @@ dependencies = [
     "pillow>=11.0,<12.0; python_version>='3.13'",
     "platformdirs>=4.0,<5.0",
     "safetensors>=0.4.4,<1.0",
-    # python 3.13 workaround for now:
-    # use temporary community build of py13 wheel, use until official project build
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -175,4 +173,4 @@ error_summary = true
 ignore_missing_imports = true
 implicit_optional = true
 # we should feel free to use new hinting features even if we back support to earlier versions
-python_version = 3.12
+python_version = "3.12"
```
anthonywu, line 176: this was gracefully accepted before but should be a string

anthonywu (commented):
Feels good to remove hacks from docs and configs.

filipstrand:
Nice that you followed up on this! I remember I looked at this earlier in the summer, but it was still blocked back then

PR #263, opened 2025-09-08
Update tests/image_generation image comparison logic to allow similar images that are _close enough_

anthonywu:
Finally got annoyed with the constantly shifting image output expectations between mlx updates, so here's my attempt to be more forgiving while passing the _eye test_

# AI Usage

Some help via chat, but it didn't write any of this PR's code. Everything was typed out.

# Strategy

1. `np.testing.assert_array_equal` -> `np.isclose` (via `image_compare.py` util)
1. see inline comments for the allowance thresholds used
1. if the diff is found, generate the diff image to allow dev to spot check where the diffs are

# Tests

On my machine, the PR began with nothing passing.

After: `pytest tests/image_generation/` all passes. 

my lib versions during test:

```
mlx==0.27.1
mlx-metal==0.27.1
```

I updated the depth pro ref image because my latest `ml-depth-pro/get_pretrained_models.sh` may have updated itself?

```
ls -l ~/Library/Caches/mflux/depth_pro/depth_pro.pt
-rw-r--r--  1 anthonywu  staff  1904446787 Sep  7 21:11 ~/Library/Caches/mflux/depth_pro/depth_pro.pt
```


tests/image_generation/helpers/image_compare.py
```diff
@@ -0,0 +1,46 @@
...
+    mismatch_threshold: float = ENV_MISMATCH_THRESHOLD,
+):
+    image1_path = Path(image1_path)
+    image2_path = Path(image2_path)
+    image1_data = np.array(Image.open(image1_path))
+    image2_data = np.array(Image.open(image2_path))
+    closeness_array = np.isclose(
+        image1_data,
+        image2_data,
+        rtol=float(os.environ.get("MFLUX_IMAGE_ALLCLOSE_RTOL", 0.1)),
```
anthonywu, line 32: if the image output R/G/B elements deviated from reference element by more than 10% of value 0-255

tests/image_generation/helpers/image_compare.py
```diff
@@ -0,0 +1,46 @@
...
+        atol=0,  # ignore absolute tolerance
```
anthonywu, line 33: we'd use `atol` or `rtol` but not both, I used `rtol` so we can think relatively in terms of magnitude of diff

tests/image_generation/helpers/image_compare.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+    num_mismatched = np.count_nonzero(~closeness_array)
```
anthonywu, line 35: I am reading this as: the number of pixels that are more than 10% diff (via `rtol` value.

tests/image_generation/helpers/image_compare.py
```diff
@@ -0,0 +1,46 @@
...
+    total_elements = image1_data.size
+    mismatch_ratio = num_mismatched / total_elements
+    if mismatch_ratio > mismatch_threshold:
```
anthonywu, line 38: so if this ratio of pixels are more than `rtol` off, we fail the test

tests/image_generation/helpers/image_generation_concept_test_helper.py
```diff
@@ -117,15 +117,15 @@ def assert_matches_reference_image_concept_from_image(
...
+            check_images_close_enough(
+                output_heatmap_path,
+                reference_heatmap_path,
+                "Generated concept from image heatmap doesn't match reference heatmap.",
             )
 
         finally:
             # cleanup
-            if os.path.exists(output_heatmap_path):
+            if os.path.exists(output_heatmap_path) and "MFLUX_PRESERVE_TEST_OUTPUT" not in os.environ:
```
anthonywu, line 128: `export MFLUX_PRESERVE_TEST_OUTPUT=anyvalue` to preserve the test run's output images, which allows devs to open both images up for side by side testing

previously, the output was immediately removed even if the test failed, leaving the dev guessing how/why the test failed

tests/image_generation/helpers/image_generation_test_helper.py
```diff
@@ -53,15 +52,14 @@ def assert_matches_reference_image(
...
+            check_images_close_enough(
+                output_image_path,
+                reference_image_path,
+                "Generated image doesn't match reference image.",
             )
-
         finally:
             # cleanup
-            if os.path.exists(output_image_path):
+            if os.path.exists(output_image_path) and "MFLUX_PRESERVE_TEST_OUTPUT" not in os.environ:
```
anthonywu, line 62: this snippet is repeated, but I don't want to scope creep this PR with other refactorings

pyproject.toml
```diff
@@ -137,7 +137,7 @@ docstring-code-line-length = "dynamic"
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 python_files = "test_*.py"
-addopts = "-v --exitfirst --failed-first --showlocals --tb=long --full-trace"
+addopts = "-v --failed-first --showlocals --tb=long --full-trace"
```
anthonywu, line 140: `--exitfirst` was running one test at a time, I think when many tests are failing most devs would want to see comprehensive test suite. One can always reactivate with `pytest --exitfirst` anytime.

tests/image_generation/helpers/image_compare.py
```diff
@@ -0,0 +1,46 @@
...
+    image1_path: str | Path,
+    image2_path: str | Path,
+    error_message_prefix: str,
+    mismatch_threshold: float = ENV_MISMATCH_THRESHOLD,
+):
+    image1_path = Path(image1_path)
+    image2_path = Path(image2_path)
+    image1_data = np.array(Image.open(image1_path))
+    image2_data = np.array(Image.open(image2_path))
+    closeness_array = np.isclose(
```
anthonywu, line 29: I started with [np.testing.assert_allclose](https://numpy.org/doc/2.2/reference/generated/numpy.testing.assert_allclose.html#) but that was not sophisticated enough to allow for two levels of testing (how far the pixels drifted + how many pixels drifted). `isclose` is the same thing but gives you the result instead of immediate AssertionError

anthonywu (commented):
And my intended across-PR workflow is that once a change passes these tests. The authors can send a diff PR to re-establish the reference images based on a new `mlx` checkpoint.

tests/image_generation/helpers/image_generation_concept_test_helper.py
filipstrand: Like it! Every time I needed to regenerate new reference images I always commented out the deletion part - not really scalable in the long term.

pyproject.toml
filipstrand: Agree!

filipstrand:
Not related to this PR, but on the topic of the test code: Clearly, there is a lot of things shared between the various `assert_matches_reference_image` helper implementations that I've also thought about refactoring sometime, but I think it is better to save that work for later (for example, in the Qwen branch I added support for negative prompts there too, which is something a more general helper method should support as well)

anthonywu:
When I was tuning the parameters, I started with small thresholds and looked at the output and reference images when tests fail. The failed test cases can pass my eye test, but the mismatch threshold can be as high as 15%, which was surprising to me. There was one exception, which is the heat map comparison, I decided to individually allow a higher threshold for that one because it's a special case. Notice I allow environment variables to tune the parameters anytime.

filipstrand:
👍 Yes I saw your `MFLUX_IMAGE_MISMATCH_THRESHOLD` which is really nice to have, I think I will use it in my existing Qwen branch where I expect no diffs at all in the FLUX tests on my machine

PR #272, opened 2025-11-07
Add troubleshooting section for hf_transfer installation

Q3g:
I discovered this project yesterday and encountered some issues during installation due to unfamiliarity with `uv`. 
When running `uv tool install --upgrade mflux` and `mflux-generate --model schnell --prompt "Luxury food photograph" --steps 2 --seed 2 -q 8`, I got the following error:
`ValueError: Fast download using 'hf_transfer' is enabled (HF_HUB_ENABLE_HF_TRANSFER=1) but 'hf_transfer' package is not available in your environment. Try pip install hf_transfer.`
  For users who prefer not to manually set environment variables or are new to `uv`, this can be confusing and prevents the installation from being truly "out-of-the-box."
So I added a collapsible troubleshooting section under the installation guide. 

## Summary by CodeRabbit

* **Documentation**
  * Added troubleshooting guidance for Hugging Face transfer download errors with installation instructions to enable faster model downloads.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
     "filelock>=3.20.1",
     "fonttools>=4.60.2",
     "huggingface-hub>=1.1.6,<2.0",
+    "hf-transfer>=0.1.9,<1.0",
     "matplotlib>=3.9.2,<4.0",
     "mlx>=0.27.0,<0.31.0; sys_platform=='darwin'",
     "mlx[cuda13]>=0.30.3,<0.31.0; sys_platform=='linux'",


README.md
```diff
@@ -137,6 +137,21 @@ For more details and troubleshooting, see the [completions documentation](src/mf
...
 </details>
 
+<details>
+
+<summary>⚠️ Troubleshooting: hf_transfer error</summary>
+
+If you encounter a `ValueError: Fast download using 'hf_transfer' is enabled (HF_HUB_ENABLE_HF_TRANSFER=1) but 'hf_transfer' package is not available`, you can install MFLUX with the `hf_transfer` package included:
+
+```sh
+uv tool install --upgrade mflux --with hf_transfer
```
anthonywu: I think an even more durable solution is just to include it in the project dependencies in `pyproject.toml`

https://pypi.org/project/hf-transfer/#history 

`hf-transfer>=0.1.9,<1.0` in this list: https://github.com/filipstrand/mflux/blob/main/pyproject.toml#L28

The library has been out for a while and it's maintained by HuggingFace so I feel comfortable just including it as a universal dependency

anthonywu (commented):
My suggestion here is to just include it in the library instead of expecting users to come back to a readme.

If you're interested in making this change, just update this PR. Otherwise we'll just do this at the next release.

README.md
Q3g: Thanks for the review!

My main hesitation was that I wasn’t entirely sure why transformers itself doesn’t include hf-transfer as a direct dependency, so I kept the previous behavior to avoid unintended issues. But given that we mostly target macOS, I agree that adding hf-transfer directly should be fine and unlikely to cause any problems.

I’ve updated the PR accordingly. Thanks again!

PR #279, opened 2025-11-21
Add Bria FIBO support

filipstrand:
Standalone PR (excluding the debugger) for the FIBO model now that it works as expected. 

Still some cleanup left to do and some documentation before release 

---
Part of the initial implementation was written with the help of Composer 1, but majority of time so far (couple of hours) was spent on debugging subtle value errors, tried GPT-5.1 and after a while it was able to hunt down the bug!

Early success - VAE output matching now (given same latents from PyTorch):

MLX left, PyTorch right

Screenshot 2025-11-13 at 23 13 18

Update: VAE roundtrip now also works

Update: Transformer denoise loop now working! 

Screenshot 2025-11-16 at 00 22 48

Update: Slight differences in image outputs but probably due to precision, both looks equally valid IMO (given equivalent text encoder inputs), last part to port is the text encoder...

Text encoder almost done but some obvious bugs still left: Pretty amusing how the result can still be conceptually correct yet still very wrong 😅 ‍(have narrowed this down to tokenisation issue...) 
Screenshot 2025-11-17 at 12 12 34

Update: Text encoder fixed - Everything finally working end to end in MLX 🎉 

Now the fun part of actually cleaning up and refactoring

Update: VLM also left to port (forgot about that one), looking into it now...could also potentially use existing VLM implementation https://github.com/Blaizzy/mlx-vlm/tree/main/mlx_vlm/models/qwen3_vl

Update: VLM now implemented, only the GENERATE part so far, will look into INSPIRE and REFINE...

Update: Vision part of VLM now working so both GENERATE and INSPIRE produces sensible results compared to PyTorch reference!

(prompt) Refine works as expected now: Example below `make the owl white color but keep everything else exactly the same` 
Screenshot 2025-11-18 at 22 15 11


filipstrand:
I'm working on separating the Fibo VLM from the generative part, at first I had these more closely linked together, but that was probably a mistake - The way I see it now is that the VLM is a very optional part of the process and should not interfere with the generative model (which always will assume JSON input).

---

Some other overall thoughts: 

Now that we have 3 big models in the project (along with a new VLM, new weight handling etc), I see tons of opportunity to refactor and take a fresh stab at trying to consolidate things. Weight handling, in particular, has always been a pain to work with and probably something that is on the top of my list to improve. This will require lots of work, but can be done incrementally, and I think we have decent test coverage of the end result for all models we support, so should be easy to catch regressions here.

Also, due to some implementations being heavily AI generated (the Fibo VLM for example), part of the of new code is very hard to read and understand and not something I'm particularly proud of. But as part of the porting process, I do enjoy being able to relatively quickly get a working MLX implementation via AI, that then can be firmly "locked down" by a few image-tests and then iterated upon safely to arrive at a much more minimal/direct implementation.

PR #284, opened 2025-11-29
Add support for Z-image Turbo & major improvements for weight loading and parameter resolution

filipstrand:
Biggest refactor made in a single PR so far (not review friendly unfortunately). Without any doubt, this would not have been possible without Opus 4.5: 

- All critical weight handling have been rewritten unified (all models now use this new approach, including flux) 
- Everything is declarative (helps a ton when struggling with weight mapping for new models to be implemented... before, I had the coding agent rewrite custom weight handling every time which was not a sustainable or scalable approach)
-  I also gave a shot at reformulating how we actually resolve things like which model to use (local vs hf), which lora (local, library vs hf), pre-quantized vs "on the fly"), which config etc. Like for weight loading, I also tried a more declarative "rule based" approach here. This can probably be improved, but I find it much better than what we had before. 

This work is done to prepare for faster and more convenient model-porting work in the future  

---

Crazy good for such a small model!! Runs faster than Flux and produces way better images overall!

Simple architecture + Opus 4.5 made this port a breeze in comparison to earlier ones!

Img2Img, LoRA, and Quantisation all supported out of the box.

Supposedly also easy to train, so will add Dreambooth support also (along with FIBO too)! 

A few LoRAs have been popping up, big fan of this one:
https://huggingface.co/renderartist/Technically-Color-Z-Image-Turbo

```
mflux-generate-z-image-turbo \
  --prompt "t3chnic4lly vibrant 1960s close-up of a woman sitting under a tree in a blue skirt and white blouse, she has blonde wavy short hair and a smile with green eyes lake scene by a garden with flowers in the foreground 1960s style film She's holding her hand out there is a small smooth frog in her palm, she's making eye contact with the toad." \
  --width 1280 \
  --height 720 \
  --steps 9 \
  --quantize 8 \
  --seed 456 \
  --lora-paths "renderartist/Technically-Color-Z-Image-Turbo" \
  --lora-scales 0.5 \
  --output ~/Desktop/img.png
```
frog_girl_2

I've also chosen to do a big restructure of the packages (feels like it's time). The total diff is not very nice, but I'm relatively confident that nothing major is broken. Will also do a bunch of manual testing to make sure our commands work as expected... 

I'm starting to believe that docstrings are completely unnecessary going forward - I've maybe gone too far here but I've chosen to remove them all completely with the motivation that I almost always interact with the code through a coding agent now, and the agents seems more and more capable of understanding the code by actually reading the source itself. In this way, there is no risk of them getting out of sync. And if something is unclear to human user, just ask the agent or improve the method names/api/interface. In a similar spirit, I'm also less convinced of having README:s all over (although `.cursorrules` and similar actually feels worth it, but I've yet to take the time to actually write one properly - it feels like actual human investment here could be worth it and not just auto-generating some generic thing).  Lets see how it goes :)


mzbac:
Great work, @filipstrand. Just wondering, do you have a benchmark for z-image-turbo on mflux? I haven't had time to look into the details, but I feel the Swift port is somehow only slightly faster on the MPS backend. Not sure if there is any performance difference between Python mlx and Swift mlx.

filipstrand:
@mzbac Thank you and likewise with your swift implementation! I haven't actually done any comparisons myself either yet. I do see a big speedup when running it quantised in MLX, and I'm honestly not sure how that works in MPS nowadays...

mzbac:
> @mzbac Thank you and likewise with your swift implementation! I haven't actually done any comparisons myself either yet. I do see a big speedup when running it quantised in MLX, and I'm honestly not sure how that works in MPS nowadays...

Cool, thanks for the hint the mflux has speed up in quant. I will do some benchmark tests, maybe I have done something wrong in the Swift port.

filipstrand:
@mzbac Oh, really nice to see! Also good confirmation that there is no real difference in practice between MLX python vs swift!

PR #287, opened 2025-12-02
Add torch dependency optimization analysis and proposal

jiangdi0924:
- Comprehensive analysis of torch usage in mflux codebase
- Found torch is only used for weight conversion, not inference
- Proposed optional dependency approach to reduce disk usage
- Created torch compatibility layer for graceful degradation
- Added migration examples and implementation guide

Key findings:
- Torch wheel size: ~75 MB (macOS ARM64)
- Installed size: ~300 MB (with dependencies)
- Usage: Only basic tensor operations for weight conversion
- Optimization: Make torch optional via extras_require

Benefits:
- Users can choose minimal installation (~200 MB saved)
- Full backward compatibility with mflux[all]
- Clear feature boundaries and better error messages
- Paves way for future torch removal

### DIFF
--- a/EXAMPLE_MIGRATION.md
+++ b/EXAMPLE_MIGRATION.md
@@ -1,0 +1,396 @@
+# 代码迁移示例
+
+这个文档展示如何将现有代码迁移到使用可选的 torch 依赖。
+
+## 示例 1: qwen_weight_handler.py
+
+### 修改前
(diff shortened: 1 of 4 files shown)


jiangdi0924:
Sorry, please ignore this PR—it was a small mistake.

PR #293, opened 2025-12-06
Fix/lora save bloat

filipstrand:
Fix for https://github.com/filipstrand/mflux/issues/217#issuecomment-3615321206

We now bake LoRA deltas into base linear layers right before saving and strip the wrappers, preventing transformer shards from ballooning when LoRAs are applied. This is isolated to the save path (runtime LoRA behavior unchanged) and guarded by a regression test that compares shard counts between base and LoRA saves.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.13.3] - 2025-12-06
+
+### 🐛 Bug Fixes
+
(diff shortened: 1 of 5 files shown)


azrahello:
Thanks for this fix, it works really well! I noticed there are some layers that could potentially be added, such as:

- diffusion_model.transformer_blocks.0.img_mod.1.lora_A.weight
- diffusion_model.transformer_blocks.0.img_mod.1.lora_B.weight
- diffusion_model.transformer_blocks.0.txt_mod.1.lora_A.weight
- diffusion_model.transformer_blocks.0.txt_mod.1.lora_B.weight
- diffusion_model.transformer_blocks.1.img_mod.1.lora_A.weight

Great to see the project growing!

filipstrand:
> Thanks for this fix, it works really well! 

Great to hear! Very nice to have more examples of what to add for additional LoRA support. Now it should be super trivial to add these as well in `z_image_lora_mapping.py` for example. If you have a link to a specific one, let me know and I can test that it loads properly

PR #323, opened 2026-01-15
Flux2 Klein support

filipstrand:
4b version in MLX:

flux2_klein_tiger_swim_5_seed_5

---

4b in 4bits: 

flux2_klein_tiger_swim_q4_12

 
---
Edit: I overlooked this initially, but this is actually a omni-model and supports editing natively as well (same weights!), so I've added support for it with multiple images as input (still not super quick as input context is fairly large, but cool that it works in just 4 steps for some cases): 

Screenshot 2026-01-17 at 00 42 30

---

Edit: They also released the [base models](https://huggingface.co/black-forest-labs/FLUX.2-klein-base-9B) (non-distilled) that requires 50 ish time steps. I haven't played around with these yet, but as they will take more time to run, I guess they will be of less direct interest to must users. For fine-tuning, however, this could be interesting (unless ostris/ai-toolkit releases an de-destill adapter, which I'm sure he will). Will look into this for next release as I'm in the process of generalising the LoRA training  

--- 

Edit: For simplicity I'll remove the negative prompt support that was in the original pipeline. From their official [docs] (https://docs.bfl.ai/guides/prompting_guide_flux2 they) bfl claim it does not support it. 

```
No negative prompts: FLUX.2 does not support negative prompts. Focus on describing what you want, not what you don’t want.
```

---

Save functionality added, the 4b in -q 4 comes in at 4.63GB in total, quite nice!


anthonywu (commented):
Love the new skill docs!

PR #351, opened 2026-02-15
Add cache-limit option

icelaglace:
This PR adds `--cache-limit-gb` so users can control MLX cache size without needing `--low-ram` mode.

## Changes
- Added a new CLI option: `--cache-limit-gb`.
- Added input validation for positive float values, to avoid user error, you never know.
- Updated callback memory registration:
  - resolves cache limit from GB to bytes
  - in `--low-ram` mode, passes cache limit to `MemorySaver` (falling back to current default if unset)
  - outside `--low-ram`, applies MLX cache limit directly and clears/resets cache tracking

## Compatibility with current flow
- No behaviour change unless `--cache-limit-gb` is provided.
- Existing `--low-ram` behaviour remains intact.

## Why
This gives high-VRAM users a simpler way to tune cache usage and improve runtime behavior without enabling full low-RAM unloading.
Reason being SeedVR high-resolution upscale, my M2 Ultra can't fit the whole model in memory. `--low-ram` is too limiting my hardware and I have plenty of RAM. 
This allows me to limit the Metal cache to 16 or 32GB, and get a pretty significant speedup without going OOM.

Please let me know if you require any changes to this branch.
Thanks!

### DIFF
--- a/src/mflux/models/common/README.md
+++ b/src/mflux/models/common/README.md
@@ -18,6 +18,7 @@ This README covers stable, shared patterns. For model-specific usage, see each m
 - [Metadata reuse](#metadata-reuse)
 - [Metadata inspection](#metadata-inspection)
 - [Resource and inspection options](#resource-and-inspection-options)
+- [MLX cache limit](#mlx-cache-limit)
 - [Cache locations](#cache-locations)
 - [Capabilities by task](#capabilities-by-task)
 - [Command reference](#command-reference)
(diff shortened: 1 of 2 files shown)


filipstrand:
I think this makes sense to have, and I like that you described your use case. 

I'm trying to keep the cli as minimal as I can, and I'm planning on doing a big cli update relatively soon for a v1 release (with breaking changes) to have things much more standardised and hopefully a bit more simplified, so there is a risk I will update this functionality / naming in the future, but I think it makes sense to have as you suggested it here.  

I had Cursor search the web and see if there was some native ENV variable one could set to achieve this, but it found no such results. We could also imagine going with some kind of MFLUX_CACHE_LIMIT env variable, but let's do the cli flag for now. 

Regarding the name, I like it, but then I remember that we also have our [mflux Disk cache](https://github.com/filipstrand/mflux/blob/main/src/mflux/models/common/README.md#cache-locations) which might confuse the user a bit, so my proposal is to name this something with `mlx` explicitly, for example `--mlx-cache-limit-gb`

icelaglace:
Ah that's nice to know! Thank you for the heads up, and that makes sense. I do like minimalistic CLI, this is one of the reasons why I do like to use MFLUX on the CLI as it's very fast to use without me doing --help to remember every time lol.

Feel free to just keep the PR open for ref in the future, no need to merge it, no worries.
I did the changes you've requested, and added 2 extra tests as a separate commit in case.

Thank you !

PR #352, opened 2026-02-15
Add SeedVR2 7B support

icelaglace:
This PR adds SeedVR2 7B support while keeping existing SeedVR2 3B behavior as the default.

## Changes
- Added a new `seedvr2-7b` model config and resolver.
- Updated SeedVR2 upscale model selection logic to resolve 3B vs 7B from model/model-path inputs.
- Generalized SeedVR2 transformer internals to support both variants:
  - configurable RoPE mode and optional text RoPE
  - configurable MLP type (SwiGLU vs normal GELU MLP)
  - configurable output Ada and last-layer behavior
- Split SeedVR2 weight definitions into dedicated 3B and 7B definitions and resolve at init time.
- Updated weight mapping to support variable block counts and optional bias tensors needed by the 7B variant.
- Modified the already made parser/model-resolution test with 3B/7B selection scenarios.

## Compatibility with current flow
- Default remains SeedVR2 3B unless 7B is requested, similar to how Flux-Klein-4B is default on image generation.
- Existing 3B workflows are unchanged this way.
- 3B still looks the same in output (verified with signal-to-ratio and my own eyes) so no regression AFAIK. But please let me know!

## Why
SeedVR2 users can now run both 3B and 7B variants as to me, 7B is miles ahead of 3B but requires much more RAM.
(Hence why I also made the cache limit option in another PR)

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ MFLUX supports the following model families. They have different strengths and w
 |[Z-Image](src/mflux/models/z_image/README.md) | Nov 2025 | 6B | Distilled & Base | Yes | Best all-rounder: fast, small, very good quality and realism. |
 |[FLUX.2](src/mflux/models/flux2/README.md) | Jan 2026 | 4B & 9B | Distilled & Base | Yes | Fastest + smallest with very good qaility and edit capabilities. |
 |[FIBO](src/mflux/models/fibo/README.md) | Oct 2025 | 8B | Base | No | Very good JSON-based prompt understanding and editability, medium speed |
-|[SeedVR2](src/mflux/models/seedvr2/README.md) | Jun 2025 | 3B | — | No | Best upscaling model. |
+|[SeedVR2](src/mflux/models/seedvr2/README.md) | Jun 2025 | 3B & 7B | — | No | Best upscaling model. |
 |[Qwen Image](src/mflux/models/qwen/README.md) | Aug 2025+ | 20B | Base | No | Large model (slower); strong prompt understanding and world knowledge. Has edit capabilities |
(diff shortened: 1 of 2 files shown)


filipstrand:
Okay, these look really good! Let me test a few more on my side too, might have been something wrong there (or just a one-off thing with that test, not sure at the moment) 

The only thing I did yesterday was to swap the 7b model for the 3b model in `seedvr2_upscale_test_helper.py` and ran it with MFLUX_PRESERVE_TEST_OUTPUT=1 and looked at the output image.

icelaglace:
Aha, it's the model, you were right about the output! 

My previous testings were in x1280. The .py test suite use x1080... the difference is huge just by asking 1110 pixels instead of 1080 pixels, very odd behaviour from the model...

To be sure, I did exactly the same on the stand-alone and yep, looks the same... look at that :

### MFLUX test.py swapped to 7b
output_seedvr2_upscaled_7b_helper_swap

### Stand-alone
output_seedvr2_upscaled_7b_standalone

Both 1080 output.
But if you do the same at 1110 pixels ?

output_seedvr2_upscaled_7b_mycommand_3

Works just fine (Mflux above)...

Do you want me to find a good royalty-free image and add 7B to the tests automatically ? As either a separate test or the same but extended. Let me know what you prefer :)

Thanks!

filipstrand:
@icelaglace Ah this is great, then this was more of an unlucky test case! Now I'm much more confident in the implementation. There was really nothing special with this test image in particular (I took it from here https://huggingface.co/jasperai/Flux.1-dev-Controlnet-Upscaler back when I implemented that method, which is now legacy) and I can update the tests on my end.

PR #353, opened 2026-02-15
feat(z-image): add sigma schedules, aspect ratios, saveinfo, shift, MCF sampler;

terribilissimo:
**Note for Filip**: Feel free to pick what you like from this PR and discard the rest. It's perfectly good to discard this PR and make another with just the things you want to integrate (if any).

**Update**: This PR now also includes aspect-ratio-preserving dimension scaling for img2img workflows.

When a reference image is provided via `--image-path`, you can specify output dimensions (or just one of them) as scale factors:

```
# Scale both dimensions by 1.2× (aspect ratio preserved)
mflux-generate-z-image-turbo \
  --image-path photo.jpg --image-strength 0.3 \
  --height 1.2x --steps 9 --seed 42
 
# Absolute width, height auto-computed from aspect ratio
mflux-generate-z-image-turbo \
  --image-path photo.jpg --image-strength 0.3 \
  --width 800 --steps 9 --seed 42
```

Changes (in `dimension_resolver.py`):

When only one dimension has a non-unity ScaleFactor (e.g. --height 1.2x) and the other is at auto/default, the scale is propagated to both dimensions — preserving the reference image's aspect ratio.
When one dimension is absolute pixels and the other is unspecified, the missing dimension is computed from the reference image's aspect ratio.
Also enabled `supports_dimension_scale_factor=True` for `mflux-generate-z-image` (was already enabled for turbo).
All values snap to multiples of 16. Behavior is unchanged when no reference image is provided.

- --shift: override automatic sigma shift (mu) for both schedulers
- --mcf-max-change: MCF sampler that clamps per-step latent changes
- --cosine / --karras / --exponential: alternative sigma schedules
- --aspect: aspect ratio presets with auto-dimension computation
- --saveinfo: descriptive filenames encoding generation parameters

- All features apply to mflux-generate-z-image-turbo and mflux-generate-z-image
- **NOTE: when no fork-specific flags are used, behavior is identical to upstream**

## Summary (about the rest)

This PR adds several quality-of-life enhancements to the Z-Image and Z-Image Turbo
pipelines, ported from our zima.py CUDA/PyTorch scripts. All features are
**opt-in** — when none of the new flags are used, behavior is **identical to upstream**.

### Noise Schedule Control
- `--shift <float>` — Override the automatic sigma shift (mu) value.
  By default mu is computed from image dimensions. Higher values push the
  noise schedule towards higher noise levels.
- `--cosine` — Smooth S-curve sigma schedule (more steps at high/low noise)
- `--karras` — Karras schedule with rho=7 (concentrates steps on fine details)
- `--exponential` — Log-spaced sigmas between sigma_max and sigma_min
  (_Only one of the three can be used at a time; mutually exclusive group_)

### Sampling
- `--mcf-max-change <float>` — MCF (Mean Change Factor) sampler that clamps
  per-step latent changes to prevent sudden jumps. Typical: 0.05–0.50, min/max 0.01/1.00.

### Convenience
- `--aspect <preset>` — Aspect ratio presets (1:1, 4:3, 3:4, 3:2, 2:3, 16:9,
  9:16, 18:9, 9:18, 21:9, 9:21). **If combined with only `--width` or `--height`,
  the missing dimension is auto-computed and rounded to multiples of 16.**
- `--saveinfo` — Save images with descriptive filenames encoding
  timestamp, seed, steps, LoRA, scheduler, and sigma schedule info
  (convenient, you won't have to look at the EXIF or json for quick reproducibility).

## Modified Files
- `src/mflux/cli/parser/parsers.py` — New args, aspect ratio dict, sigma_schedule resolution; removed `--metadata`
- `src/mflux/models/common/config/config.py` — `sigma_schedule` property
- `src/mflux/models/common/schedulers/linear_scheduler.py` — `_generate_base_sigmas()`
- `src/mflux/models/common/schedulers/flow_match_euler_discrete_scheduler.py` — same
- `src/mflux/models/z_image/variants/z_image.py` — New params in `generate_image()`
- `src/mflux/models/z_image/cli/z_image_turbo_generate.py` — Passthrough + saveinfo
- `src/mflux/models/z_image/cli/z_image_generate.py` — Same
- All CLI entry points (flux, flux2, qwen, z_image, fibo) — Removed `export_json_metadata=args.metadata`

## Testing
All features tested on Apple Silicon (M-series) with both `mflux-generate-z-image-turbo`
and `mflux-generate-z-image`. Baseline output is pixel-identical to upstream when
no new flags are used. See TESTING-SCRIPT.txt for the full test matrix (18 tests).

## Checklist
- [x] No new dependencies added
- [x] All sigma schedule math uses pure MLX (`mx.array`, `mx.cos`, `mx.linspace`)
- [x] Backward-compatible — default behavior unchanged
- [x] Passes existing tests
- [x] Both Z-Image and Z-Image Turbo entry points updated

### DIFF
--- a/src/mflux/cli/parser/parsers.py
+++ b/src/mflux/cli/parser/parsers.py
@@ -1,5 +1,6 @@
 import argparse
 import json
+import math
 import random
 import sys
 import time
@@ -10,6 +11,24 @@ from mflux.cli.defaults import defaults as ui_defaults
 from mflux.models.common.resolution.lora_resolution import LoraResolution
 from mflux.models.flux.variants.in_context.utils.in_context_loras import LORA_NAME_MAP
 from mflux.utils import box_values, scale_factor
+
+ASPECT_RATIOS = {
+    "1:1":  (1, 1),
+    "4:3":  (4, 3),
+    "3:4":  (3, 4),
+    "3:2":  (3, 2),
+    "2:3":  (2, 3),
+    "16:9": (16, 9),
+    "9:16": (9, 16),
+    "18:9": (18, 9),
+    "9:18": (9, 18),
+    "21:9": (21, 9),
+    "9:21": (9, 21),
+}
+
+
+def _ceil16(value):
+    return math.ceil(value / 16) * 16
 
 
 class ModelSpecAction(argparse.Action):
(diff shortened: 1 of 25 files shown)


terribilissimo:
Thanks!

Best regards,
A.

m-2k:
hello, when is merging planned?

fxd0h:
Closing this, with a map rather than a shrug.

The schedule and sampler flags here, `--cosine`, `--karras`, `--exponential`, `--shift` and `--mcf-max-change`, do not need to be in the tree to be usable. `--scheduler` already takes a dotted path and imports it: `try_import_external_scheduler` resolves any `your_package.module.YourScheduler` and only asks that it subclass `BaseScheduler`. Your schedules can ship as their own package and run against stock mflux today, with you owning the release cadence instead of waiting on this queue. That is what the February discussion was circling, and it costs you nothing to keep.

`--saveinfo` is close to something mflux already does: `--metadata` writes the settings to a sidecar and embeds them in the PNG, so the information survives, just not in the filename.

`--aspect` is the one thing here with no equivalent, and it is small and self-contained. A PR with only that would be reviewed on its own merits rather than as part of a bundle where the argument is about the other flags.

Sorry it sat this long. @m-2k that also answers your July question.

PR #362, opened 2026-02-20
Align Qwen-Image with diffusers reference

ciaranbor:
Hey, I've been using mflux to add image model support over at [exo](https://github.com/exo-explore/exo). Thanks for the great work!

I've been getting some strange outputs from Qwen-Image at times. I believe I've seen it mentioned in issues here too but I can't find any right now. 

This PR fixes several configuration and numerical differences between the mflux Qwen-Image implementation and the diffusers reference, improving output quality for both txt2img and edit pipelines.

Diffusers also uses `CONDITION_IMAGE_SIZE=1024*1024` but this made generation much slower so I left it as is.

Example before and after outputs for

```
PROMPT = "ronaldo"
NEGATIVE_PROMPT = " "
SEED = 2
HEIGHT, WIDTH = 1024, 1024
NUM_STEPS = 25
GUIDANCE = 4.0
```

main_output
pr_output

 # Changes

  Sigma shift schedule — Qwen-Image was not applying any sigma shift (requires_sigma_shift=None), using a plain linspace noise schedule. The diffusers reference applies a dynamic exponential shift with max_shift=0.9
  and max_image_seq_len=8192, plus a stretch-to-terminal that pins the final sigma at 0.02. The shift parameters are now configurable per model on ModelConfig (defaults match FLUX's existing hardcoded values, so
  other models are unaffected).

  SDPA in text encoder — Replaced manual matmul → mask → softmax → matmul with mx.fast.scaled_dot_product_attention, avoiding materialization of the full attention weight matrix.

  SDPA + float32 ROPE in vision encoder — Same SDPA replacement for both chunked (windowed) and full attention paths. ROPE is now computed in float32 to reduce precision loss from bfloat16 trig operations.

  Edit prompt template — Changed use_picture_prefix from True to False to match the diffusers reference prompt format. With True, images were inserted as Picture 1: <vision_tokens> dynamically; with False, the vision
   tokens are part of the template and the prompt is appended after them.

  Tokenizer config — max_length increased from 1024 to 1058 and padding set to "longest" to match the diffusers reference configuration.


filipstrand:
Tried running this with the Flux2 test image just to see what it produced 

```
uv run mflux-generate-qwen-edit 
--prompt "Make the woman wear the eyeglasses (regular glasses, not sunglasses)" 
--image-paths tests/resources/unsplash_person.jpg tests/resources/glasses.jpg 
--width 1344 
--height 896 
--steps 20 
--guidance 1.0 
--seed 45 46 47 
-q 8 
```

Screenshot 2026-02-20 at 19 11 57

Both are honestly pretty bad in both cases, perhaps more variation in the types of glasses (compared to reference input glasses) in this PR, but I honestly don't know how much it says. 

This is Flu2Klein in comparison - much more consistent, much faster (same settings but 4 steps instead of 20):

```
uv run mflux-generate-flux2-edit 
--model flux2-klein-9b 
--prompt "Make the woman wear the eyeglasses (regular glasses, not sunglasses)" 
--image-paths tests/resources/unsplash_person.jpg tests/resources/glasses.jpg 
--width 1344 
--height 896 
--steps 4 
--guidance 1.0 
--seed 45 46 47 
-q 8 
```

Screenshot 2026-02-20 at 19 19 45

filipstrand:
> Out of interest I ran the test using diffusers and this is what I got

Oh yes, I remember that "zoomed in" effect which also seemed very weird from diffusers (I remember I saw that same effect on other images too)...  That made the whole port even harder since we couldn't necessarily rely on the reference.

PR #376, opened 2026-03-08
Fix Flux2 LoRA mapping coverage

icelaglace:
This fixes Flux Klein LoRA mapping coverage for a bunch of newer HuggingFace LoRA exports that were loading but not actually applying any weights...

The issue showed up with Klein 9B LoRAs like `dx8152/Flux2-Klein-9B-Enhanced-Details`

It matches zero keys for these guys for example.

The fix is to expand `Flux2LoRAMapping` to support additional Klein LoRA key formats:

1. `.default.weight` suffixes for standard `transformer_blocks...` mappings
2. `.default.weight` suffixes for `single_transformer_blocks...` mappings
3. `diffusion_model.double_blocks...` aliases for double-block exports
4. `diffusion_model.single_blocks...` aliases for single-block exports

Also added a test covering these new mapping that LLM generated for me.
```bash
MFLUX_PRESERVE_TEST_OUTPUT=1 uv run python -m pytest -q tests/test_flux2_lora_mapping.py
```
 LoRA match checks :
dx8152/Flux2-Klein-9B-Enhanced-Details: 224 / 224

Tried on LoRAs that worked before and they load just fine too, so no regression found so far.
Also tried more around and they all load now when they didn't before.

Feel free to modify or tell me what to change to your liking :)
Thanks!

### DIFF
--- a/src/mflux/models/flux2/weights/flux2_lora_mapping.py
+++ b/src/mflux/models/flux2/weights/flux2_lora_mapping.py
@@ -20,46 +20,6 @@ class Flux2LoRAMapping(LoRAMapping):
         return targets
 
     @staticmethod
-    def _with_default_weight_patterns(*patterns: str) -> list[str]:
-        expanded_patterns: list[str] = []
-
-        for pattern in patterns:
-            expanded_patterns.append(pattern)
-
-            if pattern.endswith(".lora_A.weight"):
-                expanded_patterns.append(pattern.removesuffix(".lora_A.weight") + ".lora_A.default.weight")
-            elif pattern.endswith(".lora_B.weight"):
-                expanded_patterns.append(pattern.removesuffix(".lora_B.weight") + ".lora_B.default.weight")
-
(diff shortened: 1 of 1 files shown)


filipstrand:
@icelaglace Thanks for the addition. 

Can you try to implement this mapping without these methods

```
_with_default_weight_patterns
_with_diffusion_double_block_patterns
_with_diffusion_single_block_patterns
```

I'm just curious to see how that looks. The mappings now are indeed a bit verbose and they contain a lot of repetition, (but also trivial to edit with coding agents). I'm not quite sure this is the best format going forward, but at least it is super straight forward, and also really easy to glance and just see additions in git and that should mean no regression.

Going forward I might refactor this and then will probably include something like `_with_diffusion_double_block_patterns` etc, but maybe place in another place so that logic could be reused. For now, unless it will be extremely verbose with your additions, I'm totally for duplication and direct clarity in the mappings.

Issue #377, opened 2026-03-08
Calling mflux as a library from mlx-openai-server for flux2 klein 4b to edit results in TypeError: unsupported operand type(s) for %: 'NoneType' and 'int'


TheOrsa:
**Issue type**: Bug

**Description**: Configuring OpenWebUI to call a mlx-openai-server to run image edit with flux2 is not possible in the latest 0.16.8 version.
How to reproduce: Configure an Image edit endpoint in OpenWebUI and try to use it.

**Expected**: Image is uploaded and the prompt is used to return an edited image.

**Actual**: OpenWebUI gives an error for the "backend" 500 (which is mlx-openai-server). The logs on mlx-openai-server show the following:
`TypeError: unsupported operand type(s) for %: 'NoneType' and 'int'`

PR #378, opened 2026-03-08
Fix for #377 to allow image edit with dynamic resolution

TheOrsa:
When running mflux from cli (i.e. mflux-generate-flux2-edit) it works fine and produces an image with the same proportions as the input image.

When running mflix from mlx-openai-server to edit using flux2, it fails with a typical python error message signaling a null (None) vas passed.

While Im not sure which is correct (I assume the OpenAI spec) my workflow is OpenWebUI -> mlx-openai-server -> mflux so I had to track them all and find by myself. This fix should also fix for other models having the same problem.

CLI code calls the DimensionResolver utility to compute dynamically the width and height.
The library code assumes a non-null width and height, which I think is correct, however when calling from OpenWebUI it makes no sense to specify dimensions manually, it also needs to be dynamic.

As a developer of a minimal CLI tool myself, I'd say all the library should have the logic and the CLI be a dummy project implementing a cli parser towards the library's own API. But thats a lot of effort and regardless, introducing changes also has the probelm of breaking things to other users, so I think that is also very important.

### DIFF
--- a/tests/common/config/test_config_dimensions.py
+++ b/tests/common/config/test_config_dimensions.py
@@ -35,3 +35,32 @@ def test_partial_height_none_resolves_from_defaults_without_reference_image():
 
     assert config.width == 640
     assert config.height == ui_defaults.HEIGHT
+
+
+@pytest.mark.fast
+def test_both_dimensions_none_resolve_from_reference_image(tmp_path: Path):
+    image_path = tmp_path / "reference.png"
+    PIL.Image.new("RGB", (1200, 800)).save(image_path)
+
+    config = Config(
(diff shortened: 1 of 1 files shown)


filipstrand:
Hi @TheOrsa 

>  I'd say all the library should have the logic and the CLI be a dummy project implementing a cli parser towards the library's own API.

Yes, I agree with this. Over time the Python API and CLI have diverged and I agree that we should put effort into making them as similar in behaviour and naming as possible. 

I'll link this ticket to the ongoing design discussion for the upcoming CLI & API redesign  https://github.com/filipstrand/mflux/issues/357.

PR #380, opened 2026-03-11
Scale training images according to area, not longest side

waldheinz:
When the user limits the `max_resolution` it's very likely because of resource constaints. But the resource use is strongly correlated to the total number of pixels and not the length of the longer side of an image.

This change modifies the logic to resolve dimensions where the *area* of the resolved image is ~ `max_resoluton ^ 2`, which means non-square images get the same "pixel budget" as square images.

### DIFF
--- a/src/mflux/models/common/training/utils.py
+++ b/src/mflux/models/common/training/utils.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import math
+
 from collections.abc import Iterator
 from contextlib import contextmanager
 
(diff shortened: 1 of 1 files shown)


filipstrand:
I agree, this makes more sense, thanks for considering this use case!

At first, I thought of updating the name `max_resolution` to something more suggestive, but it is simple so let's keep it for now :) 

I'll go ahead and include it for v0.17 which is due soon.

If you have further suggestions for API related things, they are also very welcome here https://github.com/filipstrand/mflux/issues/357

Issue #400, opened 2026-04-04
mflux-generate-z-image-turbo crashes over SSH on Mac Mini


bobosola:
I have OpenClaw on a separate dedicated PC. I am trying to use it to make images with `mflux` on my Mac over a dedicated SSH user account into a shared folder on the Mac. It seems though that the `mflux` code is doing a battery check and then crashing as described below? This is a pretty niche issue but I thought you should know about it. Here's how Kimi k2.5 reported it:

## Environment

- **Device:** Mac Mini (desktop, no battery)
- **OS:** macOS
- **Connection:** SSH session
- **Installation:** mflux installed via `uv tool install mflux`
- **Python:** 3.10

## Command

```bash
mflux-generate-z-image-turbo \
  --model Tongyi-MAI/Z-Image-Turbo \
  --steps 9 \
  --width 960 \
  --height 544 \
  --prompt "test" \
  --output test.png
```

## Error

```
FileNotFoundError: [Errno 2] No such file or directory: 'system_profiler'
```

## Full Traceback

```
Traceback (most recent call last):
  File "/Users/openclaw/.local/bin/mflux-generate-z-image-turbo", line 10, in <module>
    sys.exit(main())
  File ".../mflux/models/z_image/cli/z_image_turbo_generate.py", line 48, in main
    image = model.generate_image(...)
  File ".../mflux/models/z_image/variants/z_image.py", line 100, in generate_image
    ctx.before_loop(latents)
  File ".../mflux/callbacks/generation_context.py", line 35, in before_loop
    subscriber.call_before_loop(...)
  File ".../mflux/callbacks/instances/battery_saver.py", line 24, in call_before_loop
    current_pct = self._get_battery_percentage()
  File ".../mflux/callbacks/instances/battery_saver.py", line 32, in _get_battery_percentage
    if not self._is_machine_battery_powered():
  File ".../mflux/callbacks/instances/battery_saver.py", line 57, in _is_machine_battery_powered
    machine_model = cls._get_machine_model()
  File ".../mflux/callbacks/instances/battery_saver.py", line 65, in _get_machine_model
    result = subprocess.run(
        ["system_profiler", "SPPowerDataType", "-json"],
        ...
    )
FileNotFoundError: [Errno 2] No such file or directory: 'system_profiler'
```

## Root Cause

The `battery_saver.py` callback calls `system_profiler` to check if the machine is battery-powered. This fails because:

[...]

PR #402, opened 2026-04-06
system_profiler handle FileNotFoundError and use abspath

anthonywu:
Offering a fix for #400 

I think what happened was:

- user's ssh session does not set a `PATH` that includes `/usr/sbin` so `system_profiler` is "not found", solved by using absolute path (confirm via `which`)
- suppose the path gets deleted/moved in a future OS migration, or possibly in sandbox scenarios, just no-op via the exception handler


filipstrand:
Thanks for addressing it!

PR #405, opened 2026-04-06
Fix mflux-save dropping VisionTransformer weights for Qwen edit models

f-gibellini:
## Summary

- `mflux-save` always used `QwenImage` (txt2img) for all Qwen models, silently dropping 390 VisionTransformer weights when saving edit models
- Saved Qwen-Image-Edit models loaded with random vision encoder weights, producing degraded output
- Added edit model detection in `save.py` and `save_model()` to `QwenImageEdit`

## Details

The save CLI (`save.py`) matched all Qwen models to `QwenImage`, which initializes via `_init_models()` — no `VisionTransformer`. When `QwenImageEdit` later loaded the saved file, `strict=False` silently skipped the missing 390 `encoder.visual.*` weights, leaving the vision encoder on random initialization.

**Fix:** check for `"edit"` in the model name before the generic Qwen branch, and give `QwenImageEdit` its own `save_model()` method so the full model graph is captured.

## Reproduction

```bash
# Before fix: saved model is ~29GB (missing ~1GB of vision weights)
mflux-save --model Qwen/Qwen-Image-Edit-2511 --base-model qwen --path ./saved-q6 --quantize 6

# After fix: saved model is ~30GB (all weights included)
mflux-save --model Qwen/Qwen-Image-Edit-2511 --base-model qwen --path ./saved-q6 --quantize 6
```

## Test plan

- [ ] Save a Qwen-Image-Edit model and verify output is ~1GB larger than before
- [ ] Load saved model with `mflux-generate-qwen-edit` and confirm output matches on-the-fly quality
- [ ] Verify `mflux-save` still works for non-edit Qwen models (txt2img path unchanged)

### DIFF
--- a/src/mflux/models/common/cli/save.py
+++ b/src/mflux/models/common/cli/save.py
@@ -3,6 +3,7 @@ from mflux.models.common.config import ModelConfig
 from mflux.models.fibo.variants.txt2img.fibo import FIBO
 from mflux.models.flux.variants.txt2img.flux import Flux1
 from mflux.models.flux2.variants.txt2img.flux2_klein import Flux2Klein
+from mflux.models.qwen.variants.edit.qwen_image_edit import QwenImageEdit
 from mflux.models.qwen.variants.txt2img.qwen_image import QwenImage
 from mflux.models.z_image import ZImage, ZImageTurbo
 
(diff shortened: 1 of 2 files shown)


filipstrand:
Nice! Thanks for the fix. 

In the future I think we could re-architect the model-saving part a bit so we don't forget to support specific model variants as new ones are added. 

But this PR fix is really good to have now!

f-gibellini:
Sounds good! Trying to build a product based on the library so I might be able to review that area of the code. More PRs might be coming

PR #417, opened 2026-05-06
feat: port ERNIE-Image and ERNIE-Image-Turbo (Baidu) to mflux  

azrahello:
Transformer for text-to-image generation, and its distilled 8-step variant.

  ### What's included

  - **Inference** via two new CLI entrypoints: `mflux-generate-ernie-image-turbo` and `mflux-generate-ernie-image`
  - **LoRA inference** — supports both  ai-toolkit (`diffusion_model.layers.N.*`) and kohya(`lora_unet_layers_N_*`) key formats
  - **LoRA training** via `mflux-train` for both models
  - **`mflux-save`** support

  ### Architecture notes

  ERNIE-Image uses a Mistral-based text encoder (instead of CLIP/T5), packed latents (similar to FLUX), RoPE positional embeddings, and the standard Flow Matching linear schedule with dynamic sigma shift.

  ### Changes to shared code (minimal)

  - `model_config.py` — adds `ernie-image` and `ernie-image-turbo` to the model registry; adds `lora_training_steps` / `lora_training_guidance` fields for adapter-canonical training params
  - `flux2_vae/vae.py` — `decode_packed_latents` accepts an optional `tiling_config` parameter (backward-compatible, default `None`)
  - `linear_scheduler.py` — unchanged logic, ERNIE uses the existing `requires_sigma_shift=True` path
  - `callbacks/memory_saver.py` — guard against overwriting a model's existing `tiling_config`
  - `training/runner.py` / `trainer.py` — ERNIE training adapter wired in, same pattern as z_image / flux2
  - `training/adapters/base.py` — `generate_preview_image` no longer takes a `guidance` parameter (adapters use the value stored during `create_config`); fixes a latent bug in z_image and flux2 adapters

  ### Notes

  > ERNIE-Image was trained predominantly on Prompt Enhancer (PE)-expanded prompts. This port does not include the PE — use detailed prompts for best results.


filipstrand:
@azrahello Very nice! How does the model perform?

azrahello:
Happy to hear it! :) The model works, including for training — I'm actually running a training session right now. Unfortunately I can't share the saved models due to bandwidth limitations on my end.

I've been testing it for a few days since the PR and I have to say it doesn't look bad at all.

On M2 Ultra it runs at 6–8 s/it, with relatively low memory usage — which is actually what makes me suspect there might be room for optimization there.

P.S. For the rest, I just wanted to play around with these things — I learn more by doing than by reading books :P

filipstrand:
That sounds great, I'll take a look today. 

I apologise for my late response here, I've been a bit busy for the last couple of weeks (starting a new job), so I've had to de-prioritise MFLUX a little bit. 

@azrahello Since you've been a great help so far in the project, how do you feel about becoming a maintainer for the project going forward?

azrahello:
No worries at all, I figured you might have had things going on — and congratulations on the new job! :)

For me, contributing to this project has been a genuinely formative and fun experience — even if I cursed at the screen more than once along the way :P. I come from a photography and visual arts background, so diving into ML infrastructure and open source development has been a real stretch — but an enjoyable one. Most of what I've done has been driven by practical need: I use mflux daily for my own work, so when something was missing or broken, I just tried to fix it or add it 

As for the maintainer role — I'm genuinely flattered and definitely curious, but before committing I'd love to understand what it would concretely involve. What would you expect from a maintainer on a day-to-day basis? I'd also want to be upfront: I'm not an experienced developer, so I'd need to know that you'd still have the ability to catch and fix any mistakes I might make. As long as that safety net is there, I'm open to giving it a try.

filipstrand:
Thank you

It is really great to hear that it helps you in your own work, and I think you're probably more of a power user than me :)

As for the work, I'll try to outline the main parts:

The biggest thing: Adding new model support (along with both automatic and manual testing), as you've already done here with the help of Claude. This is great, and I am pretty confident that as coding models continue to improve that this will become easer and easer over time. There are, of course, a lot of discussions now about how code quality could degrade over time as AI writes more of it, but I think we have a pretty good code and project structure now that newer models are good at following. To keep up with new model release, I do think it is crucual to adopt AI in as many good ways as we can, especially we all are doing this in our spare time with limited resources.

Besides porting the models, there is also some extra work of maintinging the project: answering qeustions (of which you're already active), identifying and fixing bugs, doing minor releases, doing release prep (e.g making sure README:s are correct, sometimes uploading models to HF hub), and spreading the word (tweeting about it).

On top of this, there are also some amount of decision making regarding what models to support and also the interface/UI used to interact with them (e.g FIBO the special JSON-promt that requires certain type of api that Flux does not etc...). Since newer models come along all the time, it is not super obvious what to support initially - I tend to follow twitter/X/reddit to get a general feel about what models people are discussing, but also, as in this case, where people from the labs reach out, that certainly is a stronger push for us to support it since they've shown interest in the project. For this stage, it is super helpful just to discuss how new models perform and if it is worth supporting them (e.g I've yet to even play around with Earnie, for example, so here you've already done the work of actually testing the model).  

Also, now with AI, it is easier than ever to just add support for newer models, but it also feels like it comes with a certain cost of maintenance. For example, one thing I'd like to have for all active models is a automatic test suite for the very standard case we support - this involves having a big huggingface .cache folder of several 100GBs worth of model files so the test suite can be run easily ever time we make a release. With more models supported, just having enough disk space becomes a problem 😅 (of course, we could redesign how we run the tests - and we might need to - but as of right now you need the complete set of models downloaded to get a 100% score on the image tests.)

But I'm thinking we could start very light to begin with... 

For example, I'll happily continue to review new model ports like this one and I can still be in charge and responsible for merging bigger things like this. The thing I care most about here is that it works as advertised and follows the conventions. As you already know, this also involves some amout of manual testing. Since we are still basing our ports on the diffusers reference, I also like to see how we compare against the reference (e.g with a benchmark of screenshots - sometimes this can be a bit tricky since RNGs differ, but I typically copy a fixed set of initial noise latents from Diffusers and read it in MLX to get a sense)   

But for smaller things, such as minor tweaks, bug fixes you've stumbled upon etc, I would be totally fine if you merged and released minor versions (like a 0.18.x or similar). Of course, there would be no pressure doing so if you didn't feel confident. But I would give you the mandate.  

No rush in deciding right now. And if you don't feel like doing it, that is also totally fine! Just helping out with answering questions, adding new model support etc., are still very helpful.

azrahello:
Thank you for the detailed breakdown — it really helps to understand what the role would actually look like in practice.
Sorry for the slow reply on my end too, by the way!
I have to be honest: I have a lot of doubts about whether I'm the right person for this. My background is in photography and visual arts, not software development, and most of what I've contributed here came from practical necessity — things I needed for my own work. I'm not sure I have the consistency or the technical confidence to take on a formal role.
That said, if it would be useful and I can genuinely help, I'm happy to give it a try — especially if you're still around to catch the bigger mistakes. I'd rather be transparent about my limits upfront than overpromise.
No pressure either way — and thanks again for the trust. It means a lot.

filipstrand:
@azrahello Hi again, I've gone through this PR more thoroughly now and only did some minor additions and some manual testing. I get nearly perfect results compared to diffusers in several tests which really makes me trust the implementation, really great work! I tried to add some additional info in the various skills based on what I updated. 

Almost 1,5 months after your first commit, I'm happy to finally merge this PR. Next time I promise to be faster :)

PR #419, opened 2026-05-21
Enable gradient checkpointing for Flux2 training to reduce peak memory

otolab:
## Motivation

During LoRA training on Flux2 Klein (4B), `nn.value_and_grad` stores all intermediate activations across 25 transformer blocks for the backward pass. On a 64GB Mac, this causes peak memory usage of ~40GB per training step, leaving little headroom for preview generation or larger batch sizes.

## Changes

- **`Flux2Transformer`**: Add `_gradient_checkpointing` flag (default `False`) and wrap both `transformer_blocks` and `single_transformer_blocks` loops with `mlx.nn.utils.checkpoint` when enabled. This recomputes intermediate activations during the backward pass instead of storing them all.
- **`TrainingTrainer.train()`**: Enable gradient checkpointing on the transformer before training begins.
- **`TrainingTrainer._generate_previews_with_optimizer_offload()`**: Add `gc.collect()` + `mx.clear_cache()` before restoring optimizer state from disk to free preview-related memory first.

Inference is unaffected — the flag stays `False` unless the trainer explicitly enables it.

## Results (Flux2 Klein 4B, 64GB Mac)

| Metric | Before | After |
|--------|--------|-------|
| Training step peak | ~40,000 MB | ~19,000 MB |
| Training step temp allocation | ~24,500 MB | ~3,700 MB |
| With preview generation peak | OOM-prone | ~25,000 MB |
| Step time | ~2.4s | ~3.1s |

Peak memory is halved with a modest ~30% increase in step time due to recomputation.

## Test plan

- [x] Ran full training loop (2 epochs, 10 steps) without preview — peak 19GB, stable
- [x] Ran full training loop with preview generation every 5 steps — peak 25GB, stable
- [ ] Verify inference is unaffected (flag defaults to `False`)

### DIFF
--- a/src/mflux/models/common/training/trainer.py
+++ b/src/mflux/models/common/training/trainer.py
@@ -117,6 +117,10 @@ class TrainingTrainer:
         adapter.freeze_base()
         TrainingTrainer._unfreeze_lora_layers(adapter.transformer())
 
+        # Enable gradient checkpointing to reduce memory usage
+        if hasattr(adapter.transformer(), '_gradient_checkpointing'):
+            adapter.transformer()._gradient_checkpointing = True
+
         train_step_function = nn.value_and_grad(
             model=adapter.model(),
             fn=lambda b: TrainingTrainer.compute_loss(adapter, training_spec, base_config, b),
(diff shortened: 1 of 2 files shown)


otolab:
Closed for now — will test more locally before submitting upstream.

PR #422, opened 2026-05-26
Add LyCORIS LoKr support

JanGrohn:
## Summary

This PR adds LyCORIS LoKr adapter support for FLUX.1 and FLUX.2 through the existing LoRA API:

- LoKr safetensors now load via `--lora-paths` / `lora_paths`
- direct `lokr_w1` / `lokr_w2` tensors are supported
- factorized LoKr tensors are supported (`lokr_w1_a`, `lokr_w1_b`, `lokr_w2_a`, `lokr_w2_b`, optional `lokr_t2`)
- common LyCORIS / ComfyUI / SimpleTuner key layouts are supported, including `lycoris_*`, `lora_unet_*`, and `diffusion_model.*`
- optional `dora_scale` is supported
- classic LoRA and LoKr adapters can be fused on the same layer
- LoRA/LoKr adapters are baked into base weights by default after load
- `--no-bake-lora` keeps runtime adapter layers when needed

The effective PR diff is mostly:

- `src/mflux/models/common/lora/layer/linear_lokr_layer.py`
- `src/mflux/models/common/lora/mapping/lokr_factors.py`
- `src/mflux/models/common/lora/mapping/lora_loader.py`
- `src/mflux/models/common/lora/mapping/lora_saver.py`
- `src/mflux/models/flux/weights/flux_lora_mapping.py`
- `src/mflux/models/flux2/weights/flux2_lora_mapping.py`
- LoRA CLI/model init plumbing for `bake_lora`
- LoKr mapping/layer/integration tests

## Why

Community FLUX.1 and FLUX.2 adapters are often published as LyCORIS LoKr safetensors rather than classic LoRA safetensors.

Before this change, those files could be resolved through `--lora-paths`, but their LoKr tensors were not applied to the model. This meant users could not use many existing LyCORIS adapters with mflux even when the adapter targeted compatible FLUX linear layers.

This PR reuses the existing LoRA loading path and extends it to recognize LoKr tensor groups, rebuild factorized LoKr weights, map common key layouts, and apply the Kronecker product delta to the target linear layer.

Baking is enabled by default because inference does not need mutable adapter layers after load. Training and assistant adapter paths explicitly pass `bake_lora=False`.

## Validation

Automated checks:

```sh
git diff --cached --name-only | rg '\.py$' | xargs uv run ruff check
uv run pytest tests/test_lokr_layer.py tests/test_lokr_integration.py tests/test_flux2_lora_mapping.py tests/test_flux_lora_mapping.py
git diff --cached --check
```

Result:

- Ruff passed
- 44 passed
- Diff whitespace check passed

Manual FLUX.2 Klein LoKr validation using [this LoKr](https://civitai.com/models/2523546/paw-patrol-movie-stills-klein-9b):

```sh
uv run mflux-generate-flux2 \
  --model flux2-klein-9b \
  --prompt "In this still from Paw Patrol: The Mighty Movie, Rocky playing basketball on a basketball court. Rocky is wearing his green mighty pups costume." \
  --steps 4 \
  --seed 610253633428956 \
  --width 768 \
  --height 768 \
  --guidance 1.0 \
  --lora-paths "./pawPatrolKlein9b.safetensors" \
  --lora-scales 1.0 \
  --output "./mflux_klein9b_lokr.png" \
  --quantize 8
  ```
  
mflux_klein9b_lokr_fast

  Baseline with the same seed/settings and no adapter:
  ```sh
  uv run mflux-generate-flux2 \
  --model flux2-klein-9b \
  --prompt "In this still from Paw Patrol: The Mighty Movie, Rocky playing basketball on a basketball court. Rocky is wearing his green mighty pups costume." \
  --steps 4 \
  --seed 610253633428956 \
  --width 768 \
  --height 768 \
  --guidance 1.0 \
  --output "./mflux_klein9b_no_lokr.png" \
  --quantize 8
  ```
  
mflux_klein9b_no_lokr_fast

### CODE REVIEW

**Comment by plz12345** [comment]
I came here to see if there was an open feature request for this, only to find a pending PR merge implementing it.  Awesome. :)

EDIT: I tested this out, locally, on a LOKR that had previously failed via the LORA CLI argument.  Works like a charm.

**Comment by fxd0h** [comment]
Heads-up from running this on Apple Silicon (Metal): `test_lokr_matmul_matches_materialized_delta` fails here. The factorized `lokr_matmul` path is correct — it matches a NumPy Kronecker reference — but MLX's Metal matmul for the small shapes in the test accumulates at lower precision than the CPU stream: the max abs difference vs the dense `delta_weight()` path is ~2e-3, while the test asserts `rtol/atol=1e-4`. It's also unseeded, so it's nondeterministic. Seeding it and relaxing the tolerance to ~1e-2 makes it pass reliably on both backends:

```python
def test_lokr_matmul_matches_materialized_delta():
    mx.random.seed(0)
    ...
    assert mx.allclose(lokr_linear.lokr_matmul(x), mx.matmul(x, dense_delta.T), rtol=1e-2, atol=1e-2)
```

Happy to send this as a small follow-up if useful.

JanGrohn:
Thanks for this, I've adjusted the test as you suggested. Before I fixed the seed I tried with a range of random seeds and all of them passed on my M2 Max with 1e-4 but I've adjusted it to 1e-2 because that still seems reasonable.

fxd0h:
Glad it helped. The M2 Max vs M5 difference makes sense: the small-shape Metal matmul accumulates a bit differently across chips and MLX builds (it's ~2e-3 here on an M5), so a hardware-agnostic 1e-2 is the safe choice. It absorbs that variance without hiding a real error, since the factorized path still matches the NumPy Kronecker reference exactly. Thanks for the quick turnaround.

fxd0h:
Nice work on this. I have been building on top of it and it works well. One gap I hit: the LoKr patterns are wired into the flux and flux2 mappings, but the krea2, qwen and ideogram4 mappings never set `possible_lokr_w1_patterns` / `possible_lokr_w2_patterns`, so a LoKr adapter for those models matches no factors, falls back to the standard lora_A/lora_B path, and fails with "No LoRA layers were applied".

I have a small addition ready: a shared `derive_lokr_patterns()` helper that builds the lokr_w1/lokr_w2 patterns from each target's existing up/down patterns (same prefix handling), wired into the qwen and ideogram4 mappings, plus the krea2 mapping's `_target` builder. Verified end to end: a Krea 2 LoKr adapter (256 modules) goes from 0 layers applied to 256/256 keys matched, and generates.

Happy to push it as a commit onto this branch if you want it here, or as a follow-up PR once this lands. Whichever you prefer.

ghost:
@fxd0h is your krea compatible branch available for trying?

fxd0h:
Yes. It lives in mflux-CV, a community build of mflux kept rebased on upstream `main` that carries curated community PRs, including @JanGrohn's LoKr work from this one. I put it on PyPI today:

```
pip install mflux-cv
```

It imports as `mflux` and every CLI keeps its name, so it drops into an existing setup unchanged. One caveat: upstream `mflux` and `mflux-cv` provide the same `mflux` import package, so run `pip uninstall mflux` first and keep only one installed.

Beyond this PR it also carries the mapping gap I described above: a shared `derive_lokr_patterns()` helper wired into the qwen, ideogram4 and krea2 mappings, so a LoKr adapter for those models matches its factors instead of falling back to lora_A/lora_B and failing with "No LoRA layers were applied". A Krea 2 LoKr adapter matches 256/256 modules and generates.

Source: https://github.com/HowDidTheCatGetSoFat/mflux-cv

@JanGrohn the offer still stands, happy to send the mapping helper as a follow-up PR here whenever this lands.

fxd0h:
A data point from running this PR's bake path in my build since June, worth having before this merges.

`bake_lora=True` as the load default silently destroys LoRAs on fp8 base layers: `bake_and_strip_lora` folds the float delta into Fp8Linear's raw uint8 codes, where it rounds away to nothing, with no error. On Ideogram 4 it is worse than a no-op: the conditional transformer loses its LoRA in the bake while the unconditional path keeps a live one, and CFG then amplifies away from the LoRA, saturating the output into a patch mosaic at any guidance above 1. That interaction is what #439 was originally filed about, now closed with the full trail.

#440 fixes the fp8 half, folding to q8 instead of destroying. I would either land that first or make the default skip baking on fp8 bases.

The LoKr support itself has been running in my build since June without issues.

plz12345:
@fxd0h 
I've pushed four commits to this branch to unblock the merge conflict. Summary of what changed and why, since some of it goes beyond a mechanical merge — I'd like a second reviewer on the `lora_saver.py` reconciliation in particular before this goes in.

## The conflicts

**`src/mflux/models/common/cli/save.py`** — mechanical. `main` added the Krea 2 branch and duplicated the model construction into the Ideogram/else arms; this branch had refactored those arms down to computing `model_config`/`model_path` once and constructing the model a single time through `lora_init_kwargs_from_args`. I kept this branch's shape and main's Krea 2 branch.

**`src/mflux/models/common/lora/mapping/lora_saver.py`** — the real one, and neither side was takeable as-is.

`main` has since gained #440, which folds a LoRA delta over an fp8 base to MLX q8 rather than truncating it into the uint8 codes. That PR modified `_apply_lora_delta` and its callers. This branch had already deleted `_apply_lora_delta` in favour of `_bake_lora_into_linear` / `_bake_lokr_into_linear` funnelling through `_bake_delta_into_linear`.

I kept this branch's funnel and moved #440's fp8 handling into it, so LoRA **and** LoKr deltas both reach the q8 fold. Both naive resolutions are broken:

- Taking #440's side drops LoKr silently — its `_bake_fused` filters `loras` to `isinstance(lora, LoRALinear)` and folds via `lora_A`/`lora_B`.
- Taking this branch's side reintroduces #440's bug — `_dense_weight()` here handles `nn.QuantizedLinear` but not `Fp8Linear`, so it would fold a float delta into raw uint8 codes.

That second one matters more than it first appears. Before this PR, baking only happened in `ModelSaver`, so #440's bug was confined to `mflux-save` with Ideogram 4. This PR makes `bake_lora=True` the **load-time default for every model**, which would have promoted it to *every Ideogram 4 generate with a LoRA*.

The fp8/quantized decode was duplicated in three places — `LoRASaver`, `LoKrLinear._dense_base_weight`, `FusedLoRALinear._dense_base_weight` — two of which still assumed a dense or quantized base. I extracted it to `common/lora/layer/dense_weight.py` and pointed all three at it. That also fixes DoRA taking a norm over raw fp8 codes on the runtime (`--no-bake-lora`) path.

## Evidence for the LoKr/fp8 path

Synthetic `Fp8Linear` with per-row-scaled e4m3 weights (as a real checkpoint stores them), adapter attached, baked layer compared against the live wrapper. "ratio" is bake error over the adapter's own effect, so lower is better and 1.0 means the adapter was destroyed:

| case | reconciled | with the fp8 branch removed |
|---|---|---|
| LoKr over fp8 | 0.059 | 50401 |
| LoKr+DoRA over fp8 | 0.014 | 2.000 |
| fused LoRA+LoKr over fp8 | 0.024 | 14097 |
| LoKr over bf16 (control) | 0.014 | 0.014 |

Worth noting the LoKr failure mode is worse than the LoRA one #440 fixed: LoRA silently lost the delta, LoKr writes garbage into the base weight. #440's own LoRA probe still reports 0.017 here, identical to on #440 alone, so the reconciliation preserves its behaviour exactly.

## A regression this PR introduced against main

Three of main's krea2 mapping tests assert the target layer is still a `LoRALinear` after `load_and_apply_lora`. That was true when the loader never baked; with baking now on by default the wrapper is stripped and the assertion sees a plain `Linear`. main + #440 passes 8/8, this branch was 5/8.

Those tests are checking key mapping, not baking, so they now pass `bake_lora=False`. Flagging it explicitly because it's a real behaviour change this PR makes, not a stale test — anything downstream that relied on the wrapper surviving `load_and_apply_lora` needs `bake_lora=False` now.

## `mflux-save` was broken for three models

`save.py` builds every model through `lora_init_kwargs_from_args`, which now carries `bake_lora`. `ErnieImage`, `Ideogram4` and `Krea2` never accepted that keyword, so `mflux-save` raised `TypeError` for all three, and their generate CLIs advertised `--bake-lora`/`--no-bake-lora` while silently ignoring it. Ernie and Ideogram 4 were already broken on this branch before the merge; Krea 2 joined them when main landed.

Threaded `bake_lora` through `__init__` → `initializer.init` → `_apply_lora` → `load_and_apply_lora` for the three, and switched their CLIs to the shared helper. Ideogram 4 also applies LoRAs to the unconditional transformer by hand, outside `load_and_apply_lora`, so that now gets baked explicitly to keep both transformers in the same state — a path that only became safe with #440's fold underneath it.

FIBO accepts `bake_lora` and drops it, because `FIBOInitializer` doesn't apply LoRAs at all. Pre-existing on main, left alone.

## Two mapping commits

`fix: add LoKr pattern support to Ideogram4 LoRA mapping` (+48) and `fix: krea2 lokr loading` (+14) — both pure pattern data in mapping files this PR doesn't otherwise touch. Without them those two models don't recognise LoKr keys, which seemed like a gap in the feature being merged rather than new scope. Happy to drop either if you'd rather keep the diff tight.

I deliberately left out a related ComfyUI-key-format fix I'm carrying locally; it lands 94 lines in `lora_loader.py`/`lora_mapping.py` and has nothing to do with LoKr, so it'll come as its own PR.

## Verification

Everything below ran offline under `HF_HUB_OFFLINE=1`, macOS 26.5.2 / arm64, mlx 0.31.0.

CI's exact selection (`pytest -m fast`, per `.github/workflows/tests.yml`) across the whole repo:

```
723 collected / 180 deselected / 543 selected
540 passed, 3 skipped in 1.92s
```

Green, including the 39 fast tests under `tests/image_generation`. `ruff check` clean. `ruff format --check` clean except `seedvr2/README.md`, which is already unformatted on main and I left alone. I did reformat three over-length lines that `ruff format` wanted — two from this branch (`flux`/`flux2` lora mapping) and one from #440 (`ideogram4_initializer`).

Also verified by signature binding that all ten model classes `save.py` can select now accept its call.

One incidental observation: `test_lokr_matmul_matches_materialized_delta` passes here on mlx 0.31.0, but fails on a local branch of mine pinned to mlx 0.32.0 — a ~4.5e-4 relative difference against a 1e-4 tolerance. Not a problem for this PR, but that assertion looks tight enough to go red if the mlx pin moves.

## What I'd most like a second opinion on

1. The `_bake_delta_into_linear` funnel in `lora_saver.py`, specifically that folding adapters one at a time (rather than summing deltas and folding once, as #440 did) is right. I kept the chaining because a LoKr carrying a `dora_scale` is a non-linear function of the *current* base weight, so each delta has to see the previous fold's result. The cost is an extra q8 round trip per adapter on a stack, which I believe is near-idempotent for affine q8 but haven't measured across a deep stack.
2. Whether `bake_lora=True` as the load-time default is the behaviour we want, given it silently changes what `load_and_apply_lora` returns to callers.

fxd0h (approved):
Reviewed the `lora_saver.py` reconciliation and ran an independent probe against this branch (512x512 fp8 base with per-row e4m3 scales, ground truth in fp32 from the decoded base, ratio defined as yours). LoKr over fp8 comes out at 0.016, LoKr+DoRA 0.016, fused LoRA+LoKr 0.017, bf16 control 0.004. All at the q8 noise floor, and identical to what mflux-cv produces: cv has shipped this exact combination (this PR's funnel with #440's fold inside it) since June, so the reconciled shape has field time behind it.

On question 1: chaining is right, keep it. Measured on stacks of plain LoRAs over an fp8 base, chained vs summed-then-folded-once:

| stack | chained | single fold |
|---|---|---|
| 2 adapters | 0.017 | 0.012 |
| 4 | 0.019 | 0.010 |
| 8 | 0.022 | 0.009 |

The extra round trips cost about a factor of 2 at depth 8 and stay at the noise floor, so near-idempotent holds in practice. The case that decides it: a DoRA adapter later in the stack. Its delta computed against the original base and folded in one shot lands at 0.16, ten times the chained 0.017, because `dora_scale` renormalizes rows of the base as it stands at that point. Summing only works when nothing in the stack is base-coupled; chaining is correct for both, and it keeps one code path.

On question 2: keep `bake_lora=True`. cv has had it as the load default since May, training paths pass `bake_lora=False` explicitly and everything else wants the fold. It deserves a line in the release notes, since callers relying on the wrapper surviving `load_and_apply_lora` do see a behavior change, exactly as your krea2 test fix shows.

The `dense_weight.py` extraction fixes a real gap, and I can confirm it from the other direction: cv's bare LoKr layer is missing the same fp8 branch on its runtime path (the fused wrapper there has it), so DoRA over fp8 with `--no-bake-lora` norms over raw codes. Porting your extraction to cv.

<details>
<summary>probe script</summary>

```python
import mlx.core as mx
from mlx import nn

from mflux.models.common.lora.layer.linear_lokr_layer import LoKrLinear
from mflux.models.common.lora.layer.linear_lora_layer import LoRALinear
from mflux.models.common.lora.mapping.lora_saver import LoRASaver
from mflux.models.ideogram4.model.ideogram4_transformer.fp8_linear import Fp8Linear

mx.random.seed(1234)
OUT, IN = 512, 512
FP8_MAX = 448.0


def make_fp8_base():
    dense = mx.random.normal((OUT, IN)).astype(mx.float32) * 0.02
    scale = mx.max(mx.abs(dense), axis=1) / FP8_MAX
    codes = mx.to_fp8((dense / scale[:, None]).astype(mx.float32))
    base = Fp8Linear(IN, OUT, bias=True)
    base.weight = codes
    base.weight_scale = scale.astype(mx.float32)
    base.bias = mx.random.normal((OUT,)).astype(mx.bfloat16) * 0.01
    w_dense = mx.from_fp8(codes, dtype=mx.float32) * scale[:, None]
    return base, w_dense


def make_bf16_base():
    w = (mx.random.normal((OUT, IN)).astype(mx.float32) * 0.02).astype(mx.bfloat16)
    base = nn.Linear(IN, OUT, bias=True)
    base.weight = w
    base.bias = mx.random.normal((OUT,)).astype(mx.bfloat16) * 0.01
    return base, w.astype(mx.float32)


def make_lokr(seed):
    mx.random.seed(seed)
    w1 = mx.random.normal((16, 16)).astype(mx.float32) * 0.1
    w2 = mx.random.normal((32, 32)).astype(mx.float32) * 0.1
    return w1, w2


def lokr_on(base, seed, dora=False, w_dense=None):
    w1, w2 = make_lokr(seed)
    dora_scale = None
    if dora:
        delta = mx.kron(w1, w2).reshape((OUT, IN))
        dora_scale = mx.linalg.norm(w_dense + delta, axis=1) * (1.0 + 0.05 * mx.random.normal((OUT,)))
    return LoKrLinear(
        linear=base, output_dims=OUT, input_dims=IN, lokr_w1=w1, lokr_w2=w2, dora_scale=dora_scale, scale=1.0
    )


def clone_fp8(base):
    b = Fp8Linear(IN, OUT, bias=True)
    b.weight = base.weight
    b.weight_scale = base.weight_scale
    b.bias = base.bias
    return b


def forward(layer, x):
    return layer(x).astype(mx.float32)


def report(name, y_true, y_base, y_baked):
    effect = float(mx.linalg.norm(y_true - y_base))
    err = float(mx.linalg.norm(y_baked - y_true))
    print(f"{name:44s} effect={effect:10.4f}  bake_err={err:10.4f}  ratio={err / effect:.4f}")


x = mx.random.normal((64, IN)).astype(mx.float32)

# A) LoKr over fp8
base, w_dense = make_fp8_base()
lokr = lokr_on(base, seed=7)
delta_true = lokr.delta_weight(base_weight=w_dense).astype(mx.float32)
y_base = x @ w_dense.T + base.bias.astype(mx.float32)
y_true = x @ (w_dense + delta_true).T + base.bias.astype(mx.float32)
baked = LoRASaver._bake_lokr_into_linear(base, lokr)
report("LoKr/fp8", y_true, y_base, forward(baked, x))

# B) LoKr+DoRA over fp8
base, w_dense = make_fp8_base()
lokr_d = lokr_on(base, seed=11, dora=True, w_dense=w_dense)
delta_true = lokr_d.delta_weight(base_weight=w_dense).astype(mx.float32)
y_base = x @ w_dense.T + base.bias.astype(mx.float32)
y_true = x @ (w_dense + delta_true).T + base.bias.astype(mx.float32)
baked = LoRASaver._bake_lokr_into_linear(base, lokr_d)
report("LoKr+DoRA/fp8", y_true, y_base, forward(baked, x))

# C) fused LoRA+LoKr over fp8, chained like _bake_fused
base, w_dense = make_fp8_base()
lora = LoRALinear.from_linear(base, r=16, scale=1.0)
mx.random.seed(21)
lora.lora_A = mx.random.normal(lora.lora_A.shape).astype(mx.float32) * 0.05
lora.lora_B = mx.random.normal(lora.lora_B.shape).astype(mx.float32) * 0.05
lokr2 = lokr_on(base, seed=23)
d_lora = (lora.scale * mx.transpose(mx.matmul(lora.lora_A, lora.lora_B))).astype(mx.float32)
d_lokr = lokr2.delta_weight(base_weight=None).astype(mx.float32)
y_base = x @ w_dense.T + base.bias.astype(mx.float32)
y_true = x @ (w_dense + d_lora + d_lokr).T + base.bias.astype(mx.float32)
step1 = LoRASaver._bake_lora_into_linear(base, lora)
step2 = LoRASaver._bake_lokr_into_linear(step1, lokr2)
report("fused LoRA+LoKr/fp8 (chained)", y_true, y_base, forward(step2, x))

# D) LoKr over bf16 control
base16, w16 = make_bf16_base()
lokr3 = lokr_on(base16, seed=31)
delta_true = lokr3.delta_weight(base_weight=w16).astype(mx.float32)
y_base = x @ w16.T + base16.bias.astype(mx.float32)
y_true = x @ (w16 + delta_true).T + base16.bias.astype(mx.float32)
baked = LoRASaver._bake_lokr_into_linear(base16, lokr3)
report("LoKr/bf16", y_true, y_base, forward(baked, x))

# Q1: chained fold vs summed single fold, N plain LoRAs over fp8
for n in (2, 4, 8):
    base, w_dense = make_fp8_base()
    mx.random.seed(100 + n)
    loras = []
    for i in range(n):
        lo = LoRALinear.from_linear(base, r=16, scale=1.0)
        lo.lora_A = mx.random.normal(lo.lora_A.shape).astype(mx.float32) * 0.05
        lo.lora_B = mx.random.normal(lo.lora_B.shape).astype(mx.float32) * 0.05
        loras.append(lo)
    deltas = [(lo.scale * mx.transpose(mx.matmul(lo.lora_A, lo.lora_B))).astype(mx.float32) for lo in loras]
    total = sum(deltas)
    y_base = x @ w_dense.T + base.bias.astype(mx.float32)
    y_true = x @ (w_dense + total).T + base.bias.astype(mx.float32)

    cur = base
    for lo in loras:
        cur = LoRASaver._bake_lora_into_linear(cur, lo)
    report(f"N={n} chained ({n} q8 round trips)", y_true, y_base, forward(cur, x))

    single = LoRASaver._bake_delta_into_linear(clone_fp8(base), total)
    report(f"N={n} summed  (1 q8 round trip)", y_true, y_base, forward(single, x))

# Q1b: DoRA in a stack, chained vs delta-against-original-base
base, w_dense = make_fp8_base()
mx.random.seed(300)
first = LoRALinear.from_linear(base, r=16, scale=1.0)
first.lora_A = mx.random.normal(first.lora_A.shape).astype(mx.float32) * 0.05
first.lora_B = mx.random.normal(first.lora_B.shape).astype(mx.float32) * 0.05
d_first = (first.scale * mx.transpose(mx.matmul(first.lora_A, first.lora_B))).astype(mx.float32)
w_after_first = w_dense + d_first
lokr_d = lokr_on(base, seed=41, dora=True, w_dense=w_after_first)
d_dora_true = lokr_d.delta_weight(base_weight=w_after_first).astype(mx.float32)
y_base = x @ w_dense.T + base.bias.astype(mx.float32)
y_true = x @ (w_after_first + d_dora_true).T + base.bias.astype(mx.float32)

cur = LoRASaver._bake_lora_into_linear(base, first)
cur = LoRASaver._bake_lokr_into_linear(cur, lokr_d)
report("chained (DoRA sees folded base)", y_true, y_base, forward(cur, x))

d_dora_naive = lokr_d.delta_weight(base_weight=w_dense).astype(mx.float32)
naive = LoRASaver._bake_delta_into_linear(clone_fp8(base), d_first + d_dora_naive)
report("summed-naive (DoRA sees original base)", y_true, y_base, forward(naive, x))
```

</details>

plz12345:
OK, with that comment validating, and the clean CI run, I'll merge this one in.  Early git clone testers have a shot at it, until the next release is cut.

PR #426, opened 2026-05-29
Add FLUX.2-klein-9b-kv KV-cache support (~2.4× speedup on multi-ref edit)

michaeltrefry:
This PR ports BFL's `Flux2KleinKVPipeline` KV-cache optimisation (now in upstream diffusers) to the mflux Flux2 transformer stack so the [FLUX.2-klein-9b-kv](https://huggingface.co/black-forest-labs/FLUX.2-klein-9b-kv) distilled variant runs on Apple Silicon with the intended cache acceleration.

## What it does

`FLUX.2-klein-9b-kv` is the same architecture as `flux2-klein-9b` (8 double + 24 single layers, 32 heads, dim 12288) with weights step-distilled at 4 inference steps to support a runtime KV-cache: reference-image (and text) K/V tensors are computed once on step 0 and re-used on steps 1-3, skipping redundant reference-token processing on every step.

## Validation

Tested on M5 Max 128 GB, 1024×1024, 4 steps, single reference image:

| Variant | Cache | Total | Per-step | Speedup |
|---|---|---|---|---|
| `flux2-klein-9b` (base) | n/a | 33.0 s | ~8.5 s | 1.0× |
| `flux2-klein-9b-kv` | **on** | **13.5 s** | **~3.4 s avg** | **2.4×** |

Step 0 (extract) ~5.5 s; steps 1-3 (cached) ~3.0 s each — clean cache effect. Output coherent, identity preserved (cat-in-wizard-hat edit, both runs).

Matches BFL's [README](https://huggingface.co/black-forest-labs/FLUX.2-klein-9b-kv) claim of \"up to 2.5×\".

## Changes

- **`Flux2KVCache`** (new): per-layer K/V container with double-stream / single-stream slots, mode (`extract`/`cached`), and ref/txt token counts.
- **`Flux2Attention`** / **`Flux2ParallelSelfAttention`**: optional `kv_cache` + `kv_cache_layer_idx` kwargs. Extract mode stores post-RoPE K/V for the ref slice; cached mode splices stored K/V at the end of the fresh K/V for the `[txt, target]` input.
- **`Flux2TransformerBlock`** / **`Flux2SingleTransformerBlock`** / **`Flux2Transformer`**: thread cache kwargs through; transformer dispatches each block with its layer index.
- **`Flux2KleinEdit.generate_image`**: auto-engages the cache when the model declares `supports_kv_cache=True` and a reference image is present. Step 0 runs the full `[txt, target, ref]` forward in extract mode; subsequent steps pass target-only `hidden_states` in cached mode. The full predict closure skips `mx.compile` for KV-cache models because the cache state is mutable Python state that compile would freeze at trace time.
- **`ModelConfig`**: adds `supports_kv_cache: bool` field and a `flux2-klein-9b-kv` entry pointing at the BFL HF repo with the same transformer overrides as `flux2-klein-9b`.
- **`MODEL_CHOICES`** (defaults): register `flux2-klein-9b-kv` so the CLI routes the alias to ModelConfig resolution instead of treating it as a model path.

## Implementation notes

mflux's post-concat token layout is `[txt, target, ref]` (because `Flux2KleinEdit._predict` concatenates `[latents, image_latents]` and `Flux2Attention` then prepends the text stream), whereas the diffusers reference uses `[txt, ref, target]`. The cache protocol is identical; this implementation slices the trailing `num_ref_tokens` tokens in extract mode and concatenates cached ref K/V at the end of the fresh K/V in cached mode. No reordering of the base path was needed, so `flux2-klein-9b` (and all other Flux2 variants) are byte-for-byte unaffected when the cache is disabled.

The cache is opt-out via `generate_image(use_kv_cache=False)` for debugging or A/B comparison against a no-cache run on the same -kv weights.

## Out of scope

- CLI flag to toggle cache from `mflux-generate-flux2-edit` (currently controlled by ModelConfig). Happy to add `--no-kv-cache` if you'd like — kept minimal here.
- Multi-reference edit timing measurements. Single-ref already shows the expected speedup; multi-ref should scale proportionally to ref token count but I haven't benched it here.


filipstrand:
@michaeltrefry Thanks for adding the kv support! I've tried this on my M3 Ultra and get 1.5x - 1.9x speedup, really good! I'm debating whether or not to make this the default option actually for the 9b edit mode (and more opt-out if you don't want to use use, like a `--no-cache` version or similar). I'm running some benchmarks over images to see how this model performs compared to the standard one.

filipstrand:
After running some benchmarks: The performance is really good, so could probably default to this model for the edit 9b case, but thinking some more about it, it is probably more honest and straight forward to simply have as it is now 

`--model flux2-klein-9b-kv`

PR #427, opened 2026-05-29
Fix FLUX.2 Klein crash in mflux-generate (FileNotFoundError: text_encoder_2)

c2p-cmd:
## Problem

Running `mflux-generate --base-model flux2-klein-4b` crashed with:

```bash
FileNotFoundError: No safetensors files found in .../text_encoder_2
```

FLUX.2 Klein uses a single Qwen3 text encoder (`text_encoder/`) — there is no `text_encoder_2/` directory. But `FluxWeightDefinition`  unconditionally included a `t5_encoder` component pointing at `text_encoder_2/`, and `mflux generate` routed all model names (including Klein) through `Flux1`.

## Changes

- **`ModelConfig.is_klein()`** — new property to identify FLUX.2 Klein models by alias, reusable by any caller
- **`FluxWeightDefinition`** — `get_components()`, `get_tokenizers()`, `get_download_patterns()` now accept an optional `model_config` and skip `t5_encoder` / `text_encoder_2` for Klein models; fully backward-compatible (`None` default keeps existing T5 behaviour for all other models)
- **`flux_generate.py`** — `mflux-generate` now resolves `ModelConfig` early and dispatches Klein models to `Flux2Klein` via `_run_klein()`, matching what `mflux-generate-flux2` already does
- **`KLEIN_MODELS`** — renamed from `__KLEIN_MODELS` (now public)

## Verification

- 418 fast tests pass (`make test-fast`)
- `make lint` — no new errors in changed files
- Non-Klein models (dev, schnell, krea-dev, z-image, etc.) are completely unaffected

### DIFF
--- a/src/mflux/models/common/config/model_config.py
+++ b/src/mflux/models/common/config/model_config.py
@@ -202,6 +202,10 @@ class ModelConfig:
     def is_canny(self) -> bool:
         return self.controlnet_model is not None and "Canny" in self.controlnet_model
 
+    def is_klein(self) -> bool:
+        _KLEIN_ALIASES = {"flux2-klein-4b", "flux2-klein-9b", "flux2-klein-base-4b", "flux2-klein-base-9b"}
+        return bool(set(self.aliases) & _KLEIN_ALIASES)
+
     @staticmethod
     def from_name(
         model_name: str,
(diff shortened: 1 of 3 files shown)


filipstrand:
Ah, yes I guess this is a bit unfortunate, the idea is to use `mflux-generate-flux2` as described here: https://github.com/filipstrand/mflux/blob/main/src/mflux/models/flux2/README.md. 

It is a bit confusing now that we also have the older mflux-generate while still supporting Flux1. The newer CLI that is in the works will make this more clear https://github.com/filipstrand/mflux/issues/357

To keep things simple in the mean time for this very legacy case, I'll just add a nicer note to the user, something like 
`FLUX.2 Klein is not supported by mflux-generate. Use mflux-generate-flux2 instead.`

PR #428, opened 2026-05-29
Add mlx-taef and mlx-teacache to Related projects

IonDen:
Hi — I've built two small MLX libraries that plug into mflux and I'd like to add them to the Related projects list:

- mlx-taef — TAESD/TAEF tiny autoencoders for live previews and low-memory FLUX decode. It hooks into mflux's stepwise callback, so you get a preview image each step without paying for the full VAE.
- mlx-teacache — a TeaCache step-skipping wrapper that speeds up FLUX on Apple Silicon by skipping denoising steps when the schedule is actually cacheable.

Both install from PyPI and keep their benchmarks and caveats in their own READMEs. I've tried to keep the caveats honest: teacache doesn't skip steps on distilled 4–8 step schedules, and taef previews are approximate, not full-VAE quality.

This only touches the Related projects list — two bullets, no code.

Happy to reword or drop either if you'd rather keep the list tight. Either way — thanks for mflux. Both libraries are built on top of it.


filipstrand:
Hi @IonDen , thanks for sharing your projects, I'll happily include them. I remember looking into teacache last year or so, but ended up not merging it at the time. I'll have a look at your projects.

IonDen:
Thank you @filipstrand !

PR #430, opened 2026-06-01
fix: always evict text encoder and clear MLX cache between seeds

plz12345:
In testing mflux while using the option to pass in multiple seeds (batch jobs), I found a couple places where memory management is risky or outright dangerous.  This fixes two of those situations.

- Frees text encoder from unified memory after encoding — it's never needed during the denoising loop, saving 8–12 GB throughout generation.  This was a bigger issue when testing on Flux.2 Klein 9B, specifically, since the official version and even the mlx-community quants don't quantiza the Qwen encoder.
- Clears MLX cache between seeds when keep_transformer=True to prevent buffer accumulation and OOM on multi-seed generation runs.  This presented only when doing mult-seed generation, where each one left RAM held until the end of the whole batch, leading to a game of OOM Russian Roulette depending on batch size.  On my 64GB MBP with Flux.2 Klein 9b q8, it only took 5 seed sin a batch to completely RAM-lock and require a hard reboot.

### DIFF
--- a/src/mflux/callbacks/callback_manager.py
+++ b/src/mflux/callbacks/callback_manager.py
@@ -63,25 +63,30 @@ class CallbackManager:
 
     @staticmethod
     def _register_memory_saver(args: Namespace, model) -> MemorySaver | None:
-        memory_saver = None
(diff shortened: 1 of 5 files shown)


filipstrand:
Thanks for proposing this fix @plz12345

PR #434, opened 2026-06-07
Add: A validation Github workflow for mflux pull requests.

ianscrivener:
Filip, reading your note today I sensed that you may be getting messy PRs that hadn't done the basic linting & tests per your dev guildelines. This sounded like somewhere I could add some value...

So for each PR, this GH Action workflow:

1. Runs `ruff`, `typos`, and `mypy against the main branch first to establish a baseline — so contributors are not penalised for pre-existing issues
2. Runs the same tools against the PR branch
3. Runs `make test-fast` (pytest fast suite, no image generation)
4. Runs `make check` (the full pre-commit git hook suite)
5. Posts a formatted markdown report as a PR comment
6. Uploads the report as a GitHub Actions artifact

Hoping this may make your great mflux work a little easier...

 Thanks for your great work Filip!


.github/workflows/mflux-pr-report-card.yml
```diff
@@ -0,0 +1,288 @@
+name: PR Report Card
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
filipstrand: Thanks for this suggestion, I have not used this earlier but it feels like a nice addition. I guess the pre-commit hooks would prevent things like this, but they can be circumvented. Sometimes not installed etc.

I did a quick review with Claude which suggested that including `synchronize` here would mean that every time someone pushed to their PR branch, a new comment with a PR report would always get added. Is this true? If so, I guess there is a risk of polluting the PR?

ianscrivener: Filip, 
The goal was to do a single autromated CI check on a PR when submitted to the **main mflux repo**... so you would get a signal as to the quality of the PR. If a contributor saw a report that was less than perfect there was some chance that they'd go back and fix the issues. So in the end you'd end up with cleaner PRs... saving you time.  

So with that goal in mind- this GitHub action runs on the **main mflux repo only** to check the pushed PR. (It will need a tiny bit of tweaking and testing to ensure that it only triggers at the correct time if you put it into the Mflux repo.)

The Claude `synchronize` suggestion would be valid if the intention is that contributors use GitHub for their **own mflux forks** and they run this GitHub action check every time they commit. To be honest, I didn't think of that... and I see some potential issues with that. That opens up the chance of contributors creating issues for the GH Action itself - which is an _anti-goal_. 

My test version ran as a manually triggered GitHub action running on my own mflux repo. I check the results in the action logs or download the full Markdown report.  

Perhaps I can suggest;
**Stage 1**: run `mflux-pr-report-card.yml` automatically **on mflux main repo only**   
**Stage 2**: run `mflux-pr-report-card.yml` optionally **manually** on contributor Github repos. (I'll can do doco for that).

ianscrivener (commented):
typos threw a false alarm on `nl -ba /tmp/pr-typos.txt | sed 's/^/  /'`
Have disabled typos for that line only

Issue #357, opened 2026-02-16
Design review and discussion: MFLUX v1.0 CLI


filipstrand:
As the project has evolved and now support a whole family of models and techniques, the current CLI starts to feel outdated and unnecessarily complex. This is a place to discuss how a new and modern interface could/should look like given what we know about model families, variants and features in 2026 and beyond. 

- Which top level commands should we support (like `mflux generate`, `mflux edit`, `mflux save` etc?)
- What are good names (e.g --image or --images might be preferred to "--image_paths")?
- How should model families be specified? 
- How should model variants be specified (4b, 7b)? 
- How should LoRAs be specified (some cleanups and redesign have already been made here, e.g how we fetch from HF where the repo might include several .safetensors files etc). 
- How should resolution be specified or inferred (via input image, via aspect ratio, default values etc?) 
- How should low-ram optimisations be specified (e.g, what should be hidden behind a single --low-ram vs treated as sepearate, e.g see. https://github.com/filipstrand/mflux/pull/351) 
- what is natural to have to specify given that we already have an input image in some case (e.g default resolution could be inferred if not specified)
- anticipate/consider what could be coming in the future (e.g more optimisations, other model types etc). 
- how do we document this for humans?
- how do we document this for agents?
- many more things...

I invite everyone to participate in this open discussion, there are no right or wrong answers :) 

I think the more we can share how we use the project currently, and how we wish things to be, the better shot we have at designing a stable and good CLI for the future.  

Once we agree on a design, I think the actual implementation should be fairly trivial with the help of coding agents. 

I'll update this doc with some of my own suggestions after finishing current ongoing work with model porting...

azrahello:
Hi again,
following up on our previous exchange — wanted to bring up a few features that I think could make a real difference in day-to-day mflux workflows.
1. API server support
With so many inference servers emerging (oMLX, MLX-LM server, and others), having a proper API layer in mflux could be a real game changer — not just for scripting, but for creative integrations: chaining LLM/VLM and T2I models together in pipelines, connecting to node-based UIs, or using tools like Claude Code as a coding agent against a local backend. An OpenAI-compatible endpoint would make mflux a first-class citizen in this growing ecosystem.
2. Shared checkpoints / component reuse
[...]

PR #438, opened 2026-06-10
Add atomic --lora and --image flags

anthonywu:
AI disclosure: assisted by Opus 4.8

-----

Implements the two highest-signal, lowest-risk proposals from #357: atomic flags that pair each path with its value on a single, repeatable flag, instead of two positionally-aligned lists.

```sh
mflux-generate --prompt "cyberpunk mermaid" \
  --lora A.safetensors 0.7 \
  --lora B.safetensors \          # scale omitted -> 1.0
  --image photo.jpg 0.6           # strength omitted -> model default
```

This removes the alignment footgun of `--lora-paths`/`--lora-scales` and `--image-path`/`--image-strength` when working with many adapters — commenting out one LoRA is now one line, with no recount of a parallel scale list.

### Design — zero downstream churn

The new flags fold into the existing `lora_paths`/`lora_scales` and `image_path`/`image_strength` namespace fields in `parse_args`, **before** the metadata-merge and path-resolution logic. So every downstream consumer (metadata reuse via `--config-from-metadata`, library/HF path resolution, model init) is untouched. Downstream code reads the explicit fields, not `vars(namespace)`, so the raw `--lora`/`--image` attrs don't leak into saved metadata.

### Backward compatibility — fully opt-in

- Legacy `--lora-paths`, `--lora-scales`, `--image-path`, `--image-strength` remain registered with identical types/defaults and behavior. Existing scripts and saved metadata work unchanged.
- **No runtime deprecation warnings** — mflux is pre-1.0, which the #357 migration plan designates as silent coexistence (warnings land at v1.0). The only signal is a passive `[DEPRECATED: use --lora]` tag in `--help`.
- No existing invocation can hit a new error path: the new+legacy conflict guard only fires when both are combined, and no current user types the new flags.

### Validation (teaching errors)

Mixing a new flag with its legacy counterpart, a non-numeric scale/strength, or too many values in one group each produce a teaching error rather than silently misparsing — this closes the argparse `nargs="+"` greedy-consumption edge (e.g. `--lora A B 0.7`).

### Note for reviewers — one deliberate behavior change

For kontext (`required=True`), the init-image requirement moved from argparse-time to post-normalization so **either** `--image` or `--image-path` satisfies it. A side effect: since kontext uses `supports_metadata_config=True`, an `image_path` supplied via `--config-from-metadata` now satisfies the requirement where it previously errored without a CLI image. This is strictly more permissive and arguably a fix, but flagging it as a decision rather than an accident. The error message also changed to a teaching message.

### Tests

- 14 new arg-parser tests: scale/strength defaulting, metadata merge, new+legacy conflict, non-numeric value, too-many-values, and the kontext required-image case satisfied by either flag.
- Completion-generator assertion locking the #357 "CLI completion must stay compatible" requirement (matches the exact option-spec token so it can't pass off the legacy `--lora-paths`/`--image-path` substrings).
- Legacy behavior stays regression-locked by the unchanged `test_lora_args` / `test_image_to_image_args`.
- Full fast suite: 489 passed, 3 skipped. New normalization code is 100% line-covered.

Refs #357.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### ✨ Improvements
+
+- **Atomic `--lora` and `--image` flags**: Pair each path with its value on a single, repeatable flag — `--lora A.safetensors 0.7 --lora B.safetensors` (scale defaults to `1.0`) and `--image photo.jpg 0.6` (strength defaults to the model default). This removes the positional-alignment footgun of the parallel `--lora-paths`/`--lora-scales` and `--image-path`/`--image-strength` lists, which remain fully supported and are marked deprecated in `--help`. (#357)
+
 ## [0.18.0] - 2026-06-07
 
 ### 🎨 New Model Support
--- a/src/mflux/cli/parser/parsers.py
+++ b/src/mflux/cli/parser/parsers.py
@@ -260,7 +260,7 @@ class CommandLineParser(argparse.ArgumentParser):
                 except ValueError:
                     self.error(f"Invalid LoRA scale '{group[1]}' for '{group[0]}'.\nTip: --lora takes a PATH and an optional numeric SCALE: --lora {group[0]} 0.7")  # fmt: off
             else:
-                self.error(f"--lora takes one PATH and an optional SCALE but got {len(group)} values: {' '.join(group)}\nTip: repeat --lora for each adapter: --lora {group[0]} {group[1]} --lora {group[2]} ...")  # fmt: off
+                self.error(f"--lora takes one PATH and an optional SCALE but got {len(group)} values: {' '.join(group)}\nTip: give each adapter its own --lora: --lora A.safetensors 0.7 --lora B.safetensors")  # fmt: off
             paths.append(path)
             scales.append(scale)
         namespace.lora_paths = paths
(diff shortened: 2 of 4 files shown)


anthonywu:
Test result:

- `uv run --extra dev pytest -m fast`
- `490 passed, 3 skipped, 82 deselected, 1 warning in 2.34s`

anthonywu:
### Manual validation

Tested on Apple Silicon against the locally-cached 4-bit model `filipstrand/FLUX.1-Krea-dev-mflux-4bit` (no downloads triggered).

**End-to-end `--image` (image-to-image):**

```sh
uv run mflux-generate \
  --model filipstrand/FLUX.1-Krea-dev-mflux-4bit --base-model krea-dev \
  --prompt "a serene alpine lake at golden hour, mist on the water" \
  --image tests/resources/reference_dev_image_to_image.png 0.6 \
  --steps 12 --height 768 --width 768 --seed 42 \
  --output /tmp/mflux_image_flag.png
```

✅ Generates successfully — `--image PATH STRENGTH` correctly resolves to `image_path` + `image_strength=0.6` and reaches the pipeline. Omitting the strength (`--image …png`) falls back to the `0.4` default.

**Validation / teaching errors (all exit during `parse_args`, before any model load):**

- ✅ `--image a.jpg --image-strength 0.6` → rejects mixing new + legacy flags
- ✅ `--image a.jpg 0.6 extra` → rejects >2 values in one group
- ✅ `--lora a.safetensors b.safetensors 0.7` → rejects >2 values in one group
- ✅ `--lora a.safetensors notanumber` → rejects non-numeric scale

Follow-ups from validation:
- Improved the `--lora` too-many-values tip, which previously echoed the user's tokens back and rendered a nonsensical example.
- Cosmetic note (not addressed here): argparse renders the usage as `--lora PATH [SCALE ...]` / `--image PATH [STRENGTH ...]`; the trailing `...` over-implies 3+ values are accepted, but the normalizer correctly rejects them. Tidying this would need a custom action; left out to keep the diff focused.

filipstrand:
@anthonywu Thank you for getting started on this. I've wanted to do the same for a while now but have been putting it off due to lack of time and energy... (originally I planned to do the whole thing in one go but this is probably a better approach). I have not tried this yet, but feel confident in your reported results!

PR #439, opened 2026-06-10
Fix Ideogram 4 LoRA generation: corrupted output at any guidance > 1

fxd0h:
## Problem
Every LoRA applied to Ideogram 4 renders as a saturated patch-mosaic at any guidance > 1, regardless of how the LoRA was trained. A reference character LoRA that produces the trained subject through its own trainer's sampler renders unrecognizable garbage through mflux.

## Root cause
Ideogram 4 ships a separate **distilled unconditional transformer** for CFG. `_apply_lora` applied the LoRA to **both** transformers — but a LoRA is trained against the *conditional* transformer's weights. Applied to the unconditional model (different weights), it pushes that model off-distribution, and CFG then amplifies the corrupted negative by the guidance factor:

`v = g*pos + (1-g)*neg`

so the corruption scales with `g` (4–7x with the shipped presets).

## Fix
- Apply the LoRA to the conditional transformer only.
- When a LoRA is active, route the CFG negative through the (LoRA'd) conditional transformer with the existing negative inputs. The LoRA's unconditional shift then largely cancels in `(pos - neg)` instead of being amplified.
- Without a LoRA, behavior is unchanged (the distilled unconditional transformer is used exactly as before).

## Validation
- Known-good character LoRA, guidance 4.0, previously mosaic → now clean and matches the reference sampler's output (same subject, same quality).
- Control: the same LoRA at guidance 1.0 (CFG disabled) was already clean before the fix, isolating CFG as the corruption path.
- Full fast test suite passes (516 passed).

### DIFF
--- a/src/mflux/models/ideogram4/ideogram4_initializer.py
+++ b/src/mflux/models/ideogram4/ideogram4_initializer.py
@@ -101,6 +101,11 @@ class Ideogram4Initializer:
 
     @staticmethod
     def _apply_lora(model, lora_paths: list[str] | None, lora_scales: list[float] | None) -> None:
+        # Apply ONLY to the conditional transformer. The unconditional transformer is a
+        # DIFFERENT (distilled) model: a LoRA trained against the conditional's weights pushes
+        # it off-distribution, and CFG then amplifies that corrupted negative by the guidance
+        # factor (output degrades into a patch mosaic at any guidance > 1). With a LoRA active,
+        # generation routes the CFG negative through the (LoRA'd) conditional transformer.
         lora_mapping = Ideogram4LoRAMapping.get_mapping()
         model.lora_paths, model.lora_scales = LoRALoader.load_and_apply_lora(
             lora_mapping=lora_mapping,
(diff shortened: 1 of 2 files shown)


anthonywu (approved):
nominating merge by @filipstrand 

I haven't downloaded this model, but it seems @fxd0h can take the lead here to validate it works as advertised

fxd0h:
Closing this one, with the test output @anthonywu asked for, because validating it properly showed the premise is wrong.

The suites, on this PR merged into current main (MacBook Pro, M5 Max 40-core GPU, macOS 26.5, Python 3.12):

```
mlx 0.31.0: 510 passed / mlx 0.32.0: 510 passed
```

Then I tried to reproduce the advertised bug, and could not. On today's main and on this PR's own base commit, with mlx 0.31.2, the same version that was current when this was filed, a character LoRA at scale 1.0 and preset guidance up to 7 renders clean with strong subject likeness, pixel-identical between the two commits. A style LoRA is clean too.

The mosaic I filed this against was real, but it came from my own build, not from this repo. My branch carried #422, whose load path bakes LoRAs into the base by default. On Ideogram's fp8 weights that bake rounds the delta into the raw uint8 codes, a silent no-op, and it only hit the conditional transformer, while the unconditional got its LoRA through a path that skips the bake. The LoRA ended up only on the negative side of CFG, and guidance amplified flight from it. This repo never had load-time baking, so it never had the bug.

Worse, the routing half of this PR harms current main: sending the CFG negative through the LoRA'd conditional cancels most of the LoRA's effect and the subject washes out. Verified side by side.

What remains true and already filed elsewhere: baking a LoRA into an fp8 base silently destroys it, which is #440, and that is the thing to guard before anything like bake-by-default lands.

Thanks for asking for the output. Running it is how this surfaced.

PR #440, opened 2026-06-10
Fix LoRA baking on fp8 base layers: fold to q8 instead of silently destroying the delta

fxd0h:
## Problem
`LoRASaver.bake_and_strip_lora` folds the delta with `weight + delta.astype(weight.dtype)`. On fp8 layers (Ideogram 4's `Fp8Linear`) `.weight` holds **raw uint8 codes**, so the small float delta truncates to ~zero: the wrapper is stripped and the LoRA silently vanishes. This bites anything that bakes over an fp8 base — e.g. saving an Ideogram 4 model with LoRAs applied via `ModelSaver`.

## Fix
Detect fp8 bases (uint8 codes + `weight_scale`), dequantize once, fold the delta(s) in float32, and requantize to MLX **q8** (`group_size=64`). q8 affine keeps more mantissa than fp8-e4m3, so nothing is lost relative to the base.

## Validation
- Real Ideogram 4 qkv layer + nontrivial LoRA: baked-q8 output vs live wrapper differs only by quantization noise (max rel ~1e-2).
- Bonus: the folded layer runs on MLX's fused quantized-matmul kernel instead of materializing the full bf16 weight matrix per forward — measured 4.40 → 2.84 ms on that layer at sequence length 1034.
- Non-fp8 layers take exactly the previous path; full fast test suite passes (516 passed).

## Update (2026-07-30): the load half

Validating end to end showed the fold alone was not enough: the saved checkpoint holds packed q8 tensors (`weight`/`scales`/`biases`) where `Fp8Linear` expects uint8 codes plus `weight_scale`, `update(strict=False)` skips them silently, and the first forward raises. A second commit rebuilds any such layer as `QuantizedLinear` before weights are applied, so mixed fp8/q8 saves load. Plain fp8 checkpoints are untouched.

End to end on an M5 Max (save ideogram4 with a character LoRA at 1.0, generate from the saved model, fixed seed):
- main: output pixel-identical to a no-LoRA baseline (max abs diff 0) — the LoRA silently vanished.
- this PR: output matches the live-LoRA render to quantization noise (mean abs diff 5.9 vs 63.8 against no-LoRA).

Two fast tests cover the rebuild and the fp8 no-op. Suite on this PR merged into current main: 512 passed on mlx 0.31.0 and 0.32.0.

### DIFF
--- a/src/mflux/models/common/lora/mapping/lora_saver.py
+++ b/src/mflux/models/common/lora/mapping/lora_saver.py
@@ -5,6 +5,20 @@ from mflux.models.common.lora.layer.fused_linear_lora_layer import FusedLoRALine
 from mflux.models.common.lora.layer.linear_lora_layer import LoRALinear
 
 
+def _is_fp8_linear(linear) -> bool:
+    # fp8 layers (e.g. Ideogram 4's Fp8Linear) store raw uint8 codes in .weight plus a
+    # per-row weight_scale. A float delta cannot be folded into the codes directly:
+    # `delta.astype(uint8)` truncates the (small) LoRA delta to zero, silently destroying
+    # the LoRA while leaving the base intact.
+    weight = getattr(linear, "weight", None)
+    return (
+        weight is not None
+        and weight.dtype == mx.uint8
+        and hasattr(linear, "weight_scale")
+        and not isinstance(linear, nn.QuantizedLinear)
+    )
+
+
 class LoRASaver:
     @staticmethod
     def bake_and_strip_lora(module: nn.Module) -> nn.Module:
@@ -20,14 +34,18 @@ class LoRASaver:
 
         def _bake_single(lora_layer: LoRALinear) -> nn.Module:
             base_linear = lora_layer.linear
+            if _is_fp8_linear(base_linear):
+                return LoRASaver._fold_fp8_loras_to_q8(base_linear, [lora_layer])
             LoRASaver._apply_lora_delta(base_linear, lora_layer)
             return base_linear
 
         def _bake_fused(fused_layer: FusedLoRALinear) -> nn.Module:
             base_linear = fused_layer.base_linear
-            for lora in fused_layer.loras:
-                if isinstance(lora, LoRALinear):
-                    LoRASaver._apply_lora_delta(base_linear, lora)
+            loras = [lora for lora in fused_layer.loras if isinstance(lora, LoRALinear)]
+            if _is_fp8_linear(base_linear):
+                return LoRASaver._fold_fp8_loras_to_q8(base_linear, loras)
+            for lora in loras:
+                LoRASaver._apply_lora_delta(base_linear, lora)
             return base_linear
 
         def _walk(obj, parent=None, attr_name=None, idx=None):
@@ -61,6 +79,28 @@ class LoRASaver:
 
         _walk(module, None, None, None)
         return module
+
+    @staticmethod
+    def _fold_fp8_loras_to_q8(base_linear: nn.Module, lora_layers: list[LoRALinear]) -> nn.Module:
+        # Dequantize the fp8 base ONCE, add the LoRA delta(s) in float32, and requantize to
+        # MLX q8 (group-64 affine keeps more mantissa than fp8-e4m3, so this loses no
+        # quality). Besides making the bake CORRECT on fp8 bases, the folded layer uses
+        # MLX's fused quantized-matmul kernel instead of materializing the full
+        # higher-precision weight matrix on every forward, which is also faster.
+        dense = mx.from_fp8(base_linear.weight, dtype=mx.float32) * base_linear.weight_scale[:, None]
+        merged = dense
+        for lora_layer in lora_layers:
+            delta = mx.transpose(mx.matmul(lora_layer.lora_A, lora_layer.lora_B))
+            merged = merged + lora_layer.scale * delta.astype(mx.float32)
+        bias = getattr(base_linear, "bias", None)
+        compute_dtype = getattr(base_linear, "compute_dtype", mx.bfloat16)
+        linear = nn.Linear(merged.shape[1], merged.shape[0], bias=bias is not None)
+        linear.weight = merged.astype(compute_dtype)
+        if bias is not None:
+            linear.bias = bias
+        quantized = nn.QuantizedLinear.from_linear(linear, group_size=64, bits=8)
+        mx.eval(quantized.parameters())
+        return quantized
 
     @staticmethod
     def _apply_lora_delta(base_linear: nn.Module, lora_layer: LoRALinear) -> None:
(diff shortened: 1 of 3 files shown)


plz12345:
@filipstrand any chance this will be merged?  I've been running off of it locally for a couple of weeks, if that holds any weight/value.

fxd0h:
For reviewers: this is a silent failure — there's no error, the bake just no-ops and the LoRA quietly disappears from the output, so it's easy to hit without noticing. The fix only touches the fp8 path (non-fp8 layers are untouched) and is independent of the other Ideogram 4 PRs, so it can land on its own.

anthonywu (approved):
if test suite result is posted, I would insta merge this cc: @filipstrand

fxd0h:
@anthonywu test output below. The PR also grew a second half while producing it: the fold alone turned the silent LoRA loss into a loud load failure (the checkpoint holds q8 `scales`/`biases` where `Fp8Linear` expects `weight_scale`), so a second commit rebuilds those layers as `QuantizedLinear` at load and the round trip completes. Description updated.

Suites, this PR merged into current main (M5 Max 40-core GPU, macOS 26.5, Python 3.12):

```
mlx 0.31.0: 512 passed, 83 deselected
mlx 0.32.0: 512 passed, 83 deselected
```

End to end: `mflux-save -m ideogram4` with [Vintage Anime](https://civitai.com/models/2685958/vintage-anime-ideogram-4) at scale 1.0, then generate from the saved model. Seed 21, V4_DEFAULT_20, "vintage anime" in the caption (the LoRA's trigger phrase). Mean abs pixel diff:

| generated from saved model | vs live LoRA | vs no LoRA |
|---|---|---|
| saved on main | 47.4 | **0.00** — the LoRA silently vanished |
| saved with this PR | **4.3** | 47.4 — the LoRA survived |

The 4.3 is q8 quantization noise: the fold requantizes the merged weights, so bit-identity to the live path is not possible. The 0.00 is exact: without the delta, saved-on-main is the base model.

<details><summary>baseline, no LoRA</summary></details>
<details><summary>live LoRA</summary></details>
<details><summary>saved on main — pixel-identical to the baseline</summary></details>
<details><summary>saved with this PR — the style survives</summary></details>

Two fast tests in the PR cover the q8 rebuild and the fp8 no-op.

plz12345:
Reviewed and reproduced this locally. Confirming the fix does what it says — LGTM.

Setup: macOS 26.5.2 / arm64, mlx 0.31.0, this branch merged into current `main` (6 commits behind, merges with zero conflicts). Everything below ran under `HF_HUB_OFFLINE=1` as a hard guard, so no weights were downloaded.

**Test results**

- `tests/test_ideogram_q8_folded_loading.py` — 2 passed in 1.9s.
- Offline subset (`tests/arg_parser tests/callbacks tests/cli tests/metadata tests/resolution tests/schedulers tests/utils tests/common` + root `tests/test_*.py`) — 276 passed, 1 failed. The failure is `test_metadata_complete`, which wants to `snapshot_download` Z-Image-Turbo and only fails because of the offline guard. No new failures.

**Independent repro of the actual claim**

The two tests added here compare a baked-q8 layer against a *loaded* q8 layer, so they cover the load-path rebuild but never check fp8 → q8 fidelity itself. To check that part I wrote a separate probe: build a 512×512 weight, quantize it to fp8-e4m3 with a per-row scale the way a real checkpoint stores it, attach a rank-16 LoRA, then compare the baked layer's output against the live (unbaked) wrapper.

| | baked layer type | LoRA's real effect | bake error | error / signal |
|---|---|---|---|---|
| `main` | `Fp8Linear` | 0.180137 | 0.180137 | **1.000** |
| this PR | `QuantizedLinear` | 0.180137 | 0.003049 | **0.017** |

On `main` the bake error equals the LoRA's effect *exactly*, and baked-vs-no-LoRA comes out at 0.000000 — the delta is bit-for-bit gone, exactly as described. With this PR it survives to within 1.7% of the LoRA's own magnitude. 

<details>
<summary>probe script</summary>

```python
import mlx.core as mx
from mlx import nn

from mflux.models.common.lora.layer.linear_lora_layer import LoRALinear
from mflux.models.common.lora.mapping.lora_saver import LoRASaver
from mflux.models.ideogram4.model.ideogram4_transformer.fp8_linear import Fp8Linear

mx.random.seed(1234)
OUT, IN = 512, 512
FP8_MAX = 448.0

# A plausible transformer weight matrix, quantized to fp8-e4m3 with a per-row scale.
dense = mx.random.normal((OUT, IN)).astype(mx.float32) * 0.02
scale = mx.max(mx.abs(dense), axis=1) / FP8_MAX
codes = mx.to_fp8((dense / scale[:, None]).astype(mx.float32))

base = Fp8Linear(IN, OUT, bias=True)
base.weight = codes
base.weight_scale = scale.astype(mx.float32)
base.bias = mx.random.normal((OUT,)).astype(mx.bfloat16) * 0.01

lora = LoRALinear.from_linear(base, r=16, scale=1.0)
lora.lora_A = mx.random.normal((IN, 16)).astype(mx.float32) * 0.05
lora.lora_B = mx.random.normal((16, OUT)).astype(mx.float32) * 0.05

x = mx.random.normal((8, IN)).astype(mx.bfloat16)
out_no_lora = base(x).astype(mx.float32)
out_live_lora = lora(x).astype(mx.float32)


class Holder(nn.Module):
    def __init__(self, layer):
        super().__init__()
        self.qkv = layer


holder = Holder(lora)
LoRASaver.bake_and_strip_lora(holder)
out_baked = holder.qkv(x).astype(mx.float32)

signal = float(mx.mean(mx.abs(out_live_lora - out_no_lora)))
err = float(mx.mean(mx.abs(out_baked - out_live_lora)))
print(f"baked layer type       : {type(holder.qkv).__name__}")
print(f"|live_lora - no_lora|   = {signal:.6f}  (the LoRA's real effect)")
print(f"|baked    - live_lora|  = {err:.6f}  (bake error)")
print(f"error / signal          = {err / signal:.3f}")
```
</details>

**One caveat, not a blocker.** My first pass at that probe filled `.weight` with uniform-random uint8 and this PR came out catastrophically wrong (error 12.25 against a signal of 0.096). That was my test's fault, not the code's — uniform-random e4m3 codes span roughly 1e6 of dynamic range inside a single group-64 block, which affine q8 cannot represent. Real per-row-scaled weights are well conditioned, hence the table above. Worth noting that the correctness of the q8 fold does rest on that conditioning assumption, and nothing in the suite currently pins it. If you feel like adding the probe (or something like it) as a third test, that would close the gap, but I'm happy to merge without it.

**Nits:** `ruff check` is clean, but `ruff format` wants two cosmetic changes in `ideogram4_initializer.py` — a double blank line before `_rebuild_q8_folded_layers`, and the `_nn.QuantizedLinear(...)` call collapsed onto one line.

**Sequencing note for anyone following along:** this needs to land before #422 (LyCORIS LoKr). That PR rewrites `lora_saver.py` — `_apply_lora_delta` is replaced by `_bake_delta_into_linear` — so the two conflict directly, and its `_dense_weight()` handles `nn.QuantizedLinear` but not `Fp8Linear`, which would reintroduce this exact bug. It also flips baking to a load-time default for every model, which would promote the silent LoRA loss from "`mflux-save` with Ideogram 4" to "every Ideogram 4 generate with a LoRA". Merging this first, then reconciling #422 on top so LoRA and LoKr deltas both route through `_fold_fp8_loras_to_q8`.

Thanks for the thorough writeup on this one — the fp8-codes-are-not-floats failure mode is a nasty one to spot.

PR #441, opened 2026-06-10
Ideogram 4: optional CFG truncation (cfg_end) and skip the unconditional pass at guidance 1.0

fxd0h:
## What
Two complementary speedups for Ideogram 4's double-transformer CFG:

1. **guidance == 1.0 skips the unconditional forward entirely.** The negative term vanishes exactly (`v = 1.0*pos + 0.0*neg = pos`), so this is a pure 2x win for CFG-free steps with bit-identical output.
2. **`generate_image(cfg_end=<fraction>)`** runs CFG only for that leading fraction of steps, cond-only for the rest. Guidance shapes composition early; the late steps mostly refine.

## Benchmark (M5 Max, 768x768)
| Config | Time |
|---|---|
| 25 steps, full CFG | 89.7s |
| 25 steps, `cfg_end=0.6` | 74.5s (output visually indistinguishable) |
| `V4_TURBO_12` | 29.8s |
| `V4_TURBO_12` + `cfg_end=0.6` | 24.7s |

Default behavior (`cfg_end=None`) is unchanged. Full fast test suite passes (516 passed).

### DIFF
--- a/src/mflux/models/ideogram4/cli/ideogram4_generate.py
+++ b/src/mflux/models/ideogram4/cli/ideogram4_generate.py
@@ -31,6 +31,13 @@ def main():
         action="store_true",
         help="Fail when an Ideogram 4 JSON caption has schema warnings.",
     )
+    parser.add_argument(
+        "--cfg-end",
+        type=float,
+        default=None,
+        help="Fraction of steps (0-1) that run CFG; the remaining steps run cond-only "
+        "(guidance 1.0, skipping the unconditional forward). Lower = faster. Default: full CFG.",
+    )
     args = parser.parse_args()
 
     model_name = args.model or "ideogram4"
@@ -74,6 +81,7 @@ def main():
                 height=height,
                 preset=args.preset,
                 strict_caption_validation=args.strict_caption_validation,
+                cfg_end=args.cfg_end,
             )
             image.save(path=args.output.format(seed=seed), export_json_metadata=args.metadata)
     except (StopImageGenerationException, PromptFileReadError) as exc:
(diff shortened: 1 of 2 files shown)


plz12345:
Nice addition — `cfg_end` is a clean speed/quality knob. One gap worth flagging:
it's only wired into `generate_image()`, not the `mflux-generate-ideogram4` CLI,
so command-line (and downstream app) users can't actually reach it.

`ideogram4_generate.py` never adds a `--cfg-end` argument, so `cfg_end` stays
`None` for every CLI invocation. And since the shipped presets are
`(3.0,) + (7.0,)*N` — guidance is never exactly `1.0` — the `_denoise_step`
skip-at-`guidance == 1.0` fast path also never triggers on its own. So without a
way to set `cfg_end`, **both halves of this PR are effectively dormant for CLI
users**; the feature is reachable only from the Python API.

A few lines in `ideogram4_generate.py` close it:

```python
parser.add_argument(
    "--cfg-end",
    type=float,
    default=None,
    help="Fraction of steps (0-1) that run CFG; the rest run cond-only "
         "(guidance 1.0, skipping the unconditional forward).",
)
...
model.generate_image(
    ...,
    cfg_end=args.cfg_end,
)
```

Happy to send a follow-up PR if helpful, but I think it fits better in yours.  FWIW, I tested this in CLI locally already.

64GB Macbook Pro M5 Pro

1024x1024 image
- without change: 185 seconds
- with 0.60 - 159 seconds

That's a single test, not averaged over a few.

fxd0h:
Thanks — you're right, that was a real gap. I've wired `--cfg-end` into `mflux-generate-ideogram4` in this PR; the argument forwards straight to `generate_image(cfg_end=...)`, so both the CFG truncation and the `guidance == 1.0` skip are now reachable from the CLI, not just the Python API. Appreciate you testing it and posting the 1024² numbers.

anthonywu (approved):
tight fix, recommending fast merge by @filipstrand

PR #444, opened 2026-06-16
fix: stepwise support for Ideogram 4 generation

plz12345:
## Problem

Stepwise image saving (`--stepwise-image-output-dir`) crashes when used with Ideogram 4 models:

```
ValueError: [broadcast_shapes] Shapes (1,32,64,64) and (1,128,1,1) cannot be broadcast.
```

`StepwiseHandler._save_stepwise_image` calls `unpack_latents` and then checks `hasattr(self.model.vae, "decode_packed_latents")` to decide whether to call that method or fall back to `vae.decode`. Ideogram 4 uses `Flux2VAE`, which **does** have `decode_packed_latents`, so the check always succeeds.

The difference is what `unpack_latents` returns for each model:

| Model | After `unpack_latents` | Channels | Next step |
|---|---|---|---|
| FLUX.2 | Patchified latents (BN-denorm still needed) | 128 | `decode_packed_latents` (BN denorm + unpatchify → 32ch) |
| Ideogram 4 | VAE-ready latents (shift/scale already applied + reshape done) | 32 | `vae.decode` directly |

Passing 32-channel VAE-ready latents into `decode_packed_latents` causes a channel shape mismatch — that method expects 128 channels, and the BN denorm parameters broadcast against the wrong axis.

## Fix

Add a channel count guard before calling `decode_packed_latents`:

```python
vae_latent_channels = getattr(self.model.vae, "latent_channels", 32)
if hasattr(self.model.vae, "decode_packed_latents") and unpack_latents.shape[1] > vae_latent_channels:
    stepwise_decoded = self.model.vae.decode_packed_latents(unpack_latents)
else:
    stepwise_decoded = self.model.vae.decode(unpack_latents)
```

`unpack_latents.shape[1]` is 128 for FLUX.2 (still patchified, needs the extra decode step) and 32 for Ideogram 4 (already unpacked). The `> vae_latent_channels` guard routes each model to the correct path without requiring a new abstract method or Ideogram-specific isinstance check.

`latent_channels` defaults to `32` via `getattr` fallback, matching all current VAEs that don't explicitly declare it.

## Verification

Reproduced the crash and confirmed the fix end-to-end on `ideogram-ai/ideogram-4-fp8`:

```
# main's handler
ValueError: [broadcast_shapes] Shapes (1,32,64,64) and (1,128,1,1) cannot be broadcast.   exit 1

# with this fix
exit 0 → 13 stepwise PNGs + composite, clean noise → image progression
```

Also adds `tests/callbacks/test_stepwise_handler_vae_routing.py` — three `@pytest.mark.fast` tests (0.6s, no weights, no image generation) asserting which decode path the handler picks for Ideogram 4 (`vae.decode`), FLUX.2 (`decode_packed_latents`), and a VAE with no `decode_packed_latents` at all (fallback). They fake the model and VAE but use the real latent creators, so the shape contract that actually broke is exercised rather than mocked away.

Reverting the fix makes exactly the Ideogram case fail (`decode_packed_latents expects 128 channels, got 32`) while the other two stay green.

See the test-run comment below for full fast/slow suite results.

### DIFF
--- a/src/mflux/callbacks/instances/stepwise_handler.py
+++ b/src/mflux/callbacks/instances/stepwise_handler.py
@@ -81,7 +81,12 @@ class StepwiseHandler(BeforeLoopCallback, InLoopCallback, InterruptCallback):
         time_steps: tqdm,
     ) -> None:
         unpack_latents = self.latent_creator.unpack_latents(latents=latents, height=config.height, width=config.width)
-        if hasattr(self.model.vae, "decode_packed_latents"):
+        # Flux2's unpack_latents returns 128-channel patchified latents that still need
+        # decode_packed_latents (BN denorm + unpatchify → 32ch).  Ideogram4's unpack_latents
+        # applies its own shift/scale and already returns 32-channel VAE-ready latents, so
+        # calling decode_packed_latents on them causes a channel shape mismatch.
+        vae_latent_channels = getattr(self.model.vae, "latent_channels", 32)
+        if hasattr(self.model.vae, "decode_packed_latents") and unpack_latents.shape[1] > vae_latent_channels:
             stepwise_decoded = self.model.vae.decode_packed_latents(unpack_latents)
         else:
             stepwise_decoded = self.model.vae.decode(unpack_latents)
--- a/tests/callbacks/test_stepwise_handler_vae_routing.py
+++ b/tests/callbacks/test_stepwise_handler_vae_routing.py
@@ -1,0 +1,104 @@
+from types import SimpleNamespace
+
+import mlx.core as mx
+import pytest
+
+from mflux.callbacks.instances.stepwise_handler import StepwiseHandler
+from mflux.models.common.config.config import Config
+from mflux.models.common.config.model_config import ModelConfig
+from mflux.models.flux.latent_creator.flux_latent_creator import FluxLatentCreator
+from mflux.models.flux2.latent_creator.flux2_latent_creator import Flux2LatentCreator
+from mflux.models.ideogram4.latent_creator.ideogram4_latent_creator import Ideogram4LatentCreator
+
+SIZE = 256
+
+
+def _decoded(height: int, width: int) -> mx.array:
+    return mx.zeros((1, 3, height, width))
+
+
+class RecordingPackedVAE:
+    """Flux2-style VAE: decode_packed_latents does BN denorm + unpatchify, so it only
+    accepts the 4x-patchified channel count and rejects VAE-ready latents."""
+
+    latent_channels = 32
+
+    def __init__(self):
+        self.calls = []
+
+    def decode(self, latents: mx.array) -> mx.array:
+        self.calls.append("decode")
+        if latents.shape[1] != self.latent_channels:
+            raise ValueError(f"decode expects {self.latent_channels} channels, got {latents.shape[1]}")
+        return _decoded(SIZE, SIZE)
+
+    def decode_packed_latents(self, latents: mx.array) -> mx.array:
+        self.calls.append("decode_packed_latents")
+        if latents.shape[1] != 4 * self.latent_channels:
+            raise ValueError(
+                f"decode_packed_latents expects {4 * self.latent_channels} channels, got {latents.shape[1]}"
+            )
+        return _decoded(SIZE, SIZE)
+
+
+class RecordingPlainVAE:
+    """Flux-style VAE with no decode_packed_latents at all."""
+
+    latent_channels = 16
+
+    def __init__(self):
+        self.calls = []
+
+    def decode(self, latents: mx.array) -> mx.array:
+        self.calls.append("decode")
+        return _decoded(SIZE, SIZE)
+
+
+def _handler(tmp_path, vae, latent_creator) -> StepwiseHandler:
+    model = SimpleNamespace(vae=vae, bits=None, lora_paths=None, lora_scales=None)
+    return StepwiseHandler(model=model, output_dir=str(tmp_path), latent_creator=latent_creator)
+
+
+def _config(model_config: ModelConfig) -> Config:
+    return Config(model_config=model_config, num_inference_steps=1, height=SIZE, width=SIZE)
+
+
+@pytest.mark.fast
+def test_stepwise_ideogram4_uses_plain_decode(tmp_path) -> None:
+    # Ideogram4's unpack_latents applies shift/scale and returns VAE-ready 32ch latents.
+    vae = RecordingPackedVAE()
+    handler = _handler(tmp_path, vae, Ideogram4LatentCreator)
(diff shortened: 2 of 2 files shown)


anthonywu (approved):
if the test suite passes, nominating for fast merge by @filipstrand

plz12345:
This test suite?  I'm lost in what you're asking for, if not.
image

anthonywu:
@plz12345 the PR only exercises the fast tests, there are all the slow image gen tests on the repo too, do `make test` in repo root

anthonywu:
## Analysis: Fast Tests in GitHub Actions vs `make test`

### GitHub Actions "Fast Tests" workflow

In `.github/workflows/tests.yml`, the **"Fast Tests"** workflow runs on `push`/`pull_request` to `main` on a macOS-14 (Apple Silicon) runner and executes:

```bash
python -m pytest -m fast -v
```

The `-m fast` flag selects **only** tests marked with `@pytest.mark.fast`. Markers are registered in `pyproject.toml`:

```toml
markers = [
  "fast: marks tests as fast (no image generation)",
  "slow: marks tests as slow (generates images)",
]
```

So "fast" = unit/logic tests that don't generate images (e.g. the callback, memory-saver, battery-saver, depth, and most ideogram4 non-golden tests). Concretely: ~266 `fast`-marked test items across 32 files.

### `make test` scope

`make test` runs **all** tests with no marker filter:

```bash
uv run python -m pytest
```

That includes:
- ~266 `fast` tests
- ~40 `slow` tests (the actual image-generation tests — flux2, fibo, qwen, z-image, ernie, seedvr2 upscale, golden image diffs, LoRA/training round-trips, model saving, VAE tiling, etc., across 22 files)
- any unmarked tests

There are also convenience targets `make test-fast` (`-m fast`) and `make test-slow` (`-m slow`) mirroring the split.

### Key differences

| | GH Actions "Fast Tests" | `make test` |
|---|---|---|
| Selection | `-m fast` only | all tests |
| Slow/image-gen tests | **excluded** | **included** |
| Runner | macos-14 (M1) CI | your local machine (Apple Silicon) |
| Env install | `uv pip install -e .` + pytest only | `uv pip install -e '.[dev]'` (pinned MLX for testing) |
| `MFLUX_PRESERVE_TEST_OUTPUT` | not set | set to `1` |

### Summary

The practical upshot: CI only gates on the fast, non-image-generating tests (cheap, no model downloads/inference), while `make test` is the full local suite that also runs the slow image-generation tests that download weights and actually generate images. The slow tests are intentionally kept out of PR CI for speed/cost reasons.

One thing worth noting: CI installs `-e .` (no `[dev]` extras) plus bare pytest, whereas `make test` installs the `[dev]` extra for a pinned MLX version — so the two don't have identical dependency resolution either.

---

*Posted by @anthonywu via Amp*

plz12345:
Ack, will run slow tests and see about adding small coverage for the revised code.

ianscrivener:
TLDR;
```
# fast tests, no image generation
# MFLUX_PRESERVE_TEST_OUTPUT=1 uv run python -m pytest
make test-fast     

# slow tests, image generation
# MFLUX_PRESERVE_TEST_OUTPUT=1 uv run python -m pytest -m slow
make test-slow

# full suite
# MFLUX_PRESERVE_TEST_OUTPUT=1 uv run python -m pytest -m fast 
make test            
```
Yeah... I was scratching my head on "test suite"suite as well...

There's nothing documented though it is mentioned in the Changelog and there's a Cursor skill that mentions it - .cursor/skills/mflux-testing/SKILL.md

plz12345:
Regarding Cursor, we should move to a root AGENTS.md .  Users can symlink their tool of choice locally, if needed, but that's a more agnostic solution.

anthonywu:
@ianscrivener acknowledge the low-doc of the test suite, it was only up to filip and me to a lesser extent to prove to each other we had run the whole suite before a merge

plz12345:
## Full test run (fast + slow)

Ran the whole ideogram surface locally, including the slow image-generating tests, per @anthonywu's note above.

**Environment:** macOS (Apple Silicon), Python 3.11.13, mlx 0.31.0 (the `[dev]` pin), `ideogram-ai/ideogram-4-fp8` (27.55 GB).

### Fast tests — `pytest -k ideogram -m "not slow"`

```
46 passed, 3 skipped, 515 deselected in 0.95s
```

The 3 skips are `test_ideogram4_checkpoint_layout.py`, which require a local Ideogram checkpoint on disk.

### Slow tests — `pytest -k ideogram -m slow`

```
1 failed, 1 passed, 562 deselected in 1173.21s (0:19:33)
```

| Test | Result |
|---|---|
| `test_save_and_load_quantized_ideogram4` | ✅ PASSED |
| `test_ideogram4_jazz_fest_json_caption` (golden) | ❌ FAILED — 28.4% different, `mismatch_threshold=0.15` |

### The golden failure is pre-existing and unrelated to this PR

Confirmed rather than assumed:

1. **Ran the same golden test on `main`** — fails identically at **28.4%**.
2. **The two output images are pixel-identical** (`np.array_equal` → `True`, max abs diff `0`). This PR changes nothing on that code path.
3. **The path never touches the changed code** — the golden helper never registers a stepwise callback, so `StepwiseHandler` is never constructed there.

The generated image is structurally correct — same composition, typography, and pose as the reference — differing only in hue/contrast. That's numerical drift (most likely the mlx version pinned here vs. whatever produced the committed reference PNG), not a decode bug; a genuine channel mismatch produces garbage, not a well-formed poster.

Not addressing that here — it's orthogonal to this fix and wants its own PR to refresh the reference image or widen the threshold.

### Stepwise repro (the path this PR actually fixes)

No existing test exercises `--stepwise-image-output-dir`, so I verified it directly:

```
mflux-generate-ideogram4 --prompt '<json caption>' --seed 42 \
  --width 512 --height 512 --preset V4_TURBO_12 \
  --stepwise-image-output-dir ./steps
```

```
# main's handler
ValueError: [broadcast_shapes] Shapes (1,32,64,64) and (1,128,1,1) cannot be broadcast.   exit 1

# with this fix
exit 0 → 13 stepwise PNGs + composite, clean noise → image progression
```

ianscrivener:
> Regarding Cursor, we should move to a root AGENTS.md . Users can symlink their tool of choice locally, if needed, but that's a more agnostic solution.

@plz12345, agree I have done just that in https://github.com/mflux-community/mflux-fork

ianscrivener:
> @ianscrivener acknowledge the low-doc of the test suite, it was only up to filip and me to a lesser extent to prove to each other we had run the whole suite before a merge

@anthonywu, Just an observation, not a criticism. 
Single/dual maintainer repos have have less documentation requirements when compared to active community projects.
While I'd love to get down and dirty in the code... my primary focus for MFlux-Community is set it up for success: doco, testing, less rough edges for new end-users, rebuild trust that MFlux can support SOTA models and functionality in a timely manner.

PR #445, opened 2026-06-17
feat: load ideogram 4 from mlx-forge converted checkpoints (bf16, int8, int4)

plz12345:
### Summary

Adds a second checkpoint-loading path for Ideogram 4 so it can load **MLX-native weights converted by [mlx-forge](https://github.com/dgrauet/mlx-forge)**, in addition to the existing FP8 layout. Because the creator gates the FP8 weights on HF, I opted to pre-cache and also gate the quantized variants.  The FP8 path is unchanged.  This is primarily aimed at reducing generation times, since mflux currently does realtime quantization during the gen process.  My testing on a 64GB M5 Pro had generation times for mid-res turbo (12 step) images drop by 30 seconds (75 seconds to 45 seconds), purely due to properly pre-caching quantized weights and loading direct to Metal.

Ideogram 4 ships its linear weights as `float8_e4m3fn` (raw FP8 bits + per-row float32 scales). `Fp8Linear` re-dequantizes those to bf16 on **every** matmul, and init forces a full synchronous dequant of all ~420 linears. The mlx-forge format moves that work to a one-time offline conversion and, for the quantized build, stores weights in MLX's native int8 layout (`mx.quantized_matmul`).

### How it works

Detection is via `split_model.json` at the model root. If present, the mlx-forge path is taken; otherwise the original FP8 layout loads exactly as before.

- **bf16**: `Fp8Linear.__call__` falls through to a plain `matmul` when the loaded weight isn't `uint8` (skips `mx.from_fp8` + scale). Weights load from flat `{component}.safetensors`, prefix stripped, via `model.update(..., strict=False)`.
- **int8**: before loading, every `Fp8Linear` is swapped for an `nn.QuantizedLinear` (same dims, bias preserved) using the standard `leaf_modules → tree_map_with_path → update_modules` idiom, so the packed `uint32` weight + bf16 `scales`/`biases` populate directly.
- Config loading is refactored to take a config-file `Path`, so the FP8 subdir layout (`transformer/config.json`) and the mlx-forge flat layout (`conditional_transformer_config.json`) share one parser.

### Backward compatibility

No behavior change for existing FP8 checkpoints — the mlx-forge branch only activates when `split_model.json` is found. Two files touched: `ideogram4_initializer.py`, `fp8_linear.py`.

### Testing

New `tests/test_ideogram4_mlx_forge_loader.py` (7 tests, `fast`, no weights/network): bf16 fallthrough numerics, `Fp8Linear → QuantizedLinear` swap (nested-in-list modules, bias preserved, embeddings left as `nn.Embedding`), and `split_model.json` detection/bits parsing. Validated end-to-end on M-series hardware: **bf16 and int8, base generation and LoRA, all working.**

### Related

- mlx-forge conversion recipe: dgrauet/mlx-forge#19
- Converted weights: [`MLXBits/ideogram-4-mlx`](https://huggingface.co/MLXBits/ideogram-4-mlx) (bf16), [`MLXBits/ideogram-4-mlx-q8`](https://huggingface.co/MLXBits/ideogram-4-mlx-q8) (int8)

### DIFF
--- a/src/mflux/models/ideogram4/ideogram4_initializer.py
+++ b/src/mflux/models/ideogram4/ideogram4_initializer.py
@@ -61,6 +61,10 @@ class Ideogram4Initializer:
         )
         if root_path is None:
             raise ValueError(f"No model path resolved for {path!r}")
+        # mlx-forge repos (resolved from an HF id or a local path) carry split_model.json;
+        # load them directly instead of demanding the FP8 checkpoint layout.
+        if (root_path / "split_model.json").exists():
+            return root_path
         return Ideogram4WeightDefinition.validate_fp8_checkpoint(root_path)
 
     @staticmethod
--- a/src/mflux/models/ideogram4/ideogram4_initializer.py
+++ b/src/mflux/models/ideogram4/ideogram4_initializer.py
@@ -3,6 +3,8 @@ from pathlib import Path
 from typing import Any
 
 import mlx.core as mx
+import mlx.nn as nn
+from mlx.utils import tree_map_with_path, tree_unflatten
 
 from mflux.callbacks.callback_registry import CallbackRegistry
 from mflux.models.common.config import ModelConfig
@@ -15,6 +17,7 @@ from mflux.models.common.weights.loading.weight_loader import WeightLoader
 from mflux.models.flux2.model.flux2_vae.vae import Flux2VAE
 from mflux.models.ideogram4.model.ideogram4_text_encoder import Qwen3TextEncoder
 from mflux.models.ideogram4.model.ideogram4_transformer import Ideogram4Config, Ideogram4Transformer
+from mflux.models.ideogram4.model.ideogram4_transformer.fp8_linear import Fp8Linear
 from mflux.models.ideogram4.weights import Ideogram4LoRAMapping, Ideogram4WeightDefinition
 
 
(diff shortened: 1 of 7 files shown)


fxd0h:
heads up: this and [#439](https://github.com/filipstrand/mflux/pull/439) both touch `ideogram4_initializer.py`, so one will need a rebase after the other. they're complementary though: #439 fixes Ideogram 4 LoRA output getting corrupted at guidance > 1, and that fix lives in `_apply_lora`, which this PR doesn't change. so the LoRA path here still wants #439 once you go above guidance 1.

fxd0h:
@plz12345 heads-up on an overlap, so we can pick an order now instead of at conflict time.

This PR rewrites `fp8_linear.py` so `Fp8Linear.__call__` falls through to a plain matmul when the weights are not uint8. That is exactly the assumption my #440 is built on: it folds a LoRA onto an fp8 base layer to q8 instead of letting it silently degrade, and it keys off that same dispatch.

Whichever of the two lands first, the other needs a small rebase rather than a rethink, so this is not a problem, just something better agreed than discovered. I am happy to go second and rebase #440 onto this, since your change is the broader one and mine is a narrow fix on top of the behaviour it establishes. If you would rather I go first so you can build on the folded path, that works too, just say which.

Adjacent but not conflicting: #439 and #441 are on the Ideogram side as well. #441 in particular still has your open point from June about `ideogram4_generate.py` never wiring `--cfg-end`, which I owe you a fix for.

plz12345:
Sure, no worries.  If this project isn't abandoned, I'll rebase as needed.  I get the feeling the project owner isn't able to contribute the time needed for a fast-moving tech stack, so I'm considering forking it, since I do.

fxd0h:
Correction to my last comment: I said I owed you the `--cfg-end` CLI wiring. I don't, it's been in #441 since 20 June, right after you flagged it. I replied to you there at the time and then lost track of my own answer. Sorry for the noise.

The rest stands: happy to go second and rebase #440 onto this one.

plz12345:
Wondering if I should retract this one and redo as just native conversion.  The repos would still be gated for the quants, since official source is gated.  Thoughts?

plz12345:
Closing this in favour of #559, which takes the other route.

This PR taught mflux to *read* an externally converted checkpoint layout. The concern raised about that is fair, and I think it's the right call: it makes a third-party tool's on-disk format (`split_model.json`, flat root-level configs) a supported input surface that mflux then has to track across that tool's changes, for exactly one model. Nothing else in the repo loads that way.

The better shape is for mflux to produce the artefacts itself, through the same `mflux-save` path every other model already uses, and it turns out there was a concrete reason nobody had noticed that path was unavailable for Ideogram 4: **it was silently broken.** Every weight-bearing linear in the model is an `Fp8Linear`, which defines no `to_quantized`, and the transformers and text encoder were additionally marked `skip_quantization=True`. So `mflux-save --model ideogram-4-fp8 -q 4` wrote an FP8 checkpoint stamped `quantization_level: 4` with only the VAE quantized — while the model README already advertised "quantization via `mflux-save`" as supported. Fixing that is #559, along with two follow-on bugs it exposed in the saved-checkpoint reload path.

The result is `-q 8` at 26 GB and `-q 4` at 14 GB, in mflux's native layout, both visually indistinguishable from FP8.

Not wasted from here — the fp8→int8 detection logic and the loader test carry over, and `_replace_fp8_with_quantized` from this branch is essentially `Fp8Linear.to_quantized` written backwards, which is what #559 adds.

Thanks for the reviews on this one; the pushback pointed at the better design.

PR #448, opened 2026-06-20
Training: add Adafactor and Lion to the optimizer enum

fxd0h:
Exposes `mlx.optimizers.Adafactor` (about half the optimizer state of Adam — useful for memory-bound training) and `Lion` alongside the existing Adam/AdamW. Selected by name in the training config's `optimizer` section. Full fast suite: 516 passed.


plz12345:
Low-risk IMO, since this is just an enum expansion, not a codepath change.  Merging.

PR #449, opened 2026-06-20
Training: support LoRA alpha (effective scale = alpha/rank)

fxd0h:
`LoraTargetSpec` gains an optional `alpha`; the injector applies `scale = alpha/rank` to each trained LoRA layer (the published-recipe convention) instead of a hardcoded 1.0. Threaded through `expand_module_paths_from_targets` and parsed from `lora_layers.targets[]`. The saver already bakes the layer scale, so exported weights carry it. `alpha=None` keeps `scale = 1.0` (unchanged default). Full fast suite: 516 passed.


anthonywu (approved):
tight fix, nominating for quick merge by @filipstrand

PR #452, opened 2026-06-20
Training: bound MLX's buffer-cache pool via cache_limit_gb

fxd0h:
`TrainingSpec.cache_limit_gb` calls `mx.set_cache_limit()` for the whole run. MLX's buffer-cache pool otherwise grows to hold the largest transient allocations seen and is never returned to the OS; bounding it caps resident memory at ~no speed cost (buffers below the cap are still reused, unlike `low_ram`'s clear-every-step). `None` = MLX default (unbounded). Full fast suite: 516 passed.


anthonywu (approved):
LGTM

plz12345:
Merging based on passing review already.

PR #453, opened 2026-06-23
feat: add Krea 2 text-to-image model

plz12345:
## Summary

Adds **Krea 2** text-to-image support — a single-stream MMDiT built on the
Qwen-Image stack. It reuses mflux's existing components and only introduces the
novel parts:

- **Transformer** (new): 28-layer single-stream MMDiT — features 6144, GQA
  (48 query / 12 KV heads, head_dim 128), SwiGLU, 3-axis Flux-style RoPE
  `[32,48,48]`, per-head QK-norm + sigmoid-gated attention, AdaLN-single 6-way
  modulation, plus a `txtfusion` adapter over the text encoder's hidden states.
- **Text encoder** (reused): Qwen3-VL-4B via the common `qwen3_vl` decoder
  layers + flux2's text rotary. Krea 2 conditions on a 12-layer hidden-state tap
  `[2,5,…,35]`, flattened layer-major; the chat-template prefix is stripped so
  only the real prompt tokens condition the DiT.
- **VAE** (reused): the Qwen-Image VAE (`QwenVAE`, Wan2.1 16-ch latent).
- **Samplers**: `er_sde` (default, matches the reference turbo) and `euler`
  (`--scheduler euler`, the official diffusers `FlowMatchEulerDiscreteScheduler`).

## CLI

- `mflux-generate-krea2` — live tqdm denoising bar, `--metadata`,
  `--stepwise-image-output-dir`, memory/battery callbacks.
- `mflux-save` — quantization caching (save a quantized model once, reload
  without re-quantizing).

## Weights

Loads from the official `krea/Krea-2-Turbo` repo. The transformer is the root
`turbo.safetensors` (native single-file, 430 tensors); the VAE, single-file
Qwen3-VL text encoder, and tokenizer come from the repo's subdirs. (The repo's
`transformer/` subdir is diffusers-format with different keys — the root file is
used, which matches the weight mapping here.)

For those that can't cover the unified memory load to do the quantization, I've done one for Q8, which you can get here https://huggingface.co/MLXBits/krea-2-mlx-q8/

## Validation

- `model_index.json` confirms the architecture: `text_encoder_select_layers
  [2,5,…,35]`, `patch_size 2`, `AutoencoderKLQwenImage`, `Qwen3VLModel`.
- Text encoder passes an LM sanity check ("The capital of France is" → " Paris").
- VAE encode→decode round-trips cleanly; transformer/TE weights verified against
  the official shards.
- End-to-end generation produces coherent, prompt-faithful images
  (turbo, 8 steps, CFG 1.0).

## Notes

- Image conditioning (edit / reference) is not wired yet; the Qwen3-VL vision
  tower is dropped at load. txt2img only for now.
- Self-contained: no dependency on other in-flight branches.


filipstrand:
Very nice! Excited to try out this model.

plz12345:
It's decent, but I found quality to be on par with a well-written FLux2 Klein 9B prompt.  But it's new, and LoRAs make the world go 'round.

Pall13:
hi, please can you tell me why I cannot quantise model ? I have this return error ; 
mflux-save \
  -m /Volumes/SSD\ Intenso\ 2\ Tera/Krea-2-Turbo \
  --path /Volumes/SSD\ Intenso\ 2\ Tera/Krea-2-Turbo-8bit \
  -q 4
------
Traceback (most recent call last):
  File "/Users/myname/.local/bin/mflux-save", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/myname/.local/share/uv/tools/mflux/lib/python3.13/site-packages/mflux/models/common/cli/save.py", line 58, in main
    model = model_class(
        quantize=args.quantize,
    ...<3 lines>...
        model_config=model_config,
    )
TypeError: Krea2Image.__init__() got an unexpected keyword argument 'lora_paths'

plz12345:
That traceback is coming from a different Krea-2 branch (mine was first, but there are 2 others...) — the Krea2Image class is from PR #454 (still WIP), not this one. In this PR the class is Krea2 and mflux-save quantization is wired to accept LoRA args, so it shouldn't hit that error.

To try this PR specifically:

`uv tool install git+https://github.com/plz12345/mflux.git@krea-2`
Then quantize with e.g.:

`mflux-save -m krea-2 --path /path/to/Krea-2-4bit -q 4`

Also, I just uploaded this pre-quantized, if you want to skip that or if your Mac doesn't have the unified memory space to quantize locally.

https://huggingface.co/MLXBits/krea-2-mlx-q4

Pall13:
thank you so much [plz12345](https://github.com/plz12345) , you are great!

filipstrand:
I've tried this out over the weekend and it works great, great job @plz12345! I also compared this impl to some to the other ones and they give very similar results given similar inputs. Some different conventions between the PR:s regarding what format of tensors to download but I think this PR works great as it is. A few minor things I'll add now (such as a local end2end image test) then it is ready to be merged.

Issue #423, opened 2026-05-26
LoRA `_split_qkv_mlp_down` splits shared down-projection and corrupts LoRA weights


danielkhalilov:
### Bug

In `models/common/lora/mapping/lora_transforms.py`, `_split_qkv_mlp_down` attempts to split a LoRA down-projection tensor when:

```python
rank % num_splits == 0
```

However, for fused QKV / MLP LoRA weights, the down weight is the shared low-rank input projection. Only the up weight should be split by output dimension.

Splitting the down weight corrupts the LoRA decomposition and can produce incorrect generation results.

This condition is true for common LoRA ranks such as 4, 8, 16, and 32, so the corruption can happen silently for many LoRA configurations.

### Current code

```python
@staticmethod
def _split_qkv_mlp_down(tensor: mx.array, index: int, num_splits: int = 4) -> mx.array:
    rank = tensor.shape[0]
    if rank % num_splits == 0:
        chunk_size = rank // num_splits
        start = index * chunk_size
        end = start + chunk_size
        return tensor[start:end, :]
    else:
        return tensor
```

### Why this is wrong

For a fused QKV / MLP LoRA, the up projection is split across the fused output dimensions. The down projection is shared and should remain intact.

Splitting the down projection changes the effective LoRA rank and breaks the decomposition.

### Suggested fix

Return the down weight unchanged:

```python
@staticmethod
def _split_qkv_mlp_down(tensor: mx.array, index: int, num_splits: int = 4) -> mx.array:
    return tensor
```

### Environment

* mflux installed as a vMLX dependency
* Affects fused QKV / MLP LoRA weights where rank is divisible by `num_splits`
* Common ranks affected: 4, 8, 16, 32

fxd0h:
Fixed in #422 on the 11th. Both `_split_qkv_down` and `_split_qkv_mlp_down` now return the tensor unchanged, for the reason you gave: the down (A) matrix is the shared low-rank input projection, and only the up (B) rows carry concatenated outputs.

Checked on main with the exact condition from your report, rank 16 so `rank % 4 == 0`: all four down transforms return the input untouched, while the up transforms split to 3072 rows for q/k/v and 12288 for the MLP.

Closing.

PR #459, opened 2026-06-28
Fix LoRA loading for fused qkv layers: never split the shared rank/down dimension

### LINKED ISSUE #458: LoRA loader splits the shared rank/down dimension for fused qkv layers → matmul shape error loading kohya/BFL FLUX LoRAs with rank divisible by the split count (e.g. rank 64)
## Summary

`LoraTransforms._split_qkv_mlp_down` (and `_split_qkv_down`) split the **rank/down** dimension of a fused qkv LoRA when the rank is divisible by the number of splits. The down projection of a fused qkv layer is **shared** across q/k/v(/mlp) and is rank-dimensioned, so it must be returned whole — only the **up** projection should be sliced (along its output dimension). Because it is sliced, the resulting per-component `down` has a smaller rank than its matching `up`, and the first denoise step fails with a matmul shape mismatch.

This breaks loading of essentially every kohya / sd-scripts (BFL-format) FLUX LoRA whose rank is divisible by 4 — i.e. ranks 64, 32, 16 — that touches the single-block `linear1` (fused qkv+mlp). Many popular community FLUX.1-dev LoRAs are exactly this format (rank 64, alpha 32).

## Environment

- mflux 0.18.0 (also reproduces on current `main`; the relevant code is byte-identical)
- Apple Silicon (M-series), macOS
- Base model: `FLUX.1-dev`, 8-bit quantized

## Reproduction

Any kohya/BFL-format FLUX.1-dev LoRA with rank 64 triggers it, e.g. [`strangerzonehf/Flux-Super-Realism-LoRA`](https://huggingface.co/strangerzonehf/Flux-Super-Realism-LoRA) (also `strangerzonehf/Flux-Midjourney-Mix2-LoRA`, `prithivMLmods/Flux-Long-Toon-LoRA`, etc.):

```bash
# download the LoRA, then:
mflux-generate \
  --base-model dev -q 8 --steps 4 --guidance 3.5 \
  --prompt "a photorealistic portrait" \
  --lora-paths ./super-realism.safetensors --lora-scales 1.0 \
  --output ./out.png
```

The LoRA keys map and apply, but generation fails at the first denoise step:

```
ValueError: [matmul] Last dimension of first input with shape (1,4608,16)
must match second to last dimension of second input with shape (64,3072)
```

(`4608` = 512 text + 4096 image tokens; `16` = 64 / 4 — the rank wrongly divided by the 4-way split.)

LoRAs in diffusers/PEFT format (pre-split, e.g. `Shakker-Labs/FLUX.1-dev-LoRA-add-details`) load fine because they never hit this transform.

## Root cause

In `src/mflux/models/common/lora/mapping/lora_transforms.py`:

```python
@staticmethod
def _split_qkv_mlp_down(tensor: mx.array, index: int, num_splits: int = 4) -> mx.array:
    rank = tensor.shape[0]
    if rank % num_splits == 0:          # <-- triggers for rank 64/32/16
        chunk_size = rank // num_splits
        start = index * chunk_size
        end = start + chunk_size
        return tensor[start:end, :]      # <-- slices the SHARED rank dim
    else:
        return tensor
```

For a fused qkv layer the LoRA stores one shared `down` of shape `(rank, in_features)` and one `up` of shape `(fused_out, rank)`. Splitting the fused layer into q/k/v(/mlp) must slice only `up` along its **output** dimension (which `_split_qkv_mlp_up` / `_split_qkv_up` already do correctly via the `dims` list); the `down` is shared and must be returned at full rank for every component.

`_split_qkv_down` (the double-block path, `num_splits=3`) has the same defect; it only escapes today because common ranks (64/32/16) aren't divisible by 3. A rank like 48 or 96 would break it too.

## Proposed fix

Make both `_split_qkv_down` and `_split_qkv_mlp_down` return the down tensor unchanged (never split the rank dimension):

```python
@staticmethod
def _split_qkv_down(tensor: mx.array, index: int, num_splits: int = 3) -> mx.array:
    # Fused qkv LoRA shares a single rank-dimensioned down projection; return it
    # whole for every q/k/v component (only the up projection is output-split).
    return tensor

@staticmethod
def _split_qkv_mlp_down(tensor: mx.array, index: int, num_splits: int = 4) -> mx.array:
    return tensor
```

This matches the already-correct `else: return tensor` branch (the non-divisible case that works today) and makes it the behavior for all ranks. I have this fix plus a unit test ready and can open a PR.

## Verification

With the fix, `strangerzonehf/Flux-Super-Realism-LoRA` and `Flux-Midjourney-Mix2-LoRA` load (`1254/1254 keys matched, applied to 494 layers`) and render cleanly on FLUX.1-dev 8-bit. Confirmed for ranks 16/32/64.

*(Found while running mflux on Apple Silicon at RBDS AI Lab.)*

deadmansahil:
## What

`LoraTransforms._split_qkv_down` and `_split_qkv_mlp_down` split the **rank/down**
dimension of a fused qkv LoRA whenever the rank is divisible by the split count
(3 and 4 respectively). The down projection of a fused qkv layer is **shared**
across q/k/v(/mlp) and is rank-dimensioned, so it must be returned whole — only
the **up** projection is sliced (along its output dimension, which
`_split_qkv_*_up` already handle correctly).

This makes both functions return the down tensor unchanged for every component.

## Why

Splitting the rank dimension yields a per-component `down` with a smaller rank
than its matching `up`, so the first denoise step fails:

```
ValueError: [matmul] Last dimension of first input with shape (1,4608,16)
must match second to last dimension of second input with shape (64,3072)
```

This breaks loading of kohya / sd-scripts (BFL-format) FLUX LoRAs whose rank is
divisible by 4 (i.e. ranks 64/32/16 — the common case) that touch the
single-block `linear1` (fused qkv+mlp). Many popular community FLUX.1-dev LoRAs
ship in exactly this format (rank 64, alpha 32). diffusers/PEFT LoRAs are
unaffected because they are already pre-split and never hit this transform.

`_split_qkv_down` (double-block path, `num_splits=3`) has the same defect; it
only escapes today because common ranks aren't divisible by 3 (a rank like 48
or 96 would break it).

The new behavior simply makes the existing, already-correct `else: return tensor`
branch (the non-divisible case that works today) apply to all ranks.

## Changes

- `src/mflux/models/common/lora/mapping/lora_transforms.py`
  - `_split_qkv_down`: return the tensor unchanged (with explanatory comment).
  - `_split_qkv_mlp_down`: return the tensor unchanged (with explanatory comment).
- `tests/test_lora_transforms.py` (new): unit tests asserting fused down
  projections stay full-rank and up projections are output-split, for ranks
  16/32/64, plus a regression check that the rank-64 `up @ down` product is
  well-formed.

## Testing

- New unit tests pass (`pytest tests/test_lora_transforms.py`, marked `fast`).
- End-to-end: `strangerzonehf/Flux-Super-Realism-LoRA` and
  `strangerzonehf/Flux-Midjourney-Mix2-LoRA` (kohya/BFL, rank 64) now load
  (`1254/1254 keys matched, applied to 494 layers`) and render cleanly on
  FLUX.1-dev 8-bit on Apple Silicon, where they previously raised the matmul
  error above. Existing diffusers-format LoRAs continue to load unchanged.

Closes #458

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for fused LoRA QKV and QKV+MLP tensor splitting.
  * Verified projection sharing, output splitting, fused-output reconstruction, and matrix multiplication compatibility.

<h3>Greptile Summary</h3>

Adds regression coverage for fused QKV and QKV+MLP LoRA tensor splitting.
- Verifies shared down projections remain full-rank.
- Verifies up projections split along their output dimension.
- Adds rank 48 to exercise the previously uncovered three-way split condition.
- Checks reconstruction and matrix-multiplication shape compatibility.

The PR appears safe to merge.

No blocking failure remains; rank 48 now exercises the three-way split condition identified in the previous review.

| Filename | Overview |
|----------|----------|
| tests/lora/test_lora_transforms.py | Adds focused fused-LoRA splitting tests, including rank 48 coverage that resolves the prior review concern. |

</details>

<sub>Reviews (2): Last reviewed commit: ["test: cover a rank divisible by the 3-wa..."](https://github.com/mflux-community/mflux/commit/6b2397326335cccdddf1bbc82b8dd7665d65d598) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51896704)</sub>


fxd0h:
The fix is correct. A fused qkv (or qkv+mlp) LoRA factorizes as `up @ down` with `down` shaped `(rank, in)` shared across the components and `up` shaped `(fused_out, rank)`. The split is an output-dimension operation, so it applies to `up` only; `down` has to stay at full rank.

The old `rank % num_splits == 0` branch sliced `down` along its rank axis, giving each component a `down` of rank `rank/num` against an `up` of rank `rank`, so `up @ down` raises the shape error you quoted on the first denoise step. It triggers whenever the rank is divisible by the split count, so the qkv+mlp path (num=4) breaks on the common 4/8/16/32/64 and the qkv path (num=3) on multiples of 3. Same code path #423 is describing.

We pulled this exact commit into our fork (mflux-cv) and have been running it there, so it holds up outside the unit tests too. The `up @ down` shape invariant in `test_fused_down_and_up_ranks_match_after_split` is the right regression to lock in.

anthonywu (approved):
thanks @fxd0h For providing the peer review to @deadmansahil 

I recommend a merge by @filipstrand after quick look

deadmansahil:
Thank you for the production validation — knowing the commit holds up in mflux-cv outside the test suite is exactly the confirmation this needed.

plz12345:
Merged `main` into the branch to clear the conflict. Heads up on what that turned up:

The code fix here already landed upstream — #422 (LyCORIS LoKr) changed both `_split_qkv_down` and `_split_qkv_mlp_down` to `return tensor` unconditionally, which is exactly this diff. So the conflict was comment wording only, with identical code on both sides. I resolved in favour of `main`'s comments to keep the diff minimal.

That leaves this PR contributing the regression tests, which are still worth having — nothing currently pins the "down projection stays full rank" invariant, and it's an easy thing to reintroduce. I moved the file to `tests/lora/test_lora_transforms.py` to match the `tests/lora/` package `main` now uses.

Net diff against `main` is now the test file alone. `pytest tests/lora` passes (59 tests), ruff check/format clean.

plz12345:
Good catch from both bots, applied in 6b23973 — the double-block ranks (16/32/64) are all divisible by 4 but none by 3, so `_split_qkv_down` could have regressed to the old conditional split with the suite still green. Added 48, which is divisible by both, and hoisted the shared list into a `RANKS` constant so the four parametrizations stay in step.

Verified it actually bites: restoring the old rank-splitting bodies for both transforms fails 6 tests, and `test_double_block_down_is_shared_full_rank[48]` is the only one covering the 3-way path.

On CodeRabbit's linked-issues warning — expected. The loader change #458 asks for is already on `main` via #422; this branch is the regression tests for it.

PR #464, opened 2026-07-01
Training: skip non-finite steps and utf-8-safe caption reading

fxd0h:
Two small robustness fixes for the training loop, both hit during LoRA training on Apple Silicon.

**Skip non-finite steps.** A bf16 activation spike can send a single step's loss to NaN/Inf. That step then updates the LoRA weights and optimizer state with garbage, and because the plotted loss is smoothed the curve can still look healthy while the run is quietly ruined. The loop now checks the loss and skips the optimizer update when it isn't finite, continuing from the last good state, and prints how many steps were skipped. `clip_grad_norm` already handles ordinary spikes; this is for the Inf/NaN case it can't.

**UTF-8-safe caption reading.** Caption `.txt` files are read with a strict utf-8 decode, so a single stray byte raises `UnicodeDecodeError` and kills the whole run. Reading with `errors="replace"` keeps training going.

Verified on z-image: normal training is unchanged (no false skips), a non-finite loss is skipped, and a caption with invalid bytes no longer crashes.

### DIFF
--- a/src/mflux/models/common/training/state/training_spec.py
+++ b/src/mflux/models/common/training/state/training_spec.py
@@ -44,7 +44,7 @@ class DataSpec:
                 base_path=base_path,
             )
             try:
-                prompt = prompt_path.read_text(encoding="utf-8").strip()
+                prompt = prompt_path.read_text(encoding="utf-8", errors="replace").strip()
             except FileNotFoundError as e:
                 raise ValueError(f"Prompt file not found: {prompt_path}") from e
         else:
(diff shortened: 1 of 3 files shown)


anthonywu (approved):
tight fix, I would merge if you post the result of the test suite

fxd0h:
@anthonywu test output, plus the tests the PR was missing.

Suites on this merged into current main (M5 Max 40-core GPU, macOS 26.5, Python 3.12):

```
mlx 0.31.0: 517 passed, 83 deselected
mlx 0.32.0: 517 passed, 83 deselected
```

The new commit covers both halves. The non-finite check moved into `TrainingTrainer._step_is_finite` so it is directly testable: NaN, Inf and -Inf trigger the skip, finite losses do not. The caption half has a red/green pair: a latin-1 caption file fails on main with UnicodeDecodeError before the first training step, and reads with replacement characters here; a valid multibyte utf-8 caption reads back intact.

plz12345:
Merging based on comment history and passing CI.

PR #468, opened 2026-07-05
Fix Ideogram 4 stepwise preview VAE decode crash

fxd0h:
Every Ideogram 4 generation with a stepwise output dir crashes at step 0.

**Repro:** `mflux-generate-ideogram4 --model ideogram4 --prompt "..." --stepwise-image-output-dir /tmp/sw ...`

```
ValueError: [broadcast_shapes] Shapes (1,32,128,128) and (1,128,1,1) cannot be broadcast.
```

(in flux2_vae `decode_packed_latents`, at `packed_latents * bn_std + bn_mean`).

**Cause:** Ideogram 4 shares the Flux2 VAE, but `Ideogram4LatentCreator.unpack_latents` denorms and unpatchifies to a 32-channel spatial latent itself, and the final image is decoded with `vae.decode(unpacked)`, not `decode_packed_latents`. The stepwise handler sees `decode_packed_latents` on the VAE and routes Ideogram's already-unpacked 32-channel latents through the Flux2 packed path, where the 128-channel batch-norm denorm cannot broadcast against 32 channels.

**Fix:** `decode_packed_latents` now detects latents whose channel count does not match the batch-norm dim (already unpacked) and decodes them directly via `self.decode()`, which is Ideogram 4's own decode path. Flux2's packed latents (channels == bn dim) are unaffected.

Verified: a full Ideogram 4 generation with `--stepwise-image-output-dir` now completes and writes both the final image and the step previews.

### DIFF
--- a/src/mflux/models/flux2/model/flux2_vae/vae.py
+++ b/src/mflux/models/flux2/model/flux2_vae/vae.py
@@ -45,6 +45,12 @@ class Flux2VAE(nn.Module):
     def decode_packed_latents(self, packed_latents: mx.array, tiling_config: TilingConfig | None = None) -> mx.array:
         if packed_latents.ndim == 5:
             packed_latents = packed_latents[:, :, 0, :, :]
+        # Latents already fully unpacked (channel count != the patchified bn dim) come from a latent
+        # creator that denorms + unpatchifies itself — e.g. Ideogram 4, which shares this VAE but calls
+        # vae.decode() directly. The bn-denorm + unpatchify below is only for the flux2 packed form, so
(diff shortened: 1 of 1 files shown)


plz12345:
I already hit this

https://github.com/filipstrand/mflux/pull/444

fxd0h:
You're right, #444 predates this and fixes it at the call site. I checked the other `decode_packed_latents` callers (flux2 txt2img, flux2 edit, ernie) and they always pass 128-channel packed latents, so the stepwise handler is the only path that hits the 32-channel mismatch. #444 covers it fully. Closing in favor of yours.

PR #481, opened 2026-07-23
Support mlx 0.32.x

Jup33Q:
## Summary
Relax the mlx version constraint to allow mlx 0.32.x, which is required by mlx-vlm 0.6.6+.

## Changes
- `mlx>=0.27.0,<0.32.0` → `mlx>=0.27.0,<0.33.0` (darwin)
- `mlx[cuda13]>=0.30.3,<0.32.0` → `mlx[cuda13]>=0.30.3,<0.33.0` (linux)

## Verification
Tested with mlx 0.32.0 on macOS Apple Silicon (M-series):
- `Flux2Klein` model loads and generates images correctly
- No breaking changes observed

This allows users to use mflux alongside mlx-vlm and other packages that require mlx 0.32.0+.


azrahello:
I started using 0.32 a few days ago, but I noticed there are differences in results, so when moving from 0.31.2 to 0.32 all the tests should be redone.

PR #488, opened 2026-07-26
perf(seedvr2): build inverse permutation in linear time

### LINKED ISSUE #470: Perf: `_hist_match` inverts a permutation via a second `np.argsort` (O(N log N) → O(N))
### Summary

`SeedVR2Util._hist_match` inverts a permutation by running a **second full `np.argsort`**, which is O(N log N) where an O(N) scatter is exact and sufficient. On the full 2×-upscaled image (tens of millions of pixels), and called 3× per upscale (L/a/b channels), this line noticeably dominated runtime in our downstream use.

### Location

`mflux/models/seedvr2/variants/upscale/seedvr2_util.py`, `_hist_match` (lines ~231–234 in v0.18.0):

```python
src_idx = np.argsort(src, kind="stable")   # forward sort
ref_sorted = np.sort(ref, kind="stable")
inv = np.argsort(src_idx, kind="stable")   # <-- inverts the permutation via ANOTHER O(N log N) sort
out[i] = ref_sorted[inv].reshape(source.shape[1:]).astype(np.float32)
```

`inv = np.argsort(src_idx, ...)` computes the inverse of the permutation `src_idx`. Since `src_idx` is a permutation of `range(N)` (unique indices, no ties), its inverse can be built directly with an O(N) scatter — no comparison sort needed.

### Why it's slow

- `apply_color_correction` → `_lab_color_transfer_exact` calls `_hist_match` 3× (L, a, b channels) per SeedVR2 upscale.
- It runs on the full, already-2×-upscaled image, so N is tens of millions of pixels per channel.
- The inverse is computed by sorting an int64 index array (O(N log N)) instead of scattering (O(N)).

### Measurement

Isolating just the inverse-permutation step on 10M elements:

- Double-`argsort` (current): **~6.6 s**
- O(N) scatter (proposed): **~0.05 s**

≈ **125×** faster on the isolated step. In our application this was the difference between a snappy upscale and multi-minute stalls / apparent hangs on large photos.

### Proposed fix (drop-in, output-identical)

```python
inv = np.empty_like(src_idx)
inv[src_idx] = np.arange(src_idx.size, dtype=src_idx.dtype)
```

This produces the exact inverse permutation (indices are unique, no ties), so the output is mathematically identical — only the complexity changes from O(N log N) to O(N).

Optional secondary note: the two forward `kind="stable"` sorts (`np.argsort(src, ...)` / `np.sort(ref, ...)`) could use the default (quicksort) since histogram matching on continuous pixel values doesn't require a stable tie-break — a smaller additional win. The inverse-permutation change above is the main one.

### Version affected

mflux `0.18.0` (installed from PyPI). Happy to open a PR if that's helpful.

Missing-Identity:
## Summary

- replace the redundant inverse `argsort` in SeedVR2 histogram matching with a scatter-built inverse permutation
- add a regression test covering batched inputs with repeated source values

Fixes #470.

## Validation

- `ruff check src/mflux/models/seedvr2/variants/upscale/seedvr2_util.py tests/image_generation/test_seedvr2_util.py`
- `ruff format --check src/mflux/models/seedvr2/variants/upscale/seedvr2_util.py tests/image_generation/test_seedvr2_util.py`
- `python -m compileall -q src/mflux/models/seedvr2/variants/upscale/seedvr2_util.py tests/image_generation/test_seedvr2_util.py`
- exact equivalence of the scatter inverse and the previous stable `argsort` inverse for 1, 2, 7, and 200,000 elements

The focused pytest test requires a usable MLX runtime. It is skipped in the Linux validation runner because no CUDA driver is available.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved histogram matching consistency for images containing repeated pixel values.
  * Ensured stable value ordering produces the expected upscale results.

* **Tests**
  * Added coverage validating histogram mapping with repeated values.

<h3>Greptile Summary</h3>

This PR replaces a redundant stable sort with a linear-time scatter operation when constructing SeedVR2's inverse histogram permutation.
- Builds the inverse permutation by assigning each sorted position its rank.
- Adds a repeated-value, batched regression test comparing the result with the previous implementation.

<h3>Confidence Score: 4/5</h3>

The PR appears safe to merge, with the non-blocking limitation that its regression test does not run in the documented Linux validation environment.

The optimized scatter operation preserves the previous inverse-permutation semantics, while the only accepted concern is reduced automated coverage caused by gating a NumPy-only test on MLX availability.

**Files Needing Attention:** tests/image_generation/test_seedvr2_util.py

| Filename | Overview |
|----------|----------|
| src/mflux/models/seedvr2/variants/upscale/seedvr2_util.py | The scatter assignment correctly constructs the inverse of the complete permutation returned by stable `argsort`. |
| tests/image_generation/test_seedvr2_util.py | The regression case checks exact equivalence, but its module-level MLX gate prevents it from running in the documented Linux validation environment. |

</details>

<details><summary>Prompt To Fix All With AI</summary>

`````markdown
### Issue 1
tests/image_generation/test_seedvr2_util.py:4
**MLX gate suppresses regression coverage**

The module-level `mlx.core` gate skips this NumPy-only regression test in the documented Linux validation environment, so changes to the inverse-permutation behavior receive no automated coverage there.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["Speed up SeedVR2 histogram matching"](https://github.com/mflux-community/mflux/commit/34b73f8d484bb01c077dafb76fdd315488fa6ecb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51896689)</sub>

> Greptile also left **1 inline comment** on this PR.

### DIFF
--- a/src/mflux/models/seedvr2/variants/upscale/seedvr2_util.py
+++ b/src/mflux/models/seedvr2/variants/upscale/seedvr2_util.py
@@ -230,6 +230,7 @@ class SeedVR2Util:
             ref = reference[i].reshape(-1).astype(np.float32)
             src_idx = np.argsort(src, kind="stable")
             ref_sorted = np.sort(ref, kind="stable")
-            inv = np.argsort(src_idx, kind="stable")
+            inv = np.empty_like(src_idx)
+            inv[src_idx] = np.arange(src_idx.size, dtype=src_idx.dtype)
             out[i] = ref_sorted[inv].reshape(source.shape[1:]).astype(np.float32)
         return out
--- a/tests/image_generation/test_seedvr2_util.py
+++ b/tests/image_generation/test_seedvr2_util.py
@@ -1,0 +1,36 @@
+import numpy as np
+import pytest
+
+pytest.importorskip("mlx.core", exc_type=ImportError)
+
+from mflux.models.seedvr2.variants.upscale.seedvr2_util import SeedVR2Util
+
+
+@pytest.mark.fast
+def test_hist_match_matches_argsort_inverse_with_repeated_values():
+    source = np.array(
+        [
+            [[0.5, 0.1, 0.5], [0.3, 0.1, 0.9]],
+            [[0.2, 0.2, 0.8], [0.4, 0.6, 0.4]],
+        ],
+        dtype=np.float32,
+    )
+    reference = np.array(
+        [
+            [[0.9, 0.7, 0.6], [0.4, 0.3, 0.1]],
+            [[0.8, 0.7, 0.5], [0.3, 0.2, 0.1]],
+        ],
+        dtype=np.float32,
+    )
+
+    expected = np.empty_like(source, dtype=np.float32)
+    for index in range(source.shape[0]):
+        src = source[index].reshape(-1)
+        src_idx = np.argsort(src, kind="stable")
+        inverse = np.argsort(src_idx, kind="stable")
+        reference_sorted = np.sort(reference[index].reshape(-1), kind="stable")
+        expected[index] = reference_sorted[inverse].reshape(source.shape[1:])
+
+    actual = SeedVR2Util._hist_match(source, reference)
+
+    np.testing.assert_array_equal(actual, expected)

### CODE REVIEW

**Comment by fxd0h** [comment]
Benchmarked this before pulling it into the `mflux-cv` community build, where it is now merged, so here are numbers for whoever reviews it.

**Equivalence first**, since a rewrite of an inverse permutation is exactly where a subtle difference hides. `argsort(argsort(x))` and the scatter agree elementwise at n=1000 and n=200000 with deliberately repeated values, which is the case where the stable-sort semantics could have diverged. Not just on random floats where every value is unique.

**Speed**, at 4M elements, roughly one channel of a 1080p frame, on an M5 Max:

| | time |
|---|---|
| `np.argsort(src_idx, kind="stable")` | 880 ms |
| `inv[src_idx] = arange(...)` | 17.3 ms |
| | **51x** |

One note for expectation-setting: issue #470 quotes 125x, and I measure 51x here. Different hardware and array size will move that number around, but the shape of the win is the same and it is the difference between a visible stall per channel and nothing.

The test you added passes, and the full fast suite is green with it (641 passed). Nothing to change from my side, this is just an outside confirmation that it does what it says.

**Comment by plz12345** [comment]
Cherry-picked this onto my local branch and gave it a run, since I use SeedVR2 heavily. LGTM — correct, and a real if modest win.

**Correctness.** `src_idx` is `argsort` output, so it's always a permutation of `0..N-1` with no ties, and the scatter builds exactly its inverse. Every slot of the `np.empty_like` gets written, dtype carries through from `src_idx.dtype`, and N=0 degenerates cleanly. Output is bit-for-bit identical to the old path, not just numerically close — I verified with `array_equal` at every size below, not `allclose`.

**Benchmarks.** Apple M5 Pro, numpy 2.4.1, float32, single channel (the real call does 3x this for L/a/b):

| size | before | after | speedup |
|---|---|---|---|
| 1 MP | 223 ms | 140 ms | 1.60x |
| 4 MP | 1375 ms | 657 ms | 2.09x |
| 16 MP | 9042 ms | 3069 ms | 2.95x |

One flag, meant constructively: I can't reproduce anything close to the 51x reported earlier in this thread, and I don't think it's reproducible in principle. This removes one of two O(N log N) sorts — the ceiling is ~2x, plus a bit more at large N because the second `argsort` has a worse access pattern than the scatter. 1.6–3x is what theory predicts and what I measure. That 51x is probably timing something other than a full `_hist_match` call. Doesn't change the verdict at all, just worth not carrying the number into the merge note.

**Optional follow-on.** The `inv` array itself is avoidable — `ref_sorted[inv]` where `inv` is the inverse permutation is by definition `out[src_idx] = ref_sorted`:

```python
src_idx = np.argsort(src, kind="stable")
ref_sorted = np.sort(ref, kind="stable")
flat = np.empty_like(ref_sorted)
flat[src_idx] = ref_sorted
out[i] = flat.reshape(source.shape[1:]).astype(np.float32)
```

That drops two N-sized int64 temporaries (~128 MB per channel on a 2x-upscaled 8 MP image) and one random-access gather pass. Benchmarks 3.16x vs 2.95x at 16 MP, still bit-identical, and the test in this PR passes against it unchanged. Entirely up to you whether it belongs here or in a follow-up — this PR stands on its own.

**Tests.** The new test passes. It lives in `tests/image_generation/` but is pure numpy and downloads nothing, which is nice. Full offline suite on my branch shows no new failures. Note the `pytest.importorskip("mlx.core")` guard can never actually skip — mlx is a hard dependency on both darwin and linux — so it's dead weight, but it's harmless and ruff doesn't flag the import that follows it (I checked E402 explicitly against both the pinned 0.12.1 and current 0.16.2).

PR #489, opened 2026-07-29
fix: bump mlx to 0.32.0 for corrupt quantized matmul above 32768 rows

plz12345:
## The bug

`mlx <0.32.0` silently returns garbage for the first `M - 32768` rows of
`mx.quantized_matmul` whenever `M > 32768`. Nothing is raised — the call just
returns wrong data.

Measured on darwin/Metal against a chunked reference (same op, 2048 rows at a
time), mlx 0.31.0 and 0.31.2:

| M | bad rows |
|---|---|
| 32768 | 32 `[0..31]` |
| 34992 | 2240 `[0..2239]` |
| 62208 | 29472 `[0..29471]` |

The bad rows are always a **prefix** — roughly `[0 .. M-32768)` rounded up to a
multiple of 32. It affects 4/6/8 bits, fp16 and fp32 inputs, and both
`nn.QuantizedLinear` and raw `mx.quantized_matmul`. **Unquantized matmul is
unaffected at any M.** mlx 0.32.0 returns 0 bad rows at all three sizes.

```python
import mlx.core as mx, numpy as np
w = mx.random.normal((4096, 3072))
wq, s, bz = mx.quantize(w, group_size=64, bits=8)
x = mx.random.normal((34992, 3072)).astype(mx.float16)
kw = dict(transpose=True, group_size=64, bits=8)
full = mx.quantized_matmul(x, wq, s, bz, **kw)
ref = mx.concatenate([mx.quantized_matmul(x[i:i+2048], wq, s, bz, **kw)
                      for i in range(0, 34992, 2048)])
d = np.abs(np.array(full.astype(mx.float32)) - np.array(ref.astype(mx.float32))).mean(axis=1)
print(np.where(d > 1e-3)[0])   # 0.31.x: [0 ... 2239];  0.32.0: []
```

## How it surfaces in mflux

The transformer's projections run over all tokens in row-major order, so a
corrupted row-prefix is a contiguous region at the *start* of the token
sequence — which is the top of the image. Tokens are `H*W/256`, so the ceiling
is **~8.4 MP of output** for any quantized model.

SeedVR2 is where this actually bites, since upscaling reaches those sizes
routinely. Upscaling a 1152x864 image with `seedvr2-7b -q 8`:

| | output | tokens | corrupted |
|---|---|---|---|
| 2x | 2304x1728 | 15,552 | none |
| 3x | 3456x2592 | 34,992 | top 6.4% |
| 4x | 4608x3456 | 62,208 | top 47% |

At 3x the affected strip is the top ~160px: the transformer emits ~0 there, so
the initial noise passes through undenoised and it reads as a band of sparkle
with a hard lower edge. Reported downstream as a suspected VAE-tiling or
window-attention artifact; it is neither. Ruled out by testing: VAE
encode/decode tiling (round-trip error profile is smooth and content-driven),
window-shift geometry (disabling shift changes nothing), and window/RoPE
layout (identical at 1x/2x/3x — window size is a constant 18x23 tokens at every
scale). The same run unquantized is clean at every size.

## Verification

Same input, same seed, `-q 8`, 3x from 1152x864, only mlx changed:

| | HF energy rows 0-160 | rows 161-400 | step at band edge |
|---|---|---|---|
| mlx 0.31.0 | 8.86 | 4.99 | **1.78x** |
| mlx 0.32.0 | 4.88 | 4.96 | 0.98x |

4x (15.9 MP) also confirmed clean on 0.32.0.

## Note on the linux pin

Left at `<0.32.0` — `mlx-cuda` tops out at 0.30.0, so it can't follow. The CUDA
backend is a separate kernel path and was not tested; it may or may not carry
the same bug.

## Summary by CodeRabbit

* **Chores**
  * Updated the macOS runtime to use the newer MLX 0.32 release.
  * Updated the macOS development environment to match MLX 0.32.
  * Improved compatibility by avoiding an issue affecting quantized matrix multiplication in earlier versions.

<h3>Greptile Summary</h3>

This PR raises the Darwin MLX requirement and development pin to 0.32.0 to avoid corrupt output from large quantized matrix multiplications.
- Updates the macOS production constraint to `mlx>=0.32.0,<0.33.0`.
- Pins macOS development environments to MLX 0.32.0 while leaving the separate Linux CUDA dependency unchanged.
- Regenerates the lock file with matching MLX and MLX Metal 0.32.0 artifacts.

The PR appears safe to merge because the Darwin dependency constraints and lock records consistently select MLX 0.32.0 without introducing a supported-target resolution conflict.

The changed manifest and lock data agree on the platform-specific versions, provide matching macOS arm64 artifacts for supported Python versions, preserve the separate Linux dependency path, and introduce no new transitive dependencies.

| Filename | Overview |
|----------|----------|
| pyproject.toml | Consistently updates the Darwin runtime and development MLX constraints while preserving the intentional Linux CUDA pin. |
| uv.lock | Consistently locks MLX and MLX Metal 0.32.0 for Darwin across supported Python versions without changing unrelated dependency versions. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix: bump mlx to 0.32.0 for corrupt quan..."](https://github.com/mflux-community/mflux/commit/a9704fa1b0b8b58def97aed3bf14c1186ba18c66) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51896687)</sub>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,10 @@ dependencies = [
     "huggingface-hub>=1.1.6,<2.0",
     "hf-transfer>=0.1.9,<1.0",
     "matplotlib>=3.9.2,<4.0",
-    "mlx>=0.27.0,<0.32.0; sys_platform=='darwin'",
+    # mlx <0.32.0 has a quantized_matmul bug: for M>32768 rows the first
+    # M-32768 rows come back garbage (silently, no error). Quantized SeedVR2
+    # above ~8.4MP output hits it as a noise band across the top of the image.
+    "mlx>=0.32.0,<0.33.0; sys_platform=='darwin'",
     "mlx[cuda13]>=0.30.3,<0.32.0; sys_platform=='linux'",
     "numpy>=2.0.1,<3.0",
     "opencv-python>=4.10.0,<5.0",
@@ -66,7 +69,7 @@ dev = [
     "matplotlib>3.10,<4.0",
     "pytest>=8.3.0,<9.0",
     "pytest-timer>=1.0,<2.0",
-    "mlx==0.31.0; sys_platform=='darwin'",
+    "mlx==0.32.0; sys_platform=='darwin'",
     "mlx[cuda13]==0.30.3; sys_platform=='linux'",
 ]
 
--- a/uv.lock
+++ b/uv.lock
@@ -1066,7 +1066,7 @@ dependencies = [
     { name = "huggingface-hub" },
     { name = "matplotlib" },
     { name = "mlx", version = "0.30.3", source = { registry = "https://pypi.org/simple" }, extra = ["cuda13"], marker = "sys_platform == 'linux'" },
-    { name = "mlx", version = "0.31.0", source = { registry = "https://pypi.org/simple" }, marker = "sys_platform == 'darwin'" },
+    { name = "mlx", version = "0.32.0", source = { registry = "https://pypi.org/simple" }, marker = "sys_platform == 'darwin'" },
     { name = "numpy", version = "2.2.6", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.11'" },
     { name = "numpy", version = "2.4.1", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.11'" },
     { name = "opencv-python" },
@@ -1091,7 +1091,7 @@ dependencies = [
 dev = [
     { name = "matplotlib" },
     { name = "mlx", version = "0.30.3", source = { registry = "https://pypi.org/simple" }, extra = ["cuda13"], marker = "sys_platform == 'linux'" },
-    { name = "mlx", version = "0.31.0", source = { registry = "https://pypi.org/simple" }, marker = "sys_platform == 'darwin'" },
+    { name = "mlx", version = "0.32.0", source = { registry = "https://pypi.org/simple" }, marker = "sys_platform == 'darwin'" },
     { name = "pytest" },
     { name = "pytest-timer" },
 ]
@@ -1104,8 +1104,8 @@ requires-dist = [
     { name = "huggingface-hub", specifier = ">=1.1.6,<2.0" },
     { name = "matplotlib", specifier = ">=3.9.2,<4.0" },
     { name = "matplotlib", marker = "extra == 'dev'", specifier = ">3.10,<4.0" },
-    { name = "mlx", marker = "sys_platform == 'darwin'", specifier = ">=0.27.0,<0.32.0" },
-    { name = "mlx", marker = "sys_platform == 'darwin' and extra == 'dev'", specifier = "==0.31.0" },
+    { name = "mlx", marker = "sys_platform == 'darwin'", specifier = ">=0.32.0,<0.33.0" },
+    { name = "mlx", marker = "sys_platform == 'darwin' and extra == 'dev'", specifier = "==0.32.0" },
     { name = "mlx", extras = ["cuda13"], marker = "sys_platform == 'linux'", specifier = ">=0.30.3,<0.32.0" },
     { name = "mlx", extras = ["cuda13"], marker = "sys_platform == 'linux' and extra == 'dev'", specifier = "==0.30.3" },
     { name = "numpy", specifier = ">=2.0.1,<3.0" },
@@ -1161,7 +1161,7 @@ cuda13 = [
 
 [[package]]
 name = "mlx"
-version = "0.31.0"
+version = "0.32.0"
 source = { registry = "https://pypi.org/simple" }
 resolution-markers = [
     "python_full_version >= '3.13' and sys_platform == 'darwin'",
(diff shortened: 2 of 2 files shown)


anthonywu:
nominated myself was reviewer, there is discussion of why/how outstanding PRs are not being merged, I'll give the feedback on this one:

@plz12345 if you show me that you've run the unit test matrix on the set of { supported python version } X { recent MLX versions } I would approve this PR instantly.

Without showing the work and output of your test runs puts the burden on the maintainer to do so.

We could solve this with a proper GitHub Actions runner, but that is not a free resource and the project lacks a financial sponsor.

plz12345:
Github runners are free, but you're not going to be able to do real validation on a limited resource runner.  Self-hosted would work, but someone would need to light that up and keep it online.  That said, understood.

We're waiting for Filip to respond on transferring stewardship of this repo, so it can move to mflux-community.  I posted this only because I hit the issue, and realized Apple had already solved it in mlx 0.32, but mflux is pinned to 0.31.

That said, here is the output of the fast test suite.  Note, this change is purely a minor version bump of the core mlx package, we'd be doing this eventually anyways.  EDIT: Just realized these test alredy ran, so not even sure why you asked for them/proof.  Github action already ran it and passed.  It's impractical to expect full image gen tests on a change like this.

```
Branch checkout — verified against the PR head:

  branch: fix/mlx-0.32-quantized-matmul
  sha:    a9704fa1b0b8b58def97aed3bf14c1186ba18c66
  PR #489 head sha: a9704fa1b0b8b58def97aed3bf14c1186ba18c66   <- identical

  Working tree clean. HEAD = a9704fa "fix: bump mlx to 0.32.0 for corrupt
  quantized matmul above 32768 rows", sitting directly on upstream
  97ac5e6 "feat: add Krea 2 text-to-image model (#453)".

Command (the repo's own Makefile target, which reinstalls .[dev] to pull the
pinned MLX before running):

  make test-fast     # = MFLUX_PRESERVE_TEST_OUTPUT=1 uv run python -m pytest -m fast

Environment:

  platform darwin -- Python 3.11.13, pytest-8.4.2, pluggy-1.6.0 -- /Users/paul/Git/mflux/.venv/bin/python3
  rootdir: /Users/paul/Git/mflux
  plugins: anyio-4.12.1, timer-1.0.0
  mlx 0.32.0          <- the version this PR pins, resolved and installed cleanly
  macOS-26.5.2-arm64

Result:

  collected 593 items / 83 deselected / 510 selected
  ================ 507 passed, 3 skipped, 83 deselected in 4.22s =================

  Zero failures, zero errors.

Accounting for the non-passes, so the numbers aren't read as hidden failures:

  3 skipped   - all in tests/image_generation/test_ideogram4_checkpoint_layout.py;
                they self-skip unless a local Ideogram 4 checkpoint is cached
                (pytest.skip("Cache {REPO_ID} or set {MODEL_PATH_ENV} ...")).
                Unrelated to this PR.

  83 deselected - the "-m fast" filter excluding the slow (image-generating)
                tests, plus "high_memory_requirement" which pyproject.toml's
                addopts deselects by default.
```

anthonywu (approved):
thanks, I am stamping with my approval but I want to give @filipstrand a chance to screen all the PRs too

perhaps he's on a vacation and we can allow a few days, i'm sure I already landed 20+ notifications in his inbox =)

fxd0h:
Independent confirmation. I reproduced this synthetically while porting the floor bump to mflux-cv: M=40000, 4 bit quantized weights, comparing quantized_matmul against the dequantized reference.

```
mlx 0.31.0: max abs err 133.9 in rows 0..7231 and 81.5 in rows 7232..39999
mlx 0.32.0: max abs err 0.0 everywhere
```

One nuance for the description: in this repro the corruption is not limited to the first M-32768 rows, the whole result comes back garbage. Even more reason for the bump.

Shipped the same floor in mflux-cv 0.18.37.

fxd0h:
@coderabbitai full review

fxd0h:
@greptileai review

fxd0h (approved):
Reproduced the corruption on an M5 Max (macOS 26.5, Metal) with your probe extended to all three sizes: mlx 0.31.0 gives the same bad-row counts as your table, 32 / 2240 / 29472, always a prefix, and 0.32.0 gives zero at all three. mflux-cv has shipped this same pin (`>=0.32.0,<0.33.0` darwin, dev `==0.32.0`) since the 7th with no fallout.

The 0.32 risk you flagged in #422, `test_lokr_matmul_matches_materialized_delta` going red on a tight tolerance, is already resolved on main: the merged test carries `rtol=1e-2`. Current main's fast suite under mlx 0.32.0: 515 passed, and the lora layer file passes whole.

On CodeRabbit's inline suggestion to reword the pyproject comment: the prefix shape it wants softened is what both your table and this repro measure, three sizes, two machines, always `[0..N)`. I'd keep your wording.

fxd0h (approved):
Reproduced the corruption on an M5 Max (macOS 26.5, Metal) with your probe extended to all three sizes: mlx 0.31.0 gives the same bad-row counts as your table, 32 / 2240 / 29472, always a prefix, and 0.32.0 gives zero at all three. mflux-cv has shipped this same pin (`>=0.32.0,<0.33.0` darwin, dev `==0.32.0`) since the 7th with no fallout.

The 0.32 risk you flagged in #422, `test_lokr_matmul_matches_materialized_delta` going red on a tight tolerance, is already resolved on main: the merged test carries `rtol=1e-2`. Current main's fast suite under mlx 0.32.0: 515 passed, and the lora layer file passes whole.

On CodeRabbit's inline suggestion to reword the pyproject comment: the prefix shape it wants softened is what both your table and this repro measure, three sizes, two machines, always `[0..N)`. I'd keep your wording.

fxd0h:
Merging this one: two maintainer approvals and the same green board, per the process sketched in #513. Repro details in the review above.

PR #492, opened 2026-07-30
fix: flux2 CLIs discard all image metadata

plz12345:
Both flux2 entry points call `ImageUtil.save_image(image:path:export_json_metadata:)` without passing `metadata=`. It defaults to `None`, so the `--metadata` sidecar is written as the literal `null`.

Every other CLI in the tree calls `image.save(path:export_json_metadata:)`, which routes through `GeneratedImage._get_metadata()`. flux2 is the only outlier — this just brings it in line.

Reproduce on `main`:

```
mflux-generate-flux2 --prompt "a puffin on a cliff" --steps 4 --metadata -o out.png
cat out.metadata.json      # -> null
```

After: the sidecar carries the usual model/seed/steps/guidance/dimensions/prompt.

Two files, +2/−12 (the `ImageUtil` import becomes unused in both). No behaviour change beyond the sidecar being written.

Noticed while reviewing #490 — flux2 was the one model where a `--pid-decode` run could not be recorded, but the bug predates that work and is not specific to it, so it's split out here rather than carried in a decoder PR.

---

**Correction to an earlier version of this description**, thanks to @azrahello: I originally wrote that no EXIF or XMP was embedded either. That is wrong. `save_image` is handed a `GeneratedImage` rather than a PIL image, so `image.save(file_path)` dispatches to `GeneratedImage.save`, which routes back through `_get_metadata()` and embeds correctly — the outer call then writes the `null` sidecar on top. Verified:

```
sidecar contents : null
embedded exif    : prompt, seed, steps, model, width, height, guidance, quantize, ...
```

So the sidecar is the broken half, not the whole of it. The impact is unchanged, since `--config-from-metadata` reads the sidecar.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved image saving for generated and edited outputs.
  - Preserved configured output paths and metadata options when saving images.
  - Removed an issue caused by reliance on an outdated image-saving method.

<h3>Greptile Summary</h3>

The PR fixes Flux2 metadata sidecar exports by routing generated images through `GeneratedImage.save`, which supplies the complete generated-image metadata.
- Updates the Flux2 text-to-image CLI to preserve metadata in requested JSON sidecars.
- Applies the same correction to the Flux2 edit CLI.
- Removes unused `ImageUtil` imports from both entry points.

The PR appears safe to merge, with both Flux2 entry points now following the established generated-image save contract.

The changed calls operate on `GeneratedImage` instances and preserve existing output-path and sidecar-naming behavior while supplying populated metadata instead of writing a null sidecar.

| Filename | Overview |
|----------|----------|
| src/mflux/models/flux2/cli/flux2_generate.py | Replaces the metadata-dropping utility call with the established `GeneratedImage.save` path; no actionable regression found. |
| src/mflux/models/flux2/cli/flux2_edit_generate.py | Applies the same correct save path to edited Flux2 images and removes the unused utility import. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix: flux2 CLIs discarded all image meta..."](https://github.com/mflux-community/mflux/commit/6ca8808b25d170c48d4ab232bca2be7badc4b0d4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51896685)</sub>

### DIFF
--- a/src/mflux/models/flux2/cli/flux2_edit_generate.py
+++ b/src/mflux/models/flux2/cli/flux2_edit_generate.py
@@ -7,7 +7,6 @@ from mflux.models.flux2.latent_creator.flux2_latent_creator import Flux2LatentCr
 from mflux.models.flux2.variants import Flux2KleinEdit
 from mflux.utils.dimension_resolver import DimensionResolver
 from mflux.utils.exceptions import PromptFileReadError, StopImageGenerationException
-from mflux.utils.image_util import ImageUtil
 from mflux.utils.prompt_util import PromptUtil
 
 
@@ -73,11 +72,7 @@ def main():
                 num_inference_steps=args.steps,
                 scheduler="flow_match_euler_discrete",
             )
-            ImageUtil.save_image(
-                image=image,
-                path=args.output.format(seed=seed),
-                export_json_metadata=args.metadata,
-            )
+            image.save(path=args.output.format(seed=seed), export_json_metadata=args.metadata)
     except (StopImageGenerationException, PromptFileReadError) as exc:
         print(exc)
     finally:
--- a/src/mflux/models/flux2/cli/flux2_generate.py
+++ b/src/mflux/models/flux2/cli/flux2_generate.py
@@ -5,7 +5,6 @@ from mflux.models.flux2.latent_creator.flux2_latent_creator import Flux2LatentCr
 from mflux.models.flux2.variants import Flux2Klein
 from mflux.utils.dimension_resolver import DimensionResolver
 from mflux.utils.exceptions import PromptFileReadError, StopImageGenerationException
-from mflux.utils.image_util import ImageUtil
 from mflux.utils.prompt_util import PromptUtil
 
 
@@ -65,11 +64,7 @@ def main():
                 image_strength=args.image_strength,
                 scheduler="flow_match_euler_discrete",
             )
-            ImageUtil.save_image(
-                image=image,
-                path=args.output.format(seed=seed),
-                export_json_metadata=args.metadata,
-            )
+            image.save(path=args.output.format(seed=seed), export_json_metadata=args.metadata)
     except (StopImageGenerationException, PromptFileReadError) as exc:
         print(exc)
     finally:


azrahello:
Verified by inspection on `main`: `flux2_generate.py` and `flux2_edit_generate.py` are the only two CLIs calling `ImageUtil.save_image(...)` without `metadata=` — every other entry point goes through `image.save(...)`, and so through `GeneratedImage._get_metadata()`. I don't have the FLUX.2 weights on this machine to run your repro, so that is a code-path confirmation rather than an output one.

Also confirming the downstream effect: on #490 the PiD flags are recorded for six of seven models, and flux2 is the exception until this merges. It is now declared there as a dependency rather than left implicit.

plz12345:
Yeah, this should go before 490 also because it's a 2-line change.

azrahello:
Ran your repro after all — the 9B was in cache and I said otherwise without checking first. Apologies for that. Sidecar comes out `null`, confirmed.

One detail for the description, since it cost me a double-take: the PNG itself does still get embedded XMP/EXIF/IPTC. `save_image` is handed a `GeneratedImage` rather than a PIL image, so `image.save(file_path)` dispatches to `GeneratedImage.save`, which routes back through `_get_metadata()` and writes them — and only then does the outer call put the `null` sidecar on top. Your point stands either way, since `--config-from-metadata` json-loads the sidecar, it's just that the sidecar is the broken half rather than the whole of it.

Agreed this should go first; #490 declares the dependency.

plz12345:
You're right, and thanks for checking it properly — I've corrected the description.

I'd asserted "no EXIF or XMP is embedded at all" from reading `save_image` and assuming `image` was a PIL image. It isn't, and your dispatch trace is exactly what happens. Verified it directly rather than taking either of our readings on trust — constructing a `GeneratedImage` and calling `save_image(image:path:export_json_metadata:)` the way flux2's CLI did:

```
sidecar contents : null
embedded exif    : ['base_model', 'created_at', 'generation_time_seconds', 'guidance',
                    'height', 'lora_paths', 'model', 'negative_prompt', 'precision',
                    'prompt', 'quantize', 'seed', 'steps', 'width', ...]
  prompt: a puffin on a cliff
  seed: 42
  steps: 4
  model: black-forest-labs/FLUX.2-klein-9B
```

Full metadata in the EXIF, `null` in the sidecar. So the blast radius is smaller than I claimed: it's the sidecar only, which happens to be the half `--config-from-metadata` reads, so the user-visible consequence is the same.

Worth noting the fix doesn't change the embedded side at all — `image.save(...)` reaches the same `_get_metadata()`, just without the outer call stamping `null` over the sidecar afterwards.

No apology needed on the cache thing; you checked before I did.

fxd0h:
@coderabbitai full review

fxd0h:
@greptileai review

fxd0h (approved):
Confirmed at the call level: `ImageUtil.save_image` with the default `metadata=None` writes the sidecar as `json.dump(None)`, the literal `null` you show, and it also skips the `metadata is not None` branch that embeds EXIF and IPTC into the PNG itself. So the fix does slightly more than the description says, in a good way: flux2 outputs also gain embedded metadata, same as every other CLI, since `GeneratedImage.save` passes `_get_metadata()` through. Checked the tree for remaining `ImageUtil.save_image` callers: outside these two, only the canny/depth savers and the concept-attention heatmaps use it, auxiliary images where generation metadata does not apply, so flux2's entry points were the last generate CLIs on that path.

mflux-cv's flux2 CLIs have shipped this same shape with no issues. LGTM.

fxd0h (approved):
Confirmed at the call level: `ImageUtil.save_image` with the default `metadata=None` writes the sidecar as `json.dump(None)`, the literal `null` you show, and it also skips the `metadata is not None` branch that embeds EXIF and IPTC into the PNG itself. So the fix does slightly more than the description says, in a good way: flux2 outputs also gain embedded metadata, same as every other CLI, since `GeneratedImage.save` passes `_get_metadata()` through. Checked the tree for remaining `ImageUtil.save_image` callers: outside these two, only the canny/depth savers and the concept-attention heatmaps use it, auxiliary images where generation metadata does not apply, so flux2's entry points were the last generate CLIs on that path.

mflux-cv's flux2 CLIs have shipped this same shape with no issues. LGTM.

fxd0h:
Merging: maintainer approval and green CI across fast-tests and both review bots, per the process sketched in #513. Verification in the review above.

PR #496, opened 2026-08-01
Define the 1-step schedule in FlowMatchEulerDiscreteScheduler

### LINKED ISSUE #494: --steps 1 raises ZeroDivisionError in shared FlowMatchEulerDiscreteScheduler (not a one-line guard)
### Summary

`--steps 1` raises `ZeroDivisionError` in `FlowMatchEulerDiscreteScheduler`.
This is in the **shared** scheduler (`mflux/models/common/schedulers/`), not in
any one model's code, so it affects every model family that selects
`flow_match_euler_discrete`.

It is **not** a one-line guard — the obvious fix moves the failure one layer
down. Details below, since that seemed worth saving you the round trip.

### Reproduction

```bash
mflux-generate-flux2 --prompt "x" --quantize 8 \
  --height 512 --width 512 --steps 1 --seed 7 --output out.png
```

```
File "mflux/models/common/schedulers/flow_match_euler_discrete_scheduler.py", line 110,
      in _compute_timesteps_and_sigmas
    - i * (sigma_max - sigma_min) * self.num_train_timesteps / (num_steps - 1)
ZeroDivisionError: float division by zero
```

Reproducible with no model or weights at all:

```python
from types import SimpleNamespace
from mflux.models.common.schedulers.flow_match_euler_discrete_scheduler import (
    FlowMatchEulerDiscreteScheduler as S,
)

for n in (1, 2, 4):
    cfg = SimpleNamespace(num_inference_steps=n, model_config=SimpleNamespace())
    S(cfg)   # n=1 -> ZeroDivisionError; n=2 and n=4 fine
```

### Layer 1

`_compute_timesteps_and_sigmas()` spreads `i` over `num_steps - 1`, which is 0
when `num_steps == 1`.

### Layer 2 — why the obvious guard is not enough

Changing the denominator to `max(num_steps - 1, 1)` gets past line 110 and then
fails in `_stretch_to_terminal()`:

```
File ".../flow_match_euler_discrete_scheduler.py", line 101, in _stretch_to_terminal
    stretched = [1.0 - (oms / scale_factor) for oms in one_minus_sigmas]
ZeroDivisionError: float division by zero
```

The `n == 1` schedule degenerates:

```
timesteps_linear = [1000.0]        # only i=0 exists
sigmas_linear    = [1.0]
sigmas_shifted   = [1.0]           # exp(1)/(exp(1) + (1/1 - 1)^1) = 1.0
one_minus_sigmas = [0.0]
scale_factor     = 0.0 / (1.0 - shift_terminal) = 0.0   # then divided by
```

So with a single step there is no terminal to stretch toward — `shift_terminal`
has nothing to act on. That is a semantics question about what a 1-step
flow-match schedule *should* be, which is why I am not proposing a patch.

### A reference point

The sibling static method in the same class, `get_timesteps_and_sigmas()` (used
by `set_image_seq_len`), handles `n == 1` without trouble:

```python
sigmas = mx.linspace(1.0, 1.0 / num_inference_steps, num_inference_steps, dtype=mx.float32)
```

`mx.linspace(1.0, 1.0, 1)` → `[1.0]`, and that path never calls
`_stretch_to_terminal`. If the two paths are meant to agree, that difference may
be the more interesting bug.

### Scope

- `flow_match_euler_discrete_scheduler.py:110` is the **only** `(num_steps - 1)`
  division among the four schedulers in `models/common/schedulers/`.
  `linear_scheduler.py` and `seedvr2_euler_scheduler.py` do not have it — which
  is why `mflux-generate-z-image-turbo --steps 1` is unaffected (turbo selects
  `linear`).
- By grep, the affected scheduler is selected by:
  - **flux2** — klein txt2img, edit, training adapter
  - **fibo** — txt2img, edit
  - **z_image** — its guidance-supporting path
- Verification honesty: I reproduced this at runtime only through
  `mflux-generate-flux2`, plus the standalone snippet above.
  I did not download fibo or non-turbo z-image, so treat those two as
  static analysis rather than confirmed.

### Impact

Low severity — `--steps 2` works and 1 step is not a useful quality setting.
Filing because the failure is a bare `ZeroDivisionError` from library internals
rather than a validation message, so it reads as a crash rather than "that value
is out of range." A parser-level minimum for schedulers that need ≥2 steps would
also be a reasonable resolution.

### Environment

- mflux 0.18.0 (PyPI), Python 3.12.13, mlx 0.32.0
- macOS 26.5.2, Apple M1 Pro, 32 GB
- Model: `black-forest-labs/FLUX.2-klein-4B`

Split out from #493 as suggested there.

fxd0h:
Fixes #494.

@TresPies-source's analysis holds on current main, both layers of it. The linear spacing divides by `(num_steps - 1)`, and guarding just that denominator moves the crash into `_stretch_to_terminal`: with a single point the stretch maps that point onto `shift_terminal` itself, which would start denoising from nearly-clean noise.

The class already defines the 1-step schedule in its two sibling paths (`get_timesteps_and_sigmas` and `set_mu`): sigmas `[1.0, 0.0]`, one full Euler step. `_compute_timesteps_and_sigmas` now returns that same schedule for `num_steps == 1` and leaves every other step count byte-identical.

Tests: the 1-step case and the sibling-path agreement both fail on main with the reported `ZeroDivisionError` and pass here; n=2 and n=4 regression tests pin the existing values (first sigma 1.0, last interior sigma at `shift_terminal` 0.02, strictly decreasing).

Test output, this branch on current main (M5 Max 40-core GPU, macOS 26.5, Python 3.12):

```
mlx 0.31.0: 517 passed, 83 deselected
mlx 0.32.0: 517 passed, 83 deselected
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed one-step inference scheduling to use the correct full denoising path.
  * Prevented invalid timestep calculations and division-by-zero errors during single-step generation.
  * Preserved existing behavior for multi-step schedules.
  * Improved reliability and consistency of generated results when using a single inference step.

* **Tests**
  * Added coverage for one-step and multi-step timestep and sigma schedules.
  * Verified consistent scheduler output across supported inference modes.

Defines the scheduler’s degenerate one-step case as a full Euler denoising step while preserving existing multi-step behavior.
- Returns sigmas `[1.0, 0.0]` and timestep `[1000.0]` for one-step inference.
- Adds regression coverage for one-step behavior, sibling-path agreement, and unchanged multi-step schedules.
</details>

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/mflux/models/common/schedulers/flow_match_euler_discrete_scheduler.py | Adds a one-step early return whose array ordering, lengths, dtypes, and values match downstream indexing and existing sibling schedule paths. |
| tests/schedulers/test_flow_match_euler_one_step.py | Covers the corrected one-step schedule and pins representative multi-step schedules against regression. |

</details>

<sub>Reviews (3): Last reviewed commit: ["Merge branch &#39;main&#39; into upstream-steps1..."](https://github.com/mflux-community/mflux/commit/7ddc9b64a082a0642d164cea72609b314b927738) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51896684)</sub>


fxd0h:
@coderabbitai full review

fxd0h:
@greptileai review

fxd0h:
Taken: the schedules for 2 and 4 steps are pinned in full now, sigmas and timesteps both. A test named "unchanged" that only checked the endpoints could not have caught a change in the interior.

anthonywu (approved):
fix is tightly scoped to the one scheduler's workflow, LGTM

ianscrivener (approved):
<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/mflux-community/mflux/pull/496)</sub>

Issue #498, opened 2026-08-01
--negative-prompt is accepted and silently discarded on FLUX.1, and inconsistently handled elsewhere


fxd0h:
The shared parser exposes `--negative-prompt` on every image CLI (`parsers.py:118`). What happens to it after that depends on the family. Checked against current main:

**FLUX.1: dead parameter.** `flux_generate.py:66` reads the flag and passes it into `Flux.generate_image`; the name appears exactly twice in the whole `src/mflux/models/flux/` tree, that call and the signature that receives it (`flux.py:59`). The value is read from the user, plumbed through one call, and dropped, on dev and schnell, with nothing said. If there is a consumer I did not find I am happy to be corrected, but a grep of the family tree comes back with only those two hits.

**Ideogram 4: no parameter at all.** There is no `negative_prompt` anywhere under `src/mflux/models/ideogram4/`, while the CLI still accepts the flag. This family already warns for `--steps` and `--guidance` (`ideogram4_generate.py:44-46`), so the idiom exists, just not for this flag.

**Z-Image and the guidance-distilled models: dropped for a correct reason, silently.** `z_image.py:60-61` pins guidance to 0.0 when `supports_guidance` is false, and the encode path skips the negative encoding at guidance <= 1.0. The math is right, CFG needs guidance above 1. But it is silent, including when the user passed an explicit `--guidance` that then got overridden. And on base Z-Image an omitted `--guidance` defaults to 0.0, so the everyday invocation with a negative prompt and no guidance flag ignores the negative prompt without a word.

Why this matters beyond tidiness: downstream tools transcribe capability lists from `--help`, and `--help` currently says every model takes a negative prompt. We hit this through a tool that had been declaring `--negative-prompt` support on models that discard it, including in its defaults.

We shipped one policy for this in mflux-cv (HowDidTheCatGetSoFat/mflux-cv#39): keep accepting the options so existing scripts do not break, and warn in Ideogram's existing message format whenever a provided option will be dropped, keyed on the effective post-resolution behavior rather than the flag value. Happy to port that here if warn-everywhere is the direction you want. Erroring instead, or removing the dead FLUX.1 parameter, are the other options; that call is yours.

PR #500, opened 2026-08-02
Read lora metadata defensively in the stepwise handler

fxd0h:
`--stepwise-image-output-dir` dies with `AttributeError` on the first preview step for models that never gain a `lora_paths` attribute. FIBO is the reproducible case on current main: `Fibo.__init__` accepts `lora_paths` and passes it into `FiboInitializer.init`, whose `_init_config` assigns `model_config`, `callbacks` and `tiling_config` but never `lora_paths`, so the handler's unconditional `self.model.lora_paths` read raises.

The fix mirrors what we shipped in mflux-cv 0.18.28 after hitting exactly this crash: read both attributes with `getattr` and a `None` default. The new test extends the routing test file from #444 with the attribute-less model shape its existing stubs never covered; it fails on main with the `AttributeError` and passes here.

Side observation for the #498 inventory: FIBO's constructor `lora_paths` parameter is itself accepted and silently dropped, since the initializer never applies or stores it.

Test output, this branch on current main (M5 Max 40-core GPU, macOS 26.5, Python 3.12):

```
mlx 0.31.0: 514 passed, 83 deselected
mlx 0.32.0: 514 passed, 83 deselected
```

The PR prevents stepwise preview generation from crashing when a model does not define LoRA metadata attributes.
- Reads `lora_paths` and `lora_scales` defensively with `None` defaults.
- Adds a regression test covering an attribute-less model and successful preview output.
</details>

The PR appears safe to merge, with no blocking or non-blocking defects identified.

The defensive reads preserve existing metadata for models that define the attributes while safely supplying supported `None` values for models that do not.

| Filename | Overview |
|----------|----------|
| src/mflux/callbacks/instances/stepwise_handler.py | Safely defaults absent LoRA metadata attributes to values already supported by the generated-image metadata contract. |
| tests/callbacks/test_stepwise_handler_vae_routing.py | Adds focused regression coverage proving stepwise output succeeds when both LoRA attributes are absent. |

</details>

<sub>Reviews (1): Last reviewed commit: ["Read lora metadata defensively in the st..."](https://github.com/mflux-community/mflux/commit/a533a9be0ae54e06f9d565125ca9af9a10113061) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51896680)</sub>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image preview generation for models that do not define optional LoRA configuration settings.
  * Ensured stepwise processing continues successfully and produces the expected preview image in this scenario.

* **Tests**
  * Added regression coverage for models without LoRA path and scale settings.

### DIFF
--- a/src/mflux/callbacks/instances/stepwise_handler.py
+++ b/src/mflux/callbacks/instances/stepwise_handler.py
@@ -97,8 +97,8 @@ class StepwiseHandler(BeforeLoopCallback, InLoopCallback, InterruptCallback):
             seed=seed,
             prompt=prompt,
             quantization=self.model.bits,
-            lora_paths=self.model.lora_paths,
-            lora_scales=self.model.lora_scales,
+            lora_paths=getattr(self.model, "lora_paths", None),
+            lora_scales=getattr(self.model, "lora_scales", None),
             generation_time=generation_time,
         )
         stepwise_img.save(
--- a/tests/callbacks/test_stepwise_handler_vae_routing.py
+++ b/tests/callbacks/test_stepwise_handler_vae_routing.py
@@ -102,3 +102,18 @@ def test_stepwise_vae_without_packed_decode_falls_back(tmp_path) -> None:
     handler.call_before_loop(seed=1, prompt="a", latents=latents, config=_config(ModelConfig.dev()))
 
     assert vae.calls == ["decode"]
+
+
+@pytest.mark.fast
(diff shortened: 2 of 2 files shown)


fxd0h:
@coderabbitai full review

fxd0h:
@greptileai review

fxd0h:
@coderabbitai full review

PR #358, opened 2026-02-16
feat: expose img2img CLI flags for Flux.2 Klein

terribilissimo:
Add --image-path and --image-strength to mflux-generate-flux2, following the same pattern used by flux_generate.py (Flux.1).

>Note: This is independent of #353  (Z-Image enhancements), and can be reviewed/merged separately.

Three additions:
1. parser.add_image_to_image_arguments(required=False)
2. reference_image_path=args.image_path in DimensionResolver.resolve()
3. image_path + image_strength passed to model.generate_image()

## Summary

The `Flux2Klein` Python class already implements full img2img support via
`_prepare_img2img_latents()`, accepting `image_path` and `image_strength`
parameters in `generate_image()`. However, the `mflux-generate-flux2` CLI
does not register `--image-path` / `--image-strength`, so users cannot
access img2img from the command line.

This PR adds 3 lines to `flux2_generate.py` to expose these flags,
following the exact same pattern used by `flux_generate.py` (Flux.1):

1. `parser.add_image_to_image_arguments(required=False)`
2. `reference_image_path=args.image_path` in `DimensionResolver.resolve()`
3. `image_path=args.image_path` + `image_strength=args.image_strength` in `generate_image()`

When `--image-path` is not provided, behavior is identical to upstream.

## **Why this is different from `mflux-generate-flux2-edit`**

Flux.2 Klein already has an **image-conditioned editing** CLI
(`mflux-generate-flux2-edit` with `--image-paths`), but that is a
fundamentally different operation from **img2img**:

| | img2img (`--image-path`) | Image-conditioned editing (`--image-paths`) |
|---|---|---|
| **Purpose** | Transform a single input image while preserving its overall structure and composition | Use one or more reference images as semantic context for generation |
| **How it works** | Encodes the input into latent space, adds noise controlled by `--image-strength`, then denoises — the result follows the spatial layout of the original | Reference images are injected as conditioning tokens alongside the text prompt — the model draws on their content freely |
| **Strength control** | `--image-strength` (0.0 = ignore input, 1.0 = pure noise) controls how much of the original is preserved | No strength parameter — the prompt determines what changes |
| **Number of inputs** | Exactly 1 image | 1 or more reference images |
| **Typical use cases** | Style transfer, color grading, detail refinement, artistic re-interpretation of a photo | "Put these sunglasses on this person", combining elements from multiple images, subject-driven editing |
| **CLI command** | `mflux-generate-flux2` (this PR) | `mflux-generate-flux2-edit` (already exists) |

[...]

PR #503, opened 2026-08-03
fix: add @with_gpu_stream decorator for thread-safe GPU stream access

### LINKED ISSUE #502: RuntimeError: 'There is no Stream(gpu, N) in current thread' when calling generate_image from non-main thread
## Bug Report

**Scenario — generate_image crashes in thread executor**

1. A caller runs `model.generate_image(...)` inside a `concurrent.futures.ThreadPoolExecutor` (e.g., a web server like ComfyUI handling requests on worker threads).
2. Inside the denoising loop, `mx.eval(latents)` triggers GPU computation.
3. MLX raises `RuntimeError: There is no Stream(gpu, 4) in current thread` because the worker thread has no GPU stream.

**Root Cause**: MLX uses thread-local GPU streams. The default stream is only created on the main thread. When `mx.eval()` runs on a non-main thread, no stream exists for that thread, causing the crash. mflux has 39 `mx.eval()` calls across 32 files and zero thread-safety / stream management code.

**Reproduction** (requires Apple Silicon + MLX):
```python
from concurrent.futures import ThreadPoolExecutor
from mflux import Flux1

model = Flux1.from_name("flux-schnell")

def gen():
    return model.generate_image(seed=42, prompt="a cat", num_inference_steps=4)

# This works:
gen()

# This crashes with RuntimeError:
with ThreadPoolExecutor(max_workers=1) as ex:
    ex.submit(gen).result()
```

**Error**:
```
RuntimeError: There is no Stream(gpu, 4) in current thread
```

**Fix**: Add a thread-local GPU stream context manager that each `generate_image()` method wraps its body in. MLX provides `mx.new_thread_local_stream(mx.gpu)` and `mx.stream()` for exactly this purpose. When running on the main thread, the context manager is a no-op (the default stream already exists). When running on a worker thread, it creates the missing stream.

This is especially important for downstream integrations (ComfyUI, web APIs, async servers) that need to run generation on worker threads.

**Affected files** (all 21 `generate_image` methods + initializers with `mx.eval`):
- `src/mflux/models/flux/variants/*/flux*.py` (10 files)
- `src/mflux/models/flux2/variants/*/flux2_*.py` (2 files)
- `src/mflux/models/krea2/variants/txt2img/krea2.py`
- `src/mflux/models/ideogram4/variants/txt2img/ideogram4.py`
- `src/mflux/models/qwen/variants/*/qwen_image*.py` (2 files)
- `src/mflux/models/fibo/variants/*/fibo*.py` (2 files)
- `src/mflux/models/z_image/variants/z_image.py`
- `src/mflux/models/ernie_image/variants/txt2img/ernie_image.py`
- `src/mflux/models/seedvr2/variants/upscale/seedvr2.py`
- `src/mflux/models/krea2/krea2_initializer.py`
- `src/mflux/models/ideogram4/ideogram4_initializer.py`
- `src/mflux/models/ernie_image/ernie_image_initializer.py`

dahai80:
## Summary

Fixes #502

When `generate_image()` or model initialization runs in a worker thread (e.g., from ComfyUI, web APIs, or `concurrent.futures.ThreadPoolExecutor`), MLX raises:

```
RuntimeError: There is no Stream(gpu, 4) in current thread
```

This happens because MLX GPU streams are thread-local — the default stream only exists on the main thread. mflux has 39 `mx.eval()` calls but zero thread-safety code.

## Fix

Added `@with_gpu_stream` decorator in `mflux/utils/gpu_stream.py` that:
- **On the main thread**: no-op, uses the existing default stream (zero overhead)
- **On worker threads**: creates a new GPU stream via `mx.new_stream(mx.gpu)` and wraps the function call in `mx.stream(stream)`

Applied to:
- All 21 `generate_image()` methods across model variants
- 3 initializer classes with `mx.eval(model)` (Krea2, Ideogram4, ErnieImage)
- `FiboVLM.generate()` method

## Testing

- 507 fast tests pass (no regressions)
- Functional test verified: `@with_gpu_stream` correctly creates GPU streams in `ThreadPoolExecutor` workers
- Without the decorator, complex model inference (multi-step denoising with compiled kernels) crashes in worker threads

## Reproduction (before fix)

```python
from concurrent.futures import ThreadPoolExecutor
from mflux import Flux1

model = Flux1.from_name("flux-schnell")

# This works:
model.generate_image(seed=42, prompt="a cat", num_inference_steps=4)

# This crashes with RuntimeError:
with ThreadPoolExecutor(max_workers=1) as ex:
    ex.submit(model.generate_image, seed=42, prompt="a cat", num_inference_steps=4).result()
```

After this fix, both cases work correctly.

### DIFF
--- a/src/mflux/models/ernie_image/ernie_image_initializer.py
+++ b/src/mflux/models/ernie_image/ernie_image_initializer.py
@@ -13,10 +13,12 @@ from mflux.models.ernie_image.model.ernie_transformer.transformer import ErnieTr
 from mflux.models.ernie_image.weights.ernie_lora_mapping import ErnieLoRAMapping
 from mflux.models.ernie_image.weights.ernie_weight_definition import ErnieWeightDefinition
 from mflux.models.flux2.model.flux2_vae.vae import Flux2VAE
+from mflux.utils.gpu_stream import with_gpu_stream
 
 
 class ErnieImageInitializer:
     @staticmethod
+    @with_gpu_stream
     def init(
         model,
         model_config: ModelConfig,
--- a/src/mflux/models/ernie_image/variants/txt2img/ernie_image.py
+++ b/src/mflux/models/ernie_image/variants/txt2img/ernie_image.py
@@ -17,6 +17,7 @@ from mflux.models.ernie_image.weights.ernie_weight_definition import ErnieWeight
 from mflux.models.flux2.model.flux2_vae.vae import Flux2VAE
 from mflux.utils.apple_silicon import AppleSiliconUtil
 from mflux.utils.exceptions import StopImageGenerationException
+from mflux.utils.gpu_stream import with_gpu_stream
 from mflux.utils.image_util import ImageUtil
 
 
@@ -43,6 +44,7 @@ class ErnieImage(nn.Module):
             lora_scales=lora_scales,
         )
 
+    @with_gpu_stream
     def generate_image(
         self,
         seed: int,
--- a/src/mflux/models/fibo/variants/edit/fibo_edit.py
+++ b/src/mflux/models/fibo/variants/edit/fibo_edit.py
@@ -18,6 +18,7 @@ from mflux.models.fibo.variants.edit.util import FiboEditUtil
 from mflux.models.fibo.weights.fibo_weight_definition import FIBOWeightDefinition
 from mflux.utils.exceptions import StopImageGenerationException
 from mflux.utils.generated_image import GeneratedImage
+from mflux.utils.gpu_stream import with_gpu_stream
 from mflux.utils.image_util import ImageUtil
 
 
@@ -44,6 +45,7 @@ class FIBOEdit(nn.Module):
             model_config=model_config,
         )
 
+    @with_gpu_stream
     def generate_image(
         self,
         seed: int,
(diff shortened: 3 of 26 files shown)


dahai80:
@fxd0h You're right, and thank you for the careful measurement.

Root cause is on our (caller) side, not mflux. We are [fusion-mlx](https://github.com/dahai80/fusion-mlx), which wraps mflux and runs `generate_image()` on a dedicated `mlx-image` `ThreadPoolExecutor` whose thread initializer (`_init_mlx_thread`) binds a thread-local default stream plus `generation_stream` for mlx-lm / mlx-vlm. A second diffusion worker (`diffusion_lane`) independently calls `mx.set_default_stream(mx.default_stream(...))`. With several such non-default streams alive in the process, a worker thread that did not itself bind a stream hit `Stream(gpu, 4)`. Plain worker threads (your repro) are fine — the crash needs the cross-thread stream context our executors create, which is why mflux alone never reproduces it.

Our fix landed entirely in fusion-mlx:
- `ImageGenEngine._generate()` now calls `mx.default_stream(mx.default_device())` at entry (commit `6025b07`, fix #329).
- Follow-ups #339 / #358 added defensive stream setup for the IP-Adapter and img2img paths.

So this PR's per-worker `mx.new_stream` is the wrong layer — it changes stream-ordering semantics for every caller to paper over a stream context our own executor was leaking. Closing #503.

#502 stays open only in case you'd like an upstream-side improvement (e.g. a clearer error message when a non-default stream is used off its creating thread). Happy to file that as a smaller issue, or close #502 too — your call.

dahai80:
Closing — root cause was in the caller (fusion-mlx executor stream binding), fixed there in #329/#339/#358. See comment above.

PR #517, opened 2026-08-10
Qwen: mixed-precision -q 4 fixes the step-noise accumulation (#484)

### LINKED ISSUE #484: Qwen-Image 4-bit: quantization noise compounds per denoising step (flat-field σ grows monotonically with steps)
## Summary

Running a 4-bit quantized Qwen-Image export on mflux 0.18.0, background grain/mottling gets **worse as step count increases**, not better. A flat-field diagnostic shows high-frequency noise growing monotonically with steps, which points to per-step accumulation of quantization error rather than a fixed noise floor.

## Environment

- mflux **0.18.0** (uv tool install), Apple M2 Max, 32 GB, macOS (Darwin 25.5)
- Model: `OsaurusAI/Qwen-Image-mflux-4bit` (saved with mflux 0.18, so not the pre-0.13 vintage problem from #296)

## Repro

Render a flat field — a prompt asking for a single solid white color block fill — and measure high-frequency noise as `σ = std(image − gaussian_blur(image, 4))` over a 512² crop:

```bash
mflux-generate-qwen --model OsaurusAI/Qwen-Image-mflux-4bit \
  --prompt "A plain solid pure white image. Uniform white color fill, nothing else." \
  --seed 42 --width 512 --height 512 --steps <N>
```

| Steps | hf σ | Mean (target 255) |
|---|---|---|
| 4 | 2.48 | ~252 |
| 8 | 4.14 | — |
| 20 | 6.42 | — |
| 50 | 7.36 | ~224 |

Both trends move the wrong way with more steps: noise grows ~3×, and the white point drifts 252 → 224. The same instrument on other 4-bit families on the same machine/runtime reads an order of magnitude lower and step-stable (FLUX.2-klein-4B 4-bit: σ 0.33; Z-Image-Turbo 4-bit: σ 0.44) — so this looks specific to how the Qwen path interacts with uniform 4-bit quantization, not a general runtime issue.

Things we ruled out: guidance (tested at 4.0), native resolution (1664×928), prompt phrasing/negative prompts — none change the grain. The RGB single-color variants of the flat-field test show the same behavior.

## Hypothesis / possible fix vector

Uniform q4 across all layers + iterative denoising means per-step weight error compounds across the trajectory (a single-pass q4 LLM never accumulates this way). The mixed-precision quantization work in #420 (protecting sensitive layers from q4) seems like the natural fix vector — Qwen-Image may be an especially strong motivating case for it, since its practical step optimum on 4-bit collapses to ~8 steps (above that, added steps degrade the image).

Happy to run further diagnostics on this hardware if useful.

fxd0h:
Fixes #484. At uniform 4-bit the flat-field noise compounds per denoising step, sigma grows 5.06 to 16.07 from 4 to 50 steps in the reproduction posted there. Bisection showed the accumulation lives entirely in the adaLN `img_mod_linear` layers: keeping just those at 8-bit recovers the q8/bf16 floor (sigma 1.10 at 20 steps, 1.37 at 50) for about 1.8GB extra, while protecting txt_mod adds nothing. Full measurements are in the issue. Same layer choice lpalbou converged on in #420, now with the data.

How:

- The Qwen quantization predicate is now bits-aware and returns `{"bits": 8}` for `.img_mod_linear` when quantizing at 4-bit. Existing two-arg predicates keep working through a signature-aware adapter.
- Loading a stored quantized model no longer assumes uniform layers: bits and group size are inferred per layer, exactly, from the stored shapes and the module's real input dims. A mixed save round-trips pixel-identically, diff 0.0 on a 26GB save.
- Qwen edit and derived variants inherit the fix through the shared weight definition. 14 fast tests cover the predicate, the adapter and the round-trips; the full fast suite passes.

Shipping in mflux-cv since 0.18.38.

## Summary by CodeRabbit

* **New Features**
  * Added support for mixed-precision quantization, allowing different layers to use different bit widths and group sizes.
  * Preserves and restores per-layer quantization settings when loading previously quantized models.
  * Supports component-level weight quantization and fallback behavior for incomplete metadata.

* **Bug Fixes**
  * Ensures selected Qwen image-processing layers remain at 8-bit precision during 4-bit quantization.

* **Tests**
  * Added coverage for mixed quantization, metadata restoration, unquantized layers, and output consistency.

The PR introduces mixed-precision Qwen quantization and reconstructs each stored layer’s bit width and group size when loading quantized checkpoints.
- Keeps Qwen `img_mod_linear` layers at 8-bit during requested 4-bit quantization.
- Adapts existing two-argument predicates while passing requested precision to bits-aware predicates.
- Restores quantized layer structure from packed weight and scale shapes before loading values.
- Adds fast tests for mixed precision, group-size restoration, skipped layers, predicate compatibility, and output equality.
</details>

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| src/mflux/models/common/weights/loading/weight_applier.py | Adds bits-aware predicate adaptation and per-layer reconstruction of stored quantization parameters before loading checkpoint tensors. |
| src/mflux/models/qwen/weights/qwen_weight_definition.py | Selects 8-bit quantization for Qwen image modulation linear layers when the requested global level is 4-bit. |
| tests/test_qwen_mixed_q4.py | Covers predicate adaptation, mixed bit widths and group sizes, unquantized layers, loading-path wiring, and output-preserving reconstruction. |

</details>

```mermaid
flowchart TD
    A[Load checkpoint metadata] --> B{Stored quantization?}
    B -- No --> C[Load unquantized weights]
    C --> D[Apply requested quantization]
    B -- Yes --> E[Inspect each layer's stored weight and scale shapes]
    E --> F[Infer stored bits and group size]
    F --> G[Rebuild quantized module structure]
    G --> H[Load stored tensor values]
    D --> I[Ready model]
    H --> I
```
</details>

<sub>Reviews (4): Last reviewed commit: ["Merge branch &#39;main&#39; into qwen-mixed-q4-u..."](https://github.com/mflux-community/mflux/commit/37341650a295c4200933f9cfe79f62956ae8f331) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51896672)</sub>


fxd0h:
@coderabbitai full review

fxd0h:
@greptileai review

fxd0h:
@coderabbitai full review

fxd0h:
Took three of the four in 1dcdfbc.

The arity check was the real one: counting every parameter kind means `def p(path, module, *, debug=False)` would have been called with `bits` positionally and raised. It now counts only positional-capable parameters, with a test for exactly that signature.

The inference fallback now names the layer and the shapes that did not add up. It structures the layer at the requested precision while the file holds another, so `update()` may reject it later, and until now that happened without a word.

On coverage you were right that everything went through the private helpers: two tests now call `_quantize` itself, covering the mixed rebuild and the `skip_quantization` skip.

Leaving the `weight_subkey` resolution in `apply_and_quantize_single` alone. No caller sets it on that path, as you note, so adding the branch means shipping code nothing exercises; it belongs with the first caller that needs it.

anthonywu (approved):

PR #519, opened 2026-08-10
Exclude slow tests from the default pytest run

fxd0h:
Running plain `pytest` on a fresh clone starts downloading full model checkpoints. The default addopts only filters `high_memory_requirement`, and every test that instantiates a model is marked `slow`, so they all run: the ernie model_saving test alone pulls ~12GB before the suite gets anywhere. I hit this on a nearly full disk this week, and the offline failure @plz12345 ran into while reviewing #440 (test_metadata_complete trying to fetch Z-Image-Turbo) is the same root cause.

This adds `not slow` to the default marker filter. The unmarked utility tests keep running, so the default remains a superset of what CI checks (tests.yml passes `-m fast` explicitly and is unaffected). With this change a bare `pytest` completes offline: 549 passed in under 2 seconds, zero downloads. The slow and golden suites stay available as an explicit opt-in via `-m slow`.

This PR makes bare pytest and `make test` skip slow model tests while preserving explicit full, fast, and slow suite targets.
- Adds `not slow` to pytest’s default marker expression.
- Adds `make test-all` to run slow tests explicitly while retaining the high-memory exclusion.
- Reclassifies two lightweight, synthetic LoKr integration tests as fast.
- Updates testing guidance and Makefile help text to document the new behavior.
</details>

The PR appears safe to merge, with the updated targets and markers consistently implementing the documented test-selection behavior.

The default invocation excludes slow tests as intended, explicit marker arguments preserve the existing fast and slow workflows, `test-all` restores the eligible full suite, and the newly fast LoKr tests do not access models or the network.

| Filename | Overview |
|----------|----------|
| pyproject.toml | Updates the default pytest marker selection to exclude slow and high-memory tests; explicit command-line marker selections continue to override this default. |
| Makefile | Changes `make test` documentation to describe the default subset and adds a functional `test-all` target that explicitly includes slow tests. |
| tests/lora/test_lokr_integration.py | Reclassifies two genuinely lightweight tests that use small synthetic local tensors from slow to fast. |
| .cursor/skills/mflux-testing/SKILL.md | Aligns repository testing guidance with the new default selection and explicit full-suite target. |

</details>

<sub>Reviews (1): Last reviewed commit: ["Remark the weight-free LoKr apply\_layers..."](https://github.com/mflux-community/mflux/commit/70422b3c6f85d2453b1ba3e68e25971a5eab425f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51896669)</sub>

### DIFF
--- a/.cursor/skills/mflux-testing/SKILL.md
+++ b/.cursor/skills/mflux-testing/SKILL.md
@@ -16,7 +16,8 @@ This repo uses pytest with image-producing tests. Always preserve outputs for in
 - Prefer the Makefile test targets:
   - `make test-fast` (fast tests, no image generation)
   - `make test-slow` (slow tests, image generation)
-  - `make test` (full suite)
+  - `make test` (default selection, skips slow model tests)
+  - `make test-all` (full suite, slow tests download model weights)
 - Always keep `MFLUX_PRESERVE_TEST_OUTPUT=1` on test runs (already built into the Makefile test targets).
 - If a change affects defaults, config resolution, metadata fields, or CLI behavior, add or update tests that cover the changed behavior directly instead of relying only on manual verification.
 - If tests fail:
--- a/Makefile
+++ b/Makefile
@@ -89,13 +89,21 @@ check: ensure-ruff
 	# 🏗️ Running pre-commit linter and formatters on files...
 	@(pre-commit run --all-files)
 
-# Run tests
+# Run the default test selection (excludes slow model tests, see addopts in pyproject.toml)
 .PHONY: test
 test: ensure-pytest
 	# 🏗️ Running tests...
 	uv pip install -e '.[dev]' # Install pinned MLX version specifically for testing
 	MFLUX_PRESERVE_TEST_OUTPUT=1 uv run python -m pytest
 	# ✅ Tests completed
+
+# Run the whole suite, slow tests included (downloads model weights)
+.PHONY: test-all
+test-all: ensure-pytest
+	# 🏗️ Running all tests (slow tests download model weights)...
+	uv pip install -e '.[dev]' # Install pinned MLX version specifically for testing
+	MFLUX_PRESERVE_TEST_OUTPUT=1 uv run python -m pytest -m "not high_memory_requirement"
+	# ✅ All tests completed
 
 # Run fast tests only (no image generation)
 .PHONY: test-fast
@@ -145,7 +153,8 @@ help:
 	@echo "  make lint        - Run ruff python linter"
 	@echo "  make format      - Run ruff code formatter"
 	@echo "  make check       - Run linters auto fixes *and* style formatter via pre-commit hook"
-	@echo "  make test        - Run all tests"
+	@echo "  make test        - Run the default test selection (skips slow model tests)"
+	@echo "  make test-all    - Run the whole suite, slow tests included (downloads weights)"
 	@echo "  make test-fast   - Run fast tests only (no image generation)"
 	@echo "  make test-slow   - Run slow tests only (image generation)"
 	@echo "  make build       - Build distribution packages and check sizes"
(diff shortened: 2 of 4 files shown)


plz12345:
I verified the core claim locally and it holds. Simulating the change with a command-line `-m` (which overrides `addopts`) and `--collect-only`, on my tree:

| selection | collected |
| --- | --- |
| `not high_memory_requirement` (today's default) | 723 |
| `not high_memory_requirement and not slow` (this PR) | 672 (51 deselected) |
| `fast` (what CI runs) | 543 |
| `slow` | 51 |
| unmarked | 129 |
| `fast and slow` | **0** |

No test carries both markers, so the new default is still a strict superset of CI's `-m fast`, and it drops exactly the 51 `slow` tests while keeping all 129 unmarked ones. `make test-slow` and `-m slow` opt-in are unaffected since a CLI `-m` wins over `addopts`. So the mechanism is sound.

One thing I'd like fixed in this PR before it lands, plus two notes.

### Please fix: two mismarked tests lose their only coverage

`test_flux2_klein_lokr_safetensors_apply_layers` (`tests/test_lokr_integration.py:51`) and `test_flux1_lycoris_lokr_safetensors_apply_layers` (`:125`) are marked `slow`, but they download nothing and run in milliseconds — they build a tiny in-memory transformer and write a safetensors file to `tmp_path`:

```
$ HF_HUB_OFFLINE=1 pytest -q -m "" tests/test_lokr_integration.py -k apply_layers
2 passed in 0.66s
  test_flux2_klein_lokr_safetensors_apply_layers: 0.0180s
  test_flux1_lycoris_lokr_safetensors_apply_layers: 0.0050s
```

They matter because they're the only tests that exercise the LyCORIS `lokr_w1`/`lokr_w2` + `dora_scale` path in `LoRALoader._apply_single_lora` — the unmarked siblings at `:78` and `:103` never pass a `dora_scale` key.

After this change nothing routine runs them: the new default deselects them, CI is `-m fast`, and `make test-slow` pulls ~12GB of checkpoints so nobody runs it casually. For these two the marker change isn't deferring a slow test, it's dropping coverage. Swapping `slow` for `fast` on both is a two-line fix and keeps this PR purely additive in terms of what gets skipped.

### Note: no target runs the whole suite anymore

`make test` is bare `pytest`, so it inherits the new filter. It's documented as "Run all tests" (`Makefile:148`) and it's what the default `all: install test` target runs (`Makefile:9`); `.cursor/skills/mflux-testing/SKILL.md:19` also describes it as the full suite. After this change `make test` drops `slow` and `make test-slow` drops `fast` and unmarked, so there's no single command for everything, and a contributor running `make all` on a fresh clone gets a green run with all image-generation, golden, training and VLM tests silently deselected.

Not necessarily this PR's job to solve, but the help text is now wrong either way. Cheapest fix is rewording it and adding a `test-all` target that passes `-m "not high_memory_requirement"` explicitly.

### Note: naming a path no longer runs its slow tests

Because the filter lives in `addopts`, it applies even when you name a file explicitly. For mixed files that's silent:

```
$ pytest tests/test_lokr_integration.py
2/4 tests collected (2 deselected)   # exit 0
```

An all-slow file at least exits non-zero (4, no tests collected), but the mixed case looks like a clean pass. That's inherent to the `addopts` approach and probably an acceptable trade, just worth knowing before someone debugs it later.

Fix the two markers and this is good to go from my side.

fxd0h:
Both markers swapped, and you were right about what they cover: they are the only tests exercising the LyCORIS `dora_scale` path in the loader. Rebased on current main while at it so the branch sees the LoKr file.

Took the Makefile note too since this PR caused it: `make test`'s help text now says it skips slow model tests, and a new `make test-all` passes `-m "not high_memory_requirement"` explicitly for the full run, with the testing skill doc updated to match. The path-naming trade-off I left as is.

On the rebased branch the default run is 590 passed offline, and the two remarked tests run under `-m fast`.

PR #590, opened 2026-08-14
dx: replace Makefile with justfile and lint it in CI

anthonywu:
## What

Replaces the `Makefile` with an idiomatic [just](https://github.com/casey/just) justfile (requires just ≥ 1.50; bare `just` / `just --list` shows all recipes with descriptions), adds a pinned `just-setup` composite action (1.58.0, SHA256-verified download), and lints the justfile in the existing CI lint job via `just --fmt --check` (`just lint-justfile`; fix with `just fmt-justfile`). `.cursor` rules/skills, this PR template, and `CHANGELOG.md` were updated from `make` to `just`; historical `make` mentions in released changelog sections are intentionally kept.

Intentional deltas from strict post-#576 Makefile parity:

- `expect-uv` now exits 1 after printing install guidance, instead of falling through into a failing `uv` invocation.
- `venv-init` uses `uv venv --clear`, so re-running `just install` is idempotent — note it removes whatever sits at `.venv`, not only recognized venvs.
- `expect-arm64` checks **host** arch via `uname -m`; just's `arch()` is the build arch of the just binary (an x86_64 just under Rosetta would falsely refuse).
- `just-setup` has no cache step: caching skipped the SHA256 check on hits, and restore cost ≈ download for a 2.4 MB tarball — so every run now downloads and verifies.
- Test recipe doc comments reworded (e.g. `test-all` names the `high_memory_requirement` exclusion); the pytest selectors are byte-identical.

## Checklist (definition of done)

- [ ] Tests added/updated run in CI by default — N/A: no runtime or test changes (dev tooling only).
- [x] `ruff check` and `ruff format` are clean (`uv run ruff` uses the version pinned in the dev dependencies of `pyproject.toml`, which is the single source of truth for pre-commit and CI; `pre-commit run -a` covers it locally).
- [x] `CHANGELOG.md`: entry under `Unreleased` referencing this PR number.
- [x] Docs updated where behavior changed — README examples/table rows are part of the API contract (see `.cursor/rules/RULE.md`). (RULE.md, mflux-* skills, PR template; README had no `make` references.)
- [ ] New model — N/A.
- [ ] New/changed CLI — N/A.

## Verification

- `just install`, `just lint`, `just lint-justfile`, `just build` all run clean locally (incl. re-run idempotency of `install`).
- `just -n` dry-runs of `install test test-all build clean` expand correctly (`uv sync` semantics; `-m "not high_memory_requirement"` quoting intact).
[...]

PR #597, opened 2026-08-15
dx: migrate type checking from mypy to ty 0.0.72

anthonywu:
## What

Migrates type checking from mypy to astral's **ty**, pinned exactly to `ty==0.0.72` (62ebbecc7 2026-08-14) in `pyproject.toml` dev deps, the `astral-sh/ty-pre-commit` rev, and CI (via `uv.lock` in the macOS tests job, where the synced venv lets ty resolve mlx/torch imports). Adds a `just typecheck` recipe and documents the workflow in the dev-env skill.

Why now:

1. The Python community has converged on astral's tooling — we already run uv and ruff, and ty is mature enough to adopt.
2. Checker speed compounds in local dev and CI: ty full-checks src+tests in **~0.4s** vs mypy's 2–9s (M4 Max; and GH Actions rounds runner minutes up, so staying under the minute boundary matters).
3. We're making sweeping dev-ux changes right now (justfile #590, CI overhaul #576) — consolidating this migration into the same effort.

One-time baseline: the 15 rules with pre-existing violations (828 diagnostics) are set to `"ignore"` in `[tool.ty.rules]` with per-rule counts. Plan: re-enable and fix piecemeal, one rule per PR (documented in `.cursor/skills/mflux-dev-env/SKILL.md`). New code must pass all rules not in the baseline. Notably, the old mypy setup (changed-files-only pre-commit hook, 10 error codes disabled, isolated env) had already let 5 errors slip into `src/`; those are part of the baseline to burn down.

Side fix: whitelist typos false positives (`hed`, `thre`, `caf`, `mis`, `unparseable`) in `_typos.toml` — the typos hook runs with `--write-changes` and would destructively rewrite the HED model (including renaming `hed.py` → `head.py`) on any commit touching those files.

- [ ] Tests added/updated run in CI by default — N/A: dev-tooling change, no runtime behavior; existing suite passes (1029 fast tests).
- [x] `ruff check` and `ruff format` are clean (`pre-commit run -a` passes: ruff-check, ruff-format, typos, ty).
- [x] `CHANGELOG.md`: entry under `Unreleased` referencing this PR number.
- [x] Docs updated where behavior changed — `justfile` (`just typecheck`), `.cursor/skills/mflux-dev-env/SKILL.md` (ty workflow + burn-down plan). README unaffected (no user-facing behavior change).
- [ ] New model — N/A.
- [ ] New/changed CLI — N/A.

## Verification

```
$ uv run ty check
All checks passed!            # exits 0 against the baseline

$ pre-commit run --all-files
ruff check / ruff format / typos / ty ......... Passed

$ just typecheck              # new recipe
🏗️ Running ty type checker (pinned via pyproject.toml)...
uv sync --all-extras
Resolved 110 packages in 3ms
Checked 73 packages in 1ms
uv run --no-sync ty check
All checks passed!
✅ Type checking complete.

$ just lint-justfile          # justfile formatting still clean

$ uv run python -m pytest -m fast -q
1029 passed, 3 skipped, 211 deselected in 4.29s
```

Timings (median of 3, M4 Max, 900 files / ~72k LOC):

| Scenario | mypy 1.16.1 | ty 0.0.72 |
|---|---|---|
| Cold, src+tests | 9.26 s | 0.39 s |
| Warm cache, src+tests | 2.70 s | 0.39 s (no cache needed) |

ty is doing strictly more work than the old config (which disabled 10 error codes and ran in mirrors-mypy's isolated env with no deps resolvable) while being ~7× faster than warm mypy and ~24× faster cold.

## Summary by CodeRabbit

* **Developer Experience**
  * Replaced mypy with the pinned ty type checker.
  * Added a `just typecheck` command for running type checks locally.
  * Integrated type checking into pre-commit hooks and macOS CI tests.
  * Improved type-checking support for dynamically provided framework modules.
* **Documentation**
  * Documented the type-checking workflow, migration baseline, and planned cleanup.
  * Added an unreleased changelog entry describing the migration.
  * Preserved intentional terms in typo-checking configuration.

<h3>Greptile Summary</h3>

The PR replaces mypy with pinned ty-based full-project type checking and integrates it into local development, pre-commit, and macOS CI.
- Adds ty configuration, an MLX escape-hatch stub, and a documented migration baseline.
- Adds `just typecheck` and runs ty after the macOS test suite.
- Updates the lockfile, pre-commit hook, developer documentation, changelog, and typo allowlist.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| pyproject.toml | Replaces mypy configuration with the pinned ty dependency, MLX stub path, and migration-baseline rule ignores. |
| .github/workflows/tests.yml | Adds type checking after tests in a macOS environment that already synchronizes all optional dependencies. |
| .pre-commit-config.yaml | Replaces the changed-files-only mypy hook with an isolated full-project ty hook. |
| typings/mlx/core/__init__.pyi | Adds an Any-based escape-hatch stub so ty can resolve the compiled MLX core package. |
| uv.lock | Locks ty 0.0.72 and records it in the development extra without changing advisory-matched package versions. |

</details>

```mermaid
flowchart LR
    P[pyproject.toml and uv.lock] --> S[uv sync --all-extras]
    S --> T[ty check]
    J[just typecheck] --> S
    C[macOS CI] --> S
    H[pre-commit ty hook] --> I[isolated uv check]
    Y[typings/mlx/core stubs] --> T
    Y --> I
```
</details>

<sub>Reviews (2): Last reviewed commit: ["review: isolate ty pre-commit hook, stub..."](https://github.com/mflux-community/mflux/commit/260a94d16b8f2a7344fab82d5ee7ad0429bebf43) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53549092)</sub>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,6 +74,9 @@ dev = [
     "pytest-timer>=1.0,<2.0",
     # Pinned exactly; keep in sync with .pre-commit-config.yaml and the CI lint job.
     "ruff==0.16.3",
+    # Pinned exactly; keep in sync with .pre-commit-config.yaml (astral-sh/ty-pre-commit rev)
+    # and the CI ty step. ty 0.0.72 (62ebbecc7 2026-08-14).
+    "ty==0.0.72",
     "mlx==0.32.0; sys_platform=='darwin'",
     "mlx[cuda13]==0.30.3; sys_platform=='linux'",
 ]
@@ -173,20 +176,22 @@ section-order = [
     "local-folder",
 ]
 
-[tool.mypy]
-disable_error_code = [
-    "annotation-unchecked",
-    "arg-type",
-    "assignment",
-    "attr-defined",
-    "import-untyped",
-    "index",
-    "union-attr",
-    "return",
-    "return-value",
-    "var-annotated",
-]
-error_summary = true
-ignore_missing_imports = true
-implicit_optional = true
-python_version = "3.13"
+[tool.ty.rules]
+# One-time migration baseline (mypy -> ty, ty 0.0.72): every rule below had
+# pre-existing violations when ty was adopted (counts as of 2026-08-14).
+# Plan: re-enable (delete the ignore line) and fix piecemeal, one rule at a time.
+unresolved-import = "ignore"             # 491 (mostly `mlx.core` not resolvable by ty)
+invalid-argument-type = "ignore"         # 174
+unresolved-attribute = "ignore"          # 39
+invalid-assignment = "ignore"            # 36
+not-subscriptable = "ignore"             # 33
+invalid-parameter-default = "ignore"     # 10
+invalid-type-form = "ignore"             # 9
+invalid-return-type = "ignore"           # 8
+too-many-positional-arguments = "ignore" # 7
+missing-argument = "ignore"              # 7
+call-non-callable = "ignore"             # 4
+not-iterable = "ignore"                  # 3
+no-matching-overload = "ignore"          # 3
+unsupported-operator = "ignore"          # 2
+unused-type-ignore-comment = "ignore"    # 2
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -176,11 +176,14 @@ section-order = [
     "local-folder",
 ]
 
+[tool.ty.environment]
+# typings/ holds an escape-hatch stub for mlx.core (compiled .so, no .pyi upstream).
+extra-paths = ["typings"]
+
 [tool.ty.rules]
 # One-time migration baseline (mypy -> ty, ty 0.0.72): every rule below had
 # pre-existing violations when ty was adopted (counts as of 2026-08-14).
 # Plan: re-enable (delete the ignore line) and fix piecemeal, one rule at a time.
-unresolved-import = "ignore"             # 491 (mostly `mlx.core` not resolvable by ty)
 invalid-argument-type = "ignore"         # 174
 unresolved-attribute = "ignore"          # 39
 invalid-assignment = "ignore"            # 36
--- a/.cursor/skills/mflux-dev-env/SKILL.md
+++ b/.cursor/skills/mflux-dev-env/SKILL.md
@@ -23,3 +23,9 @@ This repo expects macOS arm64 and prefers `uv` + justfile recipes.
 - Prefer `uv run ...` for running Python commands to ensure the correct environment.
 - When running tests, keep `MFLUX_PRESERVE_TEST_OUTPUT=1` enabled (the justfile test recipes already do this).
 
+## Type checking (ty)
+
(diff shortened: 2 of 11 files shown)


pyproject.toml
```diff
@@ -173,20 +176,22 @@ section-order = [
...
+invalid-parameter-default = "ignore"     # 10
+invalid-type-form = "ignore"             # 9
+invalid-return-type = "ignore"           # 8
+too-many-positional-arguments = "ignore" # 7
+missing-argument = "ignore"              # 7
+call-non-callable = "ignore"             # 4
+not-iterable = "ignore"                  # 3
+no-matching-overload = "ignore"          # 3
+unsupported-operator = "ignore"          # 2
+unused-type-ignore-comment = "ignore"    # 2
```
anthonywu, line 200: I excluded these violations to let the pre-existing behavior "pass" without scope creeping this tool PR to also spray a bunch of diffs.

In the future, a single PR can be raised to address each rule here for ease of human review.

plz12345: WOuld suggest one PR for all the ones with under 5 hits, 1 for each of the other rules, just to reduce that churn a bit.  Even that is 11 PR's...

plz12345 (changes requested):
Requesting changes on the pre-commit hook; two smaller notes.

**1. The `ty` hook rewrites `uv.lock` on every commit.** The hook's entry is `uv check …`, not `ty`, and uv locks + syncs by default — [upstream's README](https://github.com/astral-sh/ty-pre-commit#customizing-side-effects-of-the-hook) calls this out. With `always_run: true` and `--quiet`, any commit (including ones touching no Python) can silently re-render `uv.lock` using the hook's bundled uv rather than the `0.12.4` we pin in CI for reproducible rendering. That shows up as an unexpected `uv.lock` diff in someone's PR, or a `uv sync --locked` failure. Documented fix:

```yaml
- id: ty
  args: [--isolated, --no-python-downloads]
```

**2. `unresolved-import` doesn't belong on the burn-down list.** 469 of the 491 are `mlx.core`, which ships as a compiled `.so` with no `.pyi` — unresolvable regardless of what we fix, so that ignore is permanent rather than baseline. It also silences the other 22, which cover the lazy in-function `torch`/`huggingface_hub` imports the fast suite never executes — exactly the breakage this rule is for. Suggest a minimal `typings/mlx/core.pyi` (`def __getattr__(name: str) -> Any: ...`) via `extra-paths`, and drop the project-wide ignore.

**3. Step order.** `ty` runs ahead of the only pytest step in the macOS job, so a typing regression from an unrelated dep bump takes all test signal with it. Moving it after pytest (or `continue-on-error: true`) keeps both.

pyproject.toml
anthonywu: agree on batching low medium counts.

I like to make each very easy to review. If a reviewer can approve a PR in < 5 minutes, I consider it a successful change =)

plz12345 (approved):

PR #668, opened 2026-08-25
fix: baking a LoRA into a sub-8-bit quantized model dropped the adapter

### LINKED ISSUE #665: LoRA baking makes LoRAs completely ineffective
Keeping LoRA baking makes the selected LoRA(s) completely ineffective (or ignored at some level) even though logging indicates that they were loaded just fine. Running the same gen with baking turned off results in a very different image with LoRAs properly applied.

Tested on a locally-saved Krea 2 4-bit model.

Baking is ON by default in the API (in the CLI too?) which could make first-time users think LoRAs don't work at all in mflux.

fxd0h:
## What

Fixes #665. Baking a LoRA into a quantized model requantized the merged weight at the base layer's own precision, and below 8 bits the quantization step is coarser than the delta being folded: on Krea 2 at q4 the group step measures ~12x the darkbrush adapter's delta rms, 80% of the codes come back identical, and what survives correlates with the intended delta at only 0.22. The result generated as if the adapter were never applied, while the logs reported success. No fold order can rescue this: `quantize(W + delta)` at 4 bits cannot represent a sub-step delta at all.

Sub-8-bit layers now requantize at q8 when a delta is folded in, the same escape `_bake_delta_into_linear` already takes for fp8 bases. The per-layer loader reconstructs mixed saves from stored shapes since #607, so checkpoints written by `mflux-save` from a q4 base with a LoRA load correctly. One console line reports how many layers moved. Bases at 8 bits keep their exact precision, group size and mode, as before.

- [x] Tests added/updated run in CI by default: three tests in `tests/lora/test_bake_low_bit_quantized.py`, sized to the real failure (delta between the q8 and q4 step). Two of them fail on the previous code.
- [x] `CHANGELOG.md`: entry under `Unreleased` (added in a follow-up commit referencing this PR number).
- [ ] Docs updated where behavior changed.
- [ ] New model.
- [ ] New/changed CLI.

## Verification

On an M5, `mflux-generate-krea2 -q 4 --lora krea/Krea-2-LoRA-darkbrush 1.0`, seed 42, 8 steps, 768px. Mean absolute pixel difference:

| | vs base (no LoRA) | vs runtime adapters (`--no-bake-lora`) |
|---|---|---|
| baked, before | 5.30 | 28.88 |
| baked, after | 26.55 | 7.62 |

Before, the baked image is the base image; after, it matches the runtime-adapter reference up to q8 rounding. 263 of the adapter's 264 target layers were upgraded (the last one is unquantized).

| runtime adapters (reference) | baked before | baked after |
|---|---|---|
| ref | before | after |

Full suite with the CI selector: 1469 passed.

The PR preserves baked LoRA deltas by upgrading sub-8-bit quantized layers to q8 while leaving q8 layers at their existing precision.
- Tracks and reports the number of upgraded layers during baking.
- Adds regression coverage for q4-to-q8 baking, q8 preservation, and the upgrade diagnostic.
- Documents the fix under the unreleased changelog.
</details>

The PR appears safe to merge with no actionable defects identified in the changed paths.

The q8 requantization path is compatible with currently supported LoRA-enabled models, preserves the adapter signal tested by the regression suite, and retains existing behavior for q8 bases.

| Filename | Overview |
|----------|----------|
| src/mflux/models/common/lora/mapping/lora_saver.py | Adds the q8 escape path for sub-8-bit LoRA and LoKr folds and reports upgraded layers; no currently reachable correctness issue was identified. |
| tests/lora/test_bake_low_bit_quantized.py | Covers preservation of small adapter deltas, unchanged q8 precision, and upgrade reporting. |
| CHANGELOG.md | Accurately describes the sub-8-bit baking failure and the new q8 behavior. |

</details>

<sub>Reviews (1): Last reviewed commit: ["docs: changelog entry for #668"](https://github.com/mflux-community/mflux/commit/770c296859c5253095be683cd18ef7069fd9e914) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56553792)</sub>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed LoRA baking for models quantized below 8 bits, preserving adapter changes that were previously lost.
  - Sub-8-bit layers are now upgraded to q8 when needed during baking.
  - Applies to runtime baking and saving quantized models with LoRA adapters.
  - Added reporting of how many layers were upgraded.


ianscrivener (approved):
<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/mflux-community/mflux/pull/668)</sub>

PR #672, opened 2026-08-25
training: record the free per-step loss, batch metric behind an optional plot_frequency

### LINKED ISSUE #671: LoRA loss monitoring can substantially increase training time vs mflux-cv
I switched an existing Krea 2 Raw LoRA training workflow from `mflux-cv` 0.18.30 to `mflux` 0.19.0 and noticed a large performance difference with otherwise equivalent settings.

With `mflux-cv`, loss was recorded for every training iteration without an obvious performance issue. The test run was estimated at roughly 80 hours.

With `mflux` 0.19.0 and `plot_frequency: 1`, the same run was estimated at roughly 232 hours. Based on a quick agent-assisted analysis of the two codebases, each loss update now appears to calculate an averaged loss across a validation batch rather than use the "free" values obtained after each iteration. That work is much more expensive than recording the current training-step loss, especially when repeated after every iteration as occurs when plot_frequency is 1.

I'm not sure whether this is:

- Intentional behavior, with `plot_frequency` now expected to be much higher (What is the user expected to set that to, reasonably? Why does it vary by model in the examples?)
- A useful `mflux-cv` behavior that was not carried into the community version (the mflux-cv approach was noisy and maybe not super representative of the results, but it was fast)
- A performance issue in the current averaged-loss implementation (seems unlikely but maybe)

Another side-effect of making the plot_frequency higher is that it takes longer before we get a loss.html file for display.

The averaged validation loss may be more representative, so I'm not necessarily suggesting reverting it. I just wanted to flag the unexpectedly large training-time impact and get some input into what the situation is WRT LoRA training loss calculation right now. Particularly since some official examples use `plot_frequency: 1` and that has a major training time impact.

fxd0h:
## What

Fixes #671, with the shape agreed there: the training loss the step already computed is now recorded every step for free, the batch metric became a second series behind `plot_frequency`, and `plot_frequency` is optional with a default of 20.

- The loop captured its loss and discarded it unrecorded, then paid up to 10 extra forwards per plot tick for a smoothed re-read of training samples. The free value now feeds the primary series, every step, and `loss.html` refreshes per step.
- The batch metric stays for whoever wants the smoother curve: same computation, its own series, rendered as a second line with a toggle, at `plot_frequency` ticks. The initial pre-training point stays too.
- `plot_frequency` was required; it now defaults to 20 (both ernie examples already shipped that). The z-image and flux2 examples shipped 1 and are now 20.
- Checkpoint statistics moved from a flat list to `{"step_loss": [...], "batch_loss": [...]}`. A legacy checkpoint loads its single series as the batch metric, which is what those entries were.

- [x] Tests added/updated run in CI by default: statistics save/load for both series, legacy checkpoint load through `from_spec`, and the `plot_frequency` default and explicit-value paths.
- [x] `CHANGELOG.md`: entry under `Unreleased` (follow-up commit referencing this PR number).
- [x] Docs updated where behavior changed: the three example configs.
- [ ] New model.
- [ ] New/changed CLI.

## Verification

Krea 2 Raw QLoRA on an M5 Max (q8 base, 512px, 10-image dataset, rank 16, 10 steps, same seed):

| | s/step |
|---|---|
| before, plotting effectively off | 5.41 |
| before, `plot_frequency: 1` | 17.47 |
| after, `plot_frequency` omitted (default 20) | 5.43 |

The per-step recording and per-step `loss.html` refresh cost nothing measurable. Inspected the run's artifacts: `loss.html` carries the 10-point step series plus the initial batch point, and the checkpoint's loss file has both keys.

Full suite with the CI selector: 1475 passed.

This PR separates training statistics into per-step and periodic batch-loss series while making `plot_frequency` optional with a default of 20.
- Records the already-computed training loss after every finite monitored step.
- Preserves periodic batch-loss measurements and legacy checkpoint compatibility.
- Updates the HTML plot to render and toggle both series.
- Updates example configurations, documentation, and statistics tests.
</details>

The PR appears safe to merge, with no concrete changed-code failure identified.

The new metric lifecycle follows existing micro-batch step semantics, preserves legacy checkpoint data in the appropriate batch series, and handles empty or mixed series in the plotter.

| Filename | Overview |
|----------|----------|
| src/mflux/models/common/training/trainer.py | Records each finite micro-batch loss and retains the frequency-gated batch metric without introducing an additional MLX synchronization. |
| src/mflux/models/common/training/statistics/statistics.py | Introduces separate persisted step and batch series while mapping legacy flat checkpoints to the batch series. |
| src/mflux/models/common/training/statistics/plotter.py | Supports plotting either or both loss series and adds a batch-series visibility toggle. |
| src/mflux/models/common/training/state/training_spec.py | Defaults omitted monitoring plot frequency to 20 while retaining positive-value validation. |
| tests/training/test_loss_series.py | Covers serialization, legacy loading, and default or explicit plot-frequency behavior. |

</details>

```mermaid
flowchart TD
    A[Training micro-batch] --> B[Compute loss and gradients]
    B --> C{Loss finite?}
    C -- No --> D[Reset accumulation and skip]
    C -- Yes --> E[Record step-loss point]
    E --> F{plot_frequency reached?}
    F -- Yes --> G[Compute and record batch-loss point]
    F -- No --> H[Refresh loss plot]
    G --> H
    H --> I{Checkpoint due?}
    I -- Yes --> J[Save both loss series]
    I -- No --> A
    J --> A
```
</details>

<sub>Reviews (1): Last reviewed commit: ["training: record the free per-step loss,..."](https://github.com/mflux-community/mflux/commit/f2e78519be4ecb3ffd527e8a781f6e18c7cb702e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=56852500)</sub>

## Summary by CodeRabbit

* **New Features**
  * Training charts now distinguish step loss from batch loss, with an optional Batch visibility toggle.
  * Monitoring plot frequency defaults to every 20 steps and can be customized.
  * Existing training checkpoints remain compatible with the updated loss format.

* **Bug Fixes**
  * Reduced the overhead of training-loss monitoring by limiting plot generation frequency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### ✨ Improvements
+
+- **Training loss monitoring no longer taxes the run**: the loop computed a loss every step and discarded it, then paid up to 10 extra forward passes per plot tick for a smoothed re-read of training samples; with `plot_frequency: 1` that tripled wall time (5.4 vs 17.5 s/step on a Krea 2 Raw QLoRA, the 80 vs 232 hour run of #671). The already-computed value now feeds the loss curve every step for free, the batch metric became an optional second series at `plot_frequency` ticks, and `plot_frequency` is optional with a default of 20. Checkpoint loss statistics carry both series; a legacy checkpoint's single series loads as the batch metric it was. (#671, #672)
+
 ## [0.19.1] - 2026-08-20
 
 ### 🔒 Security
--- a/src/mflux/models/common/training/_example/train.json
+++ b/src/mflux/models/common/training/_example/train.json
@@ -24,7 +24,7 @@
   "monitoring": {
     "preview_width": 1280,
     "preview_height": 720,
-    "plot_frequency": 1,
+    "plot_frequency": 20,
     "generate_image_frequency": 15
   },
   "lora_layers": {
--- a/src/mflux/models/common/training/state/training_spec.py
+++ b/src/mflux/models/common/training/state/training_spec.py
@@ -205,7 +205,9 @@ class MonitoringSpec:
         if preview_width <= 0 or preview_height <= 0:
             raise ValueError("Monitoring preview_width/preview_height must be > 0")
 
-        plot_frequency = int(param["plot_frequency"])
+        # Optional since #671: 20 keeps the batch metric's extra forwards at ~10-15% of
+        # wall time, against ~3x at 1. The free per-step series records regardless.
+        plot_frequency = int(param.get("plot_frequency", 20))
         generate_image_frequency = int(param["generate_image_frequency"])
         if plot_frequency <= 0:
             raise ValueError("Monitoring plot_frequency must be > 0")
--- a/src/mflux/models/common/training/statistics/plotter.py
+++ b/src/mflux/models/common/training/statistics/plotter.py
@@ -13,7 +13,7 @@ class Plotter:
     @staticmethod
     def update_loss_plot(training_spec: TrainingSpec, training_state: TrainingState, target_loss: float = 0.3) -> None:
         stats = training_state.statistics
-        if not stats.steps or not stats.losses:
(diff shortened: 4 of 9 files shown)


ianscrivener (approved):
<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/mflux-community/mflux/pull/672)</sub>

PR #690, opened 2026-08-27
fix: SeedVR2 upscaling crashes on mlx 0.32.2

tduarte:
## What

SeedVR2 upscaling crashes before its first step on **mlx 0.32.2**, which is inside the `mlx>=0.32.0,<0.33.0` range mflux declares:

```
TypeError: repeat(): incompatible function arguments. The following argument types are supported:
    1. repeat(array: array, repeats: int, axis: int | None = None, *, stream: StreamOrDevice = None) -> array

Invoked with types: mlx.core.array, mlx.core.array, kwargs = { axis: int }
```

`MMAttention` expands per-batch-element values to per-window ones. `WindowPartitioner.window_counts` is a `list[int]` with one entry per batch element (how many windows that element was partitioned into), and four call sites in `seedvr2_transformer/attention.py` fed it to `mx.repeat` as `mx.array(counts)`:

| line (v.0.19.1) | call |
| --- | --- |
| 80 | `txt_shape=mx.repeat(txt_shape, mx.array(counts), axis=0)` |
| 99 | `txt_len[mx.repeat(mx.arange(len(counts)), mx.array(counts))]` |
| 121 | `mx.repeat(txt, mx.array(counts), axis=0)` — the one that throws first |
| 132 | `mx.repeat(mx.arange(len(txt_len)), mx.array(counts))` |

`mx.repeat` declares `repeats: int`. Up to and including mlx 0.32.1 pybind11 coerced a **one-element** `mx.array` to that int, so the calls happened to work; mlx 0.32.2 stopped coercing and they became a hard `TypeError`. Measured directly, with `mx.repeat(mx.array([[1, 2], [3, 4]]), reps, axis=0)`:

| mlx | `reps = mx.array([3])` | `reps = mx.array([2, 3])` |
| --- | --- | --- |
| 0.31.2 | accepted | TypeError |
| 0.32.0 | accepted | TypeError |
| 0.32.1 | accepted | TypeError |
| 0.32.2 | **TypeError** | TypeError |

Two things follow from the right-hand column. First, the code was **never** correct for more than one batch element — `window_counts` longer than one has raised this same `TypeError` on every mlx version tested, on the path it was written for. A batch of one was hiding it, so this is a latent correctness bug and not only an mlx-0.32.2 compatibility patch. Second, even where the coercion worked it was the wrong operation: `mx.repeat(x, 3, axis=0)` applies one count to every row, whereas the intent is `np.repeat(x, counts, axis=0)` — a different count per row.

CI never saw any of it because the dev group pins `mlx==0.32.0`, one release below the boundary, while the published package accepts 0.32.2.

**The fix.** A per-window batch index is built once per attention call and the four sites gather through it, which is exactly `np.repeat(x, counts, axis=0)` and uses only the documented scalar-`repeats` signature:

```python
@staticmethod
def _window_to_batch(counts: list[int]) -> mx.array:
    return mx.concatenate([mx.full((count,), index, dtype=mx.int32) for index, count in enumerate(counts)])
```

`_repeat_text_for_windows` and `_unconcat_and_coalesce` take that index instead of recomputing it (`_unconcat_and_coalesce` still takes `counts` for its text mean-pool loop). The mlx floor is unchanged; nothing here needs 0.32.

A repo-wide grep for the pattern found no other instance — every other `mx.repeat` in `src/` already passes a Python int.

- [x] Tests added/updated run in CI by default — both new tests are `@pytest.mark.fast`, no weights, ~0.03 s total.
- [x] `ruff check` and `ruff format` are clean (`pre-commit run` on the changed files passes ruff, typos, and ty).
- [x] `CHANGELOG.md`: entry under `Unreleased` referencing this PR number.
- [x] Docs updated where behavior changed — no user-facing behavior or flag changed; output for a single batch element is byte-identical.
- [x] New model: n/a.
- [x] New/changed CLI: n/a.

## Verification

**Project suites**, on the repo's pinned `mlx==0.32.0`:

```
just lint         -> All checks passed!
just typecheck    -> All checks passed!
just test-fast    -> 1326 passed, 6 skipped, 223 deselected
just test         -> 1487 passed, 6 skipped, 62 deselected
```

`just test` also run under `mlx==0.32.2` in a scratch venv: **1487 passed, 6 skipped, 62 deselected**.

**New regression test** — `tests/image_generation/test_seedvr2_window_attention.py`:

- `test_window_to_batch_expands_uneven_counts_like_numpy_repeat` — `counts = [2, 3]` checked against `np.repeat`. This is the case that was broken on every mlx version, and it needs no weights.
- `test_mm_attention_runs_with_more_than_one_batch_element` — a real `MMAttention` forward with two batch elements (so `window_counts == [2, 2]`), reaching all four call sites. It asserts more than "did not crash": windows never span batch elements, so the batched result must equal the two single-element calls it is made of, which is what proves the gather lines each window up with its own text rather than merely producing an array of the right shape.

**The test was planted against the old implementation and watched to fail** (on `mlx==0.32.0`, where the released code otherwise runs):

```
src/.../attention.py:121: in _repeat_text_for_windows
>   return mx.repeat(txt, mx.array(counts), axis=0).reshape(-1, *txt.shape[2:])
E   TypeError: repeat(): incompatible function arguments.
```

— i.e. the reported error, reproduced by a unit test on a machine with no SeedVR2 weights.

**Real end-to-end upscale**, `SeedVR2` 3B (`numz/SeedVR2_comfyUI`), unquantized, `2x` on a 256x256 source, weights offline (`HF_HUB_OFFLINE=1`):

| run | result |
| --- | --- |
| mlx 0.32.2, code as released | `TypeError` in `_repeat_text_for_windows`, before step 1 |
| mlx 0.32.2, patched | `(512, 512)` |
| mlx 0.32.1, patched | `(512, 512)` |
| mlx 0.32.1, code as released | `(512, 512)` |

All three successful outputs are **byte-identical PNGs** (`md5 bbf35005d76cf29b2a154caaf5c3be02`), so the patch changes no pixels on the single-batch path that ships today. A unit-level comparison of old vs new `MMAttention` on one batch element is bit-identical across `shift` and `rope_on_text` in both settings as well.

**Older mlx**: the two new tests also pass on `mlx==0.31.2` (forced into a scratch venv, below the package's declared floor) — the fix relies on nothing newer than the scalar-`repeats` signature.

**Not checked**: `-m slow` / golden-image suites and the Linux (`mlx[cuda13]`) wheel were not run.

<h3>Greptile Summary</h3>

This PR replaces SeedVR2 attention’s unsupported array-valued `mx.repeat` calls with a reusable window-to-batch gather index, restoring compatibility with MLX 0.32.2 and correct multi-element batch handling.
- Uses the index consistently for text tensors, text shapes, window lengths, and output coalescing.
- Adds fast regression coverage for uneven counts and batched-versus-single attention equivalence.
- Adds an Unreleased changelog entry, although its pull request reference remains a placeholder.

<h3>Confidence Score: 4/5</h3>

The implementation appears safe to merge after replacing the changelog’s `#PRNUM` placeholder with `#690`.

The attention gather follows the partitioner’s batch-major window ordering and is covered by multi-batch regression tests; the only accepted issue is an incorrect changelog reference.

**Files Needing Attention:** CHANGELOG.md

| Filename | Overview |
|----------|----------|
| src/mflux/models/seedvr2/model/seedvr2_transformer/attention.py | Replaces invalid variable-count repeats with an int32 batch-index gather whose ordering matches the partitioner and coalescing logic. |
| tests/image_generation/test_seedvr2_window_attention.py | Adds focused regression tests for gather semantics and consistency between batched and per-element attention calls. |
| CHANGELOG.md | Documents the SeedVR2 correction but leaves `#PRNUM` in place of the actual pull request number. |

</details>

```mermaid
flowchart LR
  A[Per-batch text values] --> B[Build window-to-batch index]
  C[Batch-major video windows] --> D[Gather matching text per window]
  B --> D
  D --> E[Run attention per video/text window]
  E --> F[Split video and text outputs]
  F --> G[Reverse video partition]
  F --> H[Mean-pool text windows per batch]
```
</details>

<details><summary>Prompt To Fix All With AI</summary>

`````markdown
### Issue 1
CHANGELOG.md:12
**Changelog retains PR placeholder**

The new entry displays `#PRNUM` instead of `#690`, so readers cannot use the changelog's intended pull request reference.

```suggestion
- **SeedVR2 upscaling runs on mlx 0.32.2, and its window attention is correct for more than one batch element**: the window attention expanded per-batch values to per-window ones with `mx.repeat(x, mx.array(window_counts), axis=0)`, but `mx.repeat` only takes a scalar repeat count. Up to mlx 0.32.1 pybind11 coerced a *one-element* array to an int, so a single image happened to work; mlx 0.32.2 dropped that coercion and every `mflux-upscale-seedvr2` run now dies before the first step with `TypeError: repeat(): incompatible function arguments`, inside the `mlx>=0.32.0,<0.33.0` range mflux itself declares (the dev group pins 0.32.0, which is why CI never saw it). A second batch element made `window_counts` longer than one and raised the same TypeError on *every* mlx version tested back to 0.31.2, so this was always a latent bug that a batch of one was hiding. All four call sites now gather with a window→batch index built once per attention call, which is `np.repeat(x, counts, axis=0)` semantics and needs no coercion; output for a single batch element is bit-identical to before. (#690)
```

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: [e8f7e64](https://github.com/mflux-community/mflux/commit/e8f7e649fd72e896478a2d7740aaf3c3bde5e9ce) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57734809)</sub>

> Greptile also left **1 inline comment** on this PR.

**Context used:**

- Knowledge Base — [Restoration and depth workflows](https://app.greptile.com/mflux-community/-/custom-context/knowledge-base/mflux-community/mflux/-/docs/restoration-and-depth-workflows.md)

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed SeedVR2 upscaling failures caused by window-attention processing.
  - Restored support for multi-image and batched upscaling.
  - Improved handling of uneven attention windows while preserving existing single-image output behavior.

- **Tests**
  - Added coverage for batched window attention and per-image text processing.


fxd0h (approved):
Checked the mlx claim directly: on 0.32.0 `mx.repeat(x, mx.array([2]))` works and `mx.array([2, 3])` raises the TypeError; on 0.32.2 both raise. Matches the description.

Upscale A/B on the M5, seedvr2-3b q8, 256 to 512, seed 3: this branch vs main is bit-identical (max abs diff 0). Batching two images through MMAttention on this branch gives the same pixels for the shared image as the single run, so the gather does line each window up with its own text.

One precision on the second half of the claim: `mflux-upscale-seedvr2` loops over `--image-path` one image at a time, so the multi-batch path was never reachable from the CLI. The fix there is right and the test pins it, it just is not something a user hit.

The CHANGELOG entry may collide with #685, which moves release notes into a fenced block in the PR body; if that merges first this PR will need the block instead of the changelog line. Not a blocker either way. Approving.

PR #592, opened 2026-08-14
fix: derive --model choices from AVAILABLE_MODELS instead of a stale list

plz12345:
## What

`--model`/`--base-model` accepted names from a hand-maintained list (`ui_defaults.MODEL_CHOICES`) that had drifted from `AVAILABLE_MODELS`, so 16 canonical keys and ~40 aliases were treated as local paths or rejected outright. The choices now derive from the registry, and `--base-model` is validated against `ConfigResolution.base_model_names()` — the same list its explicit-base rule uses — instead of a parallel argparse `choices=`. Validation runs post-parse, so a `base_model` restored from a metadata sidecar is checked too.

Fixes #591. First checkbox of #577.

## Checklist (definition of done)

- [x] Tests added/updated run in CI by default (`tests/arg_parser/test_model_choices.py`, all `@pytest.mark.fast`).
- [x] `ruff check` and `ruff format` are clean.
- [x] `CHANGELOG.md`: entry under `Unreleased`.
- [x] Docs updated where behavior changed — `--model` help now lists the canonical keys; the model-porting skill no longer tells porters to hand-edit `MODEL_CHOICES`.
- [ ] New model: n/a.
- [ ] New/changed CLI: n/a — no options added or dropped.

## Verification

Before:

```
$ mflux-generate-lens --model lens-turbo --prompt test
FileNotFoundError: Model not found: 'lens-turbo'

$ mflux-generate-qwen --model some-org/some-finetune --base-model qwen-image --prompt test
mflux-generate-qwen: error: argument --base-model: invalid choice: 'qwen-image'
```

After: both parse. All 84 registry spellings sweep clean — `model_path is None` and `ModelConfig.from_name` resolves — while `some-org/some-finetune` and `/local/path` still pass through as paths. That sweep is the new test file.

`make test` (default selector): 1178 passed, 3 skipped. `ruff check` / `ruff format --check`: clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CLI model-name validation using the model registry.
  - Added support for model aliases and built-in repository identifiers.
  - Ensured restored metadata values receive the same base-model validation.
  - Preserved support for unknown model paths while preventing invalid model names.

- **Documentation**
  - Updated the unreleased changelog with model-validation improvements.
  - Clarified model registry wiring requirements.

<h3>Greptile Summary</h3>

The PR replaces stale CLI model-name lists with values derived from `AVAILABLE_MODELS` and aligns post-parse `--base-model` validation with `ConfigResolution`.

[...]

PR #696, opened 2026-08-31
fix: --config-from-metadata replays runs made without --base-model again

### LINKED ISSUE #695: --config-from-metadata rejects every sidecar written without --base-model ("invalid choice: 'None'")
## What

Replaying a run from its own sidecar fails on 0.19.x whenever the run had no `--base-model`, which is nearly every run with a builtin model:

```
mflux-generate-flux2 --prompt "a cup" --width 256 --height 256 --steps 1 --seed 1 --metadata --output run.png
mflux-generate-flux2 -C run.metadata.json --output replay.png
mflux-generate-flux2: error: argument --base-model: invalid choice: 'None' (choose from krea-2, ...)
```

The sidecar has always written `"base_model": str(self.model_config.base_model)`, so a builtin run stores the literal string `"None"`. The restore path in `parse_args` copies it into `namespace.base_model` as is. Until #592 that was harmless: a builtin name hits the exact-match rule before the explicit-base rule ever looks at it. #592 validates `--base-model` against `ConfigResolution.base_model_names()` before resolution, and the string `"None"` is not a name, so the replay exits 2.

## Fix

Two sides, since sidecars with the string are already out there:

- the writer stores the value itself (`null` for a builtin run, the base repo id for a custom checkpoint, both of which the validator accepts);
- the restore treats `null` and the literal `"None"` as "nothing to restore".

Plus a test that writes a builtin sidecar and replays it through the parser. Taking this one.

fxd0h:
## What

Replaying a run from its own sidecar exits 2 on 0.19.x whenever the run had no `--base-model`, which is nearly every run with a builtin model: `argument --base-model: invalid choice: 'None'`. The sidecar stored `str(None)`, the literal string `"None"`, and since #592 validates `--base-model` against the resolver's list before resolution, the restore handed that string to the validator.

Two sides, since sidecars with the string are already on disk: the writer stores the value itself (`null` for a builtin run, the base a custom checkpoint resolved to), and the restore treats `null` and the legacy `"None"` alike. Fixes #695.

- [x] Tests added/updated run in CI by default: a parser test replaying a builtin sidecar with `null` and with the legacy `"None"` (the legacy case fails without the fix), and a writer test pinning `null` for a builtin run.
- [x] `CHANGELOG.md`: entry under `Unreleased` referencing this PR's issue.
- [n/a] Docs updated where behavior changed.
- [n/a] New model.
- [n/a] New/changed CLI options.

## Verification

On the M5, flux2-klein-4b, 256x256, 1 step, seed 1:

```
mflux-generate-flux2 --prompt "a cup" --width 256 --height 256 --steps 1 --seed 1 --metadata --output run.png
mflux-generate-flux2 -C run.metadata.json --output replay.png
```

On main the second command exits 2 with the message above. On this branch a sidecar written by main (with the string `"None"`) replays and the image is pixel-identical to the original (max abs diff 0); a sidecar written by this branch stores `"base_model": null` and replays identically too.

Full fast suite: 1494 passed, `ty` clean.

## Release note

```release-note
Replaying a run with --config-from-metadata works again for runs made without --base-model; sidecars written by earlier versions replay too.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed replaying configurations from metadata for runs without a base model.
  * Existing metadata containing the legacy `"None"` value now replays successfully.
  * New metadata correctly records an unset base model as `null`.
  * Fixed SeedVR2 window attention for multi-batch inputs and newer MLX versions.


ianscrivener (approved):
<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/mflux-community/mflux/pull/696)</sub>

PR #711, opened 2026-09-04
feat: opt-in gradient_checkpointing for training, with the backward evaluated last-block-first

qruz-hq:
## What

Edit-LoRA training for FLUX.2 klein is killed by macOS at the very first step on a 36 GB machine, whatever `max_resolution`, `quantize` or `low_ram` is set to: the trainer keeps every transformer block's activations for the backward pass, and for an edit pair (target + reference tokens) on klein-base-4B that is more than the machine has. This PR adds an opt-in `"gradient_checkpointing": true` training-config flag for FLUX.2, Z-Image and Ernie (the mechanism Krea 2 already uses internally) and, because checkpointing alone does not help in MLX unless the gradients are evaluated in the right order, a trainer change that evaluates them last-block-first whenever checkpointing is active. With both, a klein-base-4B edit run that peaked above 50 GB runs on an M5 Max with 36 GB.

## Why checkpointing alone is not enough

MLX builds the backward lazily, and `nn.utils.checkpoint` turns each block's vjp into "recompute the block from its saved input, then differentiate". The trainer used to call `mx.eval` on the whole gradient tree at once, which walks it in parameter order: block 1's gradients first. Producing them needs the cotangent at block 1's output, so every later block is recomputed on the way there, and those recomputations stay referenced until their own gradients are collected further down the tree. The result is that most recomputed activations are alive at the same time and the peak ends up close to the un-checkpointed number.

`TrainingTrainer._evaluate_backward_in_block_order` evaluates the loss first, then the gradients one block at a time from the last block backwards, so only a couple of blocks' recomputed activations exist at any moment. The block is read off the parameter path (the first numeric path segment, e.g. `transformer.single_transformer_blocks.7`), which follows the forward order for all four transformers that can train.

Measured on a small `Flux2Transformer` with LoRA layers (same seed, loss and gradient norm identical to six decimals in every mode):

| blocks (double + single), image tokens | stock | checkpointing, one `mx.eval` in tree order | checkpointing + last-block-first |
|---|---|---|---|
| 2 + 12, 512 (CPU build) | 194 MB | 106 MB | **69 MB** |
| 2 + 12, 512 (Metal, MLX 0.32.0) | 381 MB | 280 MB | **191 MB** |
| 2 + 12, 2048 (Metal) | 2523 MB | 1604 MB | **1250 MB** |
| 5 + 20, 1024 (Metal) | 1318 MB | 715 MB | **468 MB** |

A single `mx.eval` over the gradient leaves in reverse order does not help (312 / 2042 / 1000 MB on the three Metal rows, i.e. worse than tree order), so the per-block `mx.eval` calls are the mechanism, not a stylistic choice. The saving grows with block count and sequence length: on a 26-block variant with 2k tokens the ordered backward reached 13% of the stock peak.

Real run, M5 Max 36 GB, `flux2-klein-base-4b`, edit mode, 12 pairs, target 768×992, reference 576×752, `quantize: 8`, `low_ram: true`, rank 32:

- stock: first step climbs 35 → 47 → 50+ GB, `zsh: killed`
- checkpointing only (this patch minus the ordering): identical, killed at 50+ GB
- checkpointing + last-block-first: runs. Steps were about 30% slower on that run; the per-block evaluation also adds one GPU sync per block, which is a fixed cost that is only visible on toy-sized models.

## Changes

- `TrainingSpec.gradient_checkpointing` (default `false`), parsed from the training JSON and carried through checkpoints, so a resumed run keeps it.
- `TrainingAdapter.set_gradient_checkpointing(enabled)` on the protocol, implemented by every adapter. The FLUX.2, Z-Image and Ernie transformers gain the optional `gradient_checkpointing` attribute Krea 2's transformer already has and wrap each block in `nn.utils.checkpoint` when it is set. The attribute defaults to off, so inference is unaffected; during training previews the wrap is forward-only, exactly as Krea 2 runs today.
- Krea 2 forces checkpointing on in `freeze_base`, so its `set_gradient_checkpointing` keeps it on and the trainer reads the effective state back from the transformer. Krea 2 therefore gets the ordered backward too.
- `TrainingTrainer._evaluate_backward_in_block_order`, applied right after the train step whenever checkpointing is active. Gradient clipping, accumulation and the non-finite skip see fully materialized gradients, as before.
- Docs: the common training README explains what the flag does and when to turn it on; the FLUX.2 README config example gains the field and a tip for 32–36 GB Macs.

## Tests

`tests/training/test_gradient_checkpointing.py`, all marked `fast`, no downloads, CPU-runnable, on a small real `Flux2Transformer` with LoRA layers:

- identical loss (`rel=1e-4`) and, tensor by tensor, identical gradients (`rtol=1e-3`, `atol=1e-6`; the largest observed difference is 1e-6) with and without checkpointing;
- the ordered backward lowers peak memory below plain tree-order checkpointing (`< 0.8×`; 0.68× on Metal, 0.65× on the CPU build. The ratio depends on MLX's allocator, so if a future MLX moves it the threshold is the thing to revisit, not the trainer);
- block-prefix grouping of parameter paths, including non-block paths such as `proj_out`.

## Follow-ups (not in this PR)

- In edit training the `_in` reference is encoded at up to `MAX_REFERENCE_AREA` (1 MP) regardless of `max_resolution`, so lowering `max_resolution` reduces memory less than users expect.
- Training previews render at the preview image's native size (`_preview_dimensions` passes `max_resolution=None`), so a 1584×2048 `preview_1.jpg` makes every preview a 3 MP render.

Both will be filed as issues; a fix that makes `max_resolution` cover the reference and the previews is ready as a separate PR.

- The ordering effect probably applies to any MLX trainer that uses per-layer `mx.checkpoint` and evaluates a full gradient tree at once; mlx-lm may want to check.

- [x] Tests added/updated run in CI by default (selector: `-m "not slow and not high_memory_requirement"`, same as `just test`). Only mark `@pytest.mark.slow` or `@pytest.mark.high_memory_requirement` when a test exceeds CI's time or memory budget (typically weight downloads / image generation).
- [x] `ruff check` and `ruff format` are clean (`uv run ruff` uses the version pinned in the dev dependencies of `pyproject.toml`, which is the single source of truth for pre-commit and CI; `pre-commit run -a` covers it locally).
- [x] Release note block below filled in (or `none` for changes users never see).
- [x] Docs updated where behavior changed — README examples/table rows are part of the API contract (see `.cursor/rules/RULE.md`).
- [ ] New/changed CLI: not applicable (training config field only).

## Release note

```release-note
Training: new opt-in `"gradient_checkpointing": true` config flag for FLUX.2, Z-Image and Ernie LoRA training, and the backward pass is now evaluated last-block-first whenever checkpointing is active (Krea 2 included), so the memory saving actually materializes. FLUX.2 klein-base-4B edit training fits on 32–36 GB Macs with it.
```

## Verification

```sh
uv run ruff check && uv run ruff format --check          # clean
uv run pytest tests/training/test_gradient_checkpointing.py   # 3 passed (Metal, MLX 0.32.0)
uv run pytest tests/training -m "not slow and not high_memory_requirement"   # 243 passed
```

Real run: `mflux-train` on the klein-base-4B edit config above with `"gradient_checkpointing": true`, `"quantize": 8`, `"low_ram": true` trains on a 36 GB M5 Max; the same config without the flag is killed at step 0. The memory tables above come from a script that builds the test-sized transformer and compares `mx.get_peak_memory()` across the four evaluation strategies.

## Summary by CodeRabbit

* **New Features**
  * Added optional gradient checkpointing for training to reduce peak memory usage by recomputing activations during backpropagation.
  * Added a `gradient_checkpointing` training configuration option, disabled by default.
  * Supported across FLUX.2, Krea 2, ERNIE Image, and Z Image training workflows.
  * Added memory-conscious backward evaluation to further reduce training memory usage.

* **Documentation**
  * Added guidance for fitting training runs on 32–36 GB machines, including quantization recommendations and expected performance trade-offs.

### DIFF
--- a/src/mflux/models/common/README.md
+++ b/src/mflux/models/common/README.md
@@ -308,6 +308,7 @@ FLUX.1 uses `mflux-generate` with `--model dev` (or `schnell` / `krea-dev`) and 
 - Training supports normal (txt2img) and edit modes; the mode is auto-detected from your data folder layout (see examples below).
 - Expected layout: a config file plus a data folder with images and optional prompt files.
 - **Edit mode limitation**: Edit training currently supports only FLUX.2-klein-base models (`flux2-klein-base-4b` or `flux2-klein-base-9b`).
+- **Memory**: `"gradient_checkpointing": true` recomputes each transformer block's activations during the backward pass instead of keeping all of them, and evaluates the gradients last-block-first (the evaluation order matters in MLX; in tree order the recomputed blocks all stay alive and the saving disappears). Steps get ~30% slower and peak memory drops to roughly "weights + a couple of blocks". Turn it on when a run is killed at the first training step; it is what makes FLUX.2 klein edit training fit on 32–36 GB machines. Combine with `"quantize": 8` and `"low_ram": true` on the tightest machines.
 - To train a local copy of a model, you can specify the path to that model in the training spec JSON alongside the `model` using the `model_path` key.
 - For model-specific config fields, see the [Z-Image README](../z_image/README.md) and [FLUX.2 README](../flux2/README.md).
 
--- a/src/mflux/models/common/training/trainer.py
+++ b/src/mflux/models/common/training/trainer.py
@@ -2,13 +2,14 @@ from __future__ import annotations
 
 import gc
 import random
+import re
 import tempfile
 from pathlib import Path
 
 import mlx.core as mx
 from mlx import nn
 from mlx.optimizers import clip_grad_norm
-from mlx.utils import tree_map, tree_unflatten
+from mlx.utils import tree_flatten, tree_map, tree_unflatten
 from tqdm import tqdm
 
 from mflux.models.common.latent_creator.latent_creator import LatentCreator
@@ -24,6 +25,10 @@ from mflux.utils.exif_orientation import oriented_size
 
 
 class TrainingTrainer:
+    # Path prefix up to and including the first numeric segment, e.g.
+    # "transformer.single_transformer_blocks.7" from "...single_transformer_blocks.7.attn.to_out.lora_A".
+    _BLOCK_PREFIX = re.compile(r"^(.*?\.\d+)(?:\.|$)")
+
     @staticmethod
     def _sample_timestep_index(timestep_type: str, low: int, high: int, rng) -> int:
         # Map a [0,1) draw shaped by the distribution to a timestep index in [low, high).
@@ -141,6 +146,10 @@ class TrainingTrainer:
         # Freeze base weights and unfreeze LoRA weights
         adapter.freeze_base()
         TrainingTrainer._unfreeze_lora_layers(adapter.transformer())
+        if training_spec.gradient_checkpointing:
+            adapter.set_gradient_checkpointing(True)
+        # Krea 2 forces checkpointing on in freeze_base, so read the effective state back.
+        gradient_checkpointing = bool(getattr(adapter.transformer(), "gradient_checkpointing", False))
 
         train_step_function = nn.value_and_grad(
             model=adapter.model(),
@@ -169,6 +178,8 @@ class TrainingTrainer:
         nonfinite_skips = 0
         for batch in batches:
             loss, grads = train_step_function(batch)
+            if gradient_checkpointing:
+                TrainingTrainer._evaluate_backward_in_block_order(loss, grads)
             if not TrainingTrainer._step_is_finite(loss):
                 del loss, grads
                 nonfinite_skips += 1
@@ -228,6 +239,22 @@ class TrainingTrainer:
         if nonfinite_skips:
             print(f"Skipped {nonfinite_skips} non-finite (NaN/Inf) training step(s).")
         training_state.save(adapter, training_spec)
+
+    @staticmethod
+    def _evaluate_backward_in_block_order(loss: mx.array, grads) -> None:
+        # MLX is lazy, so the order the gradients are evaluated in decides how many recomputed
+        # block activations are alive at once. Evaluating them first-block-first (tree order)
+        # forces every later block to be recomputed and kept until its own gradient is collected,
+        # which silently undoes gradient checkpointing (peak memory ends up the same as without
+        # it). Evaluate the forward, then walk the blocks from the last one backwards so only a
+        # couple of blocks' activations exist at any time.
+        mx.eval(loss)
+        groups: dict[str, list[mx.array]] = {}
+        for path, grad in tree_flatten(grads):
+            block_prefix = TrainingTrainer._BLOCK_PREFIX.match(path)
+            groups.setdefault(block_prefix.group(1) if block_prefix else "", []).append(grad)
+        for key in reversed(list(groups)):
+            mx.eval(*groups[key])
 
     @staticmethod
     def _unfreeze_lora_layers(module: nn.Module) -> None:
--- a/src/mflux/models/common/training/adapters/base.py
+++ b/src/mflux/models/common/training/adapters/base.py
@@ -20,6 +20,8 @@ class TrainingAdapter(Protocol):
 
     def freeze_base(self) -> None: ...
 
+    def set_gradient_checkpointing(self, enabled: bool) -> None: ...
+
     def encode_data(
         self,
         *,
--- a/src/mflux/models/common/training/state/training_spec.py
+++ b/src/mflux/models/common/training/state/training_spec.py
@@ -260,6 +260,11 @@ class TrainingSpec:
     # to hold the largest transient allocations seen; bounding it returns freed buffers to
     # the OS at ~no speed cost (unlike low_ram's clear-every-step). None = MLX default.
     cache_limit_gb: float | None = None
+    # Recompute each transformer block's activations in the backward pass instead of keeping
+    # them, and evaluate the gradients last-block-first. Trades ~30% step time for a peak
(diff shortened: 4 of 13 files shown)


src/mflux/models/common/training/trainer.py
```diff
@@ -169,6 +178,8 @@ def train(
         nonfinite_skips = 0
         for batch in batches:
             loss, grads = train_step_function(batch)
+            if gradient_checkpointing:
+                TrainingTrainer._evaluate_backward_in_block_order(loss, grads)
             if not TrainingTrainer._step_is_finite(loss):
```
IonDen, line 183: suggestion (non-blocking): check the loss is finite before walking the backward

A step whose loss is NaN now computes every block's gradient and then discards it; before this change the skip threw the unevaluated backward away for free. Checking the loss first keeps the skip as cheap as it was, and clipping and accumulation still see fully evaluated gradients.

```suggestion
            loss, grads = train_step_function(batch)
            step_is_finite = TrainingTrainer._step_is_finite(loss)
            if step_is_finite and gradient_checkpointing:
                TrainingTrainer._evaluate_backward_in_block_order(loss, grads)
            if not step_is_finite:
```

src/mflux/models/common/training/trainer.py
```diff
@@ -229,6 +240,22 @@ def train(
...
+        # which silently undoes gradient checkpointing (peak memory ends up the same as without
+        # it). Evaluate the forward, then walk the blocks from the last one backwards so only a
+        # couple of blocks' activations exist at any time.
+        mx.eval(loss)
+        groups: dict[str, list[mx.array]] = {}
+        for path, grad in tree_flatten(grads):
+            block_prefix = TrainingTrainer._BLOCK_PREFIX.match(path)
+            groups.setdefault(block_prefix.group(1) if block_prefix else "", []).append(grad)
+        for key in reversed(list(groups)):
+            mx.eval(*groups[key])
```
IonDen, line 257: suggestion (non-blocking): say that the walk assumes parameter order is forward order

The groups are walked in reverse of the order the layers are defined in, which matches the order they run in for all four block stacks today. `Krea2Transformer` is the exception waiting to happen: it defines `txtfusion` after `blocks` but runs it first, so a LoRA target under `txtfusion.layers.N` would be evaluated first, pull the whole backward through every block at once, and quietly lose the memory saving. Nothing targets it today; a one-line comment stating the assumption, or restricting the grouping to the known block prefixes, keeps the next target list from tripping it.

src/mflux/models/common/README.md
```diff
@@ -308,6 +308,7 @@ FLUX.1 uses `mflux-generate` with `--model dev` (or `schnell` / `krea-dev`) and
 - Training supports normal (txt2img) and edit modes; the mode is auto-detected from your data folder layout (see examples below).
 - Expected layout: a config file plus a data folder with images and optional prompt files.
 - **Edit mode limitation**: Edit training currently supports only FLUX.2-klein-base models (`flux2-klein-base-4b` or `flux2-klein-base-9b`).
+- **Memory**: `"gradient_checkpointing": true` recomputes each transformer block's activations during the backward pass instead of keeping all of them, and evaluates the gradients last-block-first (the evaluation order matters in MLX; in tree order the recomputed blocks all stay alive and the saving disappears). Steps get ~30% slower and peak memory drops to roughly "weights + a couple of blocks". Turn it on when a run is killed at the first training step; it is what makes FLUX.2 klein edit training fit on 32–36 GB machines. Combine with `"quantize": 8` and `"low_ram": true` on the tightest machines.
```
IonDen, line 311: nitpick (non-blocking): "~30% slower" is one run, say so

The 30% comes from a single klein-base-4B edit run on an M5 Max, and the per-block sync is a fixed cost that weighs more on small models. "About a third slower on a klein-base-4B edit run" keeps the number honest for other models and machines.

IonDen (commented):
Merge as is once CI has run; nothing below blocks it. Checked on an M1 Max 32 GB with mlx 0.32.0: ruff is clean, the three new tests pass, tests/training under the CI selector gives 243 passed / 0 failed / 3 deselected, and checkpointing plus the new evaluation order uses 68% of the memory that checkpointing alone needs, matching the author's numbers.

Three findings inline, all non-blocking: two suggestions, NaN-step cost and block-order assumption, and one nitpick, single-run benchmark number. One note outside this diff: the Ernie README's example config could show `"gradient_checkpointing": false` next to `low_ram` like the FLUX.2 one now does.

The required lint, test and release-note workflows have not run on this commit; all four sit at "action required", waiting for someone to approve them for this contributor, so the only green check is CodeRabbit.

ianscrivener (approved):
<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/mflux-community/mflux/pull/711)</sub>

fxd0h (approved):
Measured on an M5 Max 128 GB: Z-Image Turbo q8, LoRA on q/k/v of all 30 layers, 6 steps at 1024 px. Peak memory footprint 63.3 GB with the flag off and 52.9 GB on; step time 5.1 s to 6.8 s, so about a third slower here as well. Per-step losses are identical to the last digit with the flag on, off, and on main. At 512 px there's no visible memory difference, the weights dominate, so the README's "turn it on when the first step is killed" is the right framing.

Fast suite: 1597 passed here, and CI is green now that the workflow runs were approved. IonDen's finite-loss check before the walk is worth taking; the rest reads fine to me.

PR #715, opened 2026-09-06
fix: evaluate Z-Image's RoPE tables at construction so a seed's output is stable across a process

### LINKED ISSUE #714: Z-Image: the RoPE tables are built lazily, and one eager touch between compiled runs changes the output
1. Load Z-Image Turbo (q8) once and generate twice in the same process with the same seed. The two images are identical, as expected. This is on a chip where mflux compiles `predict` (anything but a base or Pro M1/M2; an M1 Max here).
2. Between two such runs, evaluate the transformer's RoPE tables eagerly: `mx.eval(*model.transformer.rope_embedder.freqs_cis)`. Nothing else. A direct `transformer(...)` call, a callback, or any tool that runs the model outside `mx.compile` does the same thing.
3. Generate again with the same seed. The image is now different, and stays that way for the rest of the process. On the decoded 512x512 image up to 83 of 255 per channel change, and about three quarters of the pixels move. The latent itself barely shifts (max abs 6.25e-2, cosine 0.999977), so the VAE is what turns a tiny difference into a visible one.

Fix: `mx.eval(*self.freqs_cis)` at the end of `RopeEmbedder.__init__`. Three small float32 arrays, so it costs nothing, and the same run is then bit identical however the process got there.

Our reading of why: `RopeEmbedder` is a plain class, not an `nn.Module`, so its cos/sin tables are not in the parameter tree that gets evaluated after loading. They sit as a pending graph until something touches them. When the first touch is inside `mx.compile(predict)`, MLX 0.32 traces that pending graph into the compiled function. An eager evaluation later materialises the tables with the eager kernels, and from then on the compiled function captures those values instead. The two are evidently not the same bits.

Repro (about three minutes on an M1 Max):

```python
import mlx.core as mx, numpy as np
from mflux.models.common.config.model_config import ModelConfig
from mflux.models.z_image.variants.z_image import ZImage

model = ZImage(quantize=8, model_config=ModelConfig.z_image())
kw = dict(prompt="a portrait of a young woman, soft window light", seed=42,
          num_inference_steps=8, width=512, height=512, guidance=4.0)
run = lambda: np.asarray(model.generate_image(**kw).image, dtype=np.int16)
a, b = run(), run()
mx.eval(*model.transformer.rope_embedder.freqs_cis)
c = run()
print("run 2 vs 1:", np.abs(b - a).max(), " run 3 vs 1:", np.abs(c - a).max())
```

Prints:

```
run 2 vs 1: 0  run 3 vs 1: 83
```

Run 2 matches run 1 exactly. Run 3, after the one eager `mx.eval`, differs by 83.

Seen on an M1 Max 32 GB, mflux 0.19.1, mlx 0.32.2. The same steps on mlx 0.31.2 (with mflux 0.18.0 or 0.19.1) do not drift, so this is MLX 0.32's compile meeting the lazy tables. mflux 0.19 pins mlx 0.32, so every 0.19 user on a compiling chip has it.

IonDen:
## What

Z-Image's `RopeEmbedder` builds its rotary-embedding tables in `__init__` but never evaluates them, so they stay a pending graph. The class is a plain object rather than an `nn.Module`, so its tables are not in the parameter tree that gets evaluated when the weights load. Z-Image compiles `predict` and rebuilds it on every `generate_image` call. The first compiled run traces the pending tables; once anything materialises them eagerly, the next run captures different values. The same seed then gives a different image depending on what ran earlier in the process.

The fix evaluates the three small float32 tables once when the embedder is built, before any compile sees them.

Fixes #714.

## Release note

```release-note
Fix Z-Image producing a different image for the same seed when other code (a preview callback, another library) had run the model earlier in the same process.
```

## Verification

M1 Max 32 GB, mflux 0.19.1, mlx 0.32.2. Before the fix, evaluating the tables between two same-seed runs changed the decoded 512x512 image by up to 83 of 255 per channel across about three quarters of the pixels. After, the two runs are byte-identical.

New test `tests/test_z_image_rope_embedder.py` (no weights, runs in the default selector): compile the embedder, touch its tables, compile again, and assert the two outputs match. Red without the fix on mlx 0.32, green with it. Local fast suite: 1589 passed, 0 failed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved consistency of generated images across repeated calls in the same process.
  - Prevented positional embedding calculations from producing different results depending on prior evaluations.
  - Ensured precomputed positional data is ready before image generation begins.

- **Tests**
  - Added coverage verifying identical outputs before and after eager evaluation and compilation.
  - Organized regression coverage for positional embedding consistency.

<h3>Greptile Summary</h3>

The PR eagerly evaluates Z-Image’s precomputed RoPE tables during embedder construction so subsequent compiled prediction functions capture stable values.
- Adds a one-time `mx.eval` for the three float32 RoPE tables.
- Adds a focused regression test matching production geometry, input dtype, and recompilation behavior.

The PR appears safe to merge, with the eager evaluation correctly placed before any production compilation path.

The change evaluates three small RoPE tables once during eager model initialization, and the regression test accurately reflects the production table geometry, input contract, and fresh-compile lifecycle.

| Filename | Overview |
|----------|----------|
| src/mflux/models/z_image/model/z_image_transformer/rope_embedder.py | Materializes small, non-parameter RoPE tables before compiled prediction paths can observe different lazy/eager states; no defect identified. |
| tests/test_z_image_rope_embedder.py | Covers the reported recompilation instability using production-equivalent geometry and position-ID shape and dtype; no defect identified. |

</details>

<sub>Reviews (1): Last reviewed commit: ["fix: evaluate Z-Image&#39;s RoPE tables at c..."](https://github.com/mflux-community/mflux/commit/e39799d52636db862e533932461dc1e6c67c28de) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=60970527)</sub>

### DIFF
--- a/src/mflux/models/z_image/model/z_image_transformer/rope_embedder.py
+++ b/src/mflux/models/z_image/model/z_image_transformer/rope_embedder.py
@@ -15,6 +15,11 @@ class RopeEmbedder:
 
         self.axes_dims = axes_dims
         self.freqs_cis = RopeEmbedder._precompute_freqs_cis(axes_dims, axes_lens, theta)
+        # Materialise the tables now: they are plain arrays outside the module's parameter tree,
+        # so nothing else evaluates them. Left lazy, they are captured as a pending graph by the
+        # first mx.compile of predict and diverge from a later eager touch, changing the output
+        # across generate_image calls in the same process (issue #714). Three small float32 arrays.
+        mx.eval(*self.freqs_cis)
 
     def __call__(self, ids: mx.array) -> mx.array:
         result = []
--- a/tests/test_z_image_rope_embedder.py
+++ b/tests/test_z_image_rope_embedder.py
@@ -3,38 +3,34 @@ import mlx.core as mx
 from mflux.models.z_image.model.z_image_transformer.rope_embedder import RopeEmbedder
 
 
-def _rope() -> RopeEmbedder:
-    # The geometry the Z-Image transformer builds (axes_dims / axes_lens from
-    # z_image_transformer.transformer). Constructed directly: no weights, no model load.
-    return RopeEmbedder(theta=256.0, axes_dims=[32, 48, 48], axes_lens=[1024, 512, 512])
+class TestRopeEmbedderEagerEval:
+    @staticmethod
+    def _rope() -> RopeEmbedder:
+        # The geometry z_image_transformer builds; constructed directly, no weights, no model load.
+        return RopeEmbedder(theta=256.0, axes_dims=[32, 48, 48], axes_lens=[1024, 512, 512])
 
+    @staticmethod
+    def _pos_ids() -> mx.array:
+        n = mx.arange(8)
+        return mx.stack([n, n % 512, n % 512], axis=1).astype(mx.int32)
 
-def _pos_ids() -> mx.array:
-    # (N, 3): one token index and two spatial indices, each below its axis length.
-    n = mx.arange(8)
-    return mx.stack([n, n % 512, n % 512], axis=1).astype(mx.int32)
+    def test_rope_output_survives_a_recompile_after_the_tables_are_touched(self):
+        # Regression for issue #714: a fresh mx.compile after the RoPE tables are eagerly evaluated
+        # must give the same output as the first compile. Left lazy, the tables are captured as a
+        # pending graph and the next compile picks up different values, so a seed's image depended
+        # on process history. Reproduces on the pinned mlx 0.32 (mlx 0.31 did not drift); delete the
+        # mx.eval in RopeEmbedder.__init__ to see it go red.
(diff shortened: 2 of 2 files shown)


fxd0h (approved):
Ran it here on an M5 Max with the dev pin (mlx 0.32.0). The new test fails 3 of 3 runs without the `mx.eval` line and passes with it. The drift is one float32 ulp (2.4e-7) on 128 of the 1024 table values, compiled cos/sin against eager, which is all the VAE needs to turn into the 83/255 you measured in #714. Krea 2 computes its RoPE per call, so it doesn't have this; the Z-Image ControlNet variant shares this transformer, so it's covered. Fast suite: 1595 passed.

PR #722, opened 2026-09-16
Composite alpha-bearing images before RGB conversion

Siilent-Sea:
## What

`ImageUtil.load_image` converts every source to RGB with `convert("RGB")`. On an
image that carries alpha, that call **discards the alpha channel** and keeps the RGB
stored underneath fully transparent pixels. Two files that look identical on screen
therefore load as different arrays, and the model is conditioned on hidden RGB that
no viewer ever sees. Compositing alpha-bearing sources (RGBA/LA, or a palette with a
transparency chunk) over opaque white before the RGB conversion makes the load
depend only on the visible picture. An opaque RGB input is unaffected: alpha is 255
everywhere, so the composite is the identity.

This matters for cut-out and matted inputs (PNG cut-outs, layered edits, RGBA
references), where the under-alpha RGB is commonly garbage — often black — and can
bias generation toward a dark or hard-edged surround.

## Repro

CPU-only, no weights, no GPU, no generation. Save and run:

```python
import PIL.Image
from mflux.utils.image_util import ImageUtil

def cutout(hidden_rgb):
    img = PIL.Image.new("RGBA", (8, 8), (*hidden_rgb, 0))
    img.putpixel((4, 4), (200, 60, 30, 255))  # the only visible pixel
    return img

a = ImageUtil.load_image(cutout((0, 0, 0)))       # hidden black
b = ImageUtil.load_image(cutout((240, 30, 10)))   # hidden dark red

print(a.getpixel((0, 0)), b.getpixel((0, 0)))  # (0,0,0) (240,30,10)  <-- differs
print(a.getpixel((4, 4)), b.getpixel((4, 4)))  # (200,60,30) (200,60,30)  visible
print("identical visible inputs load identically:", a.tobytes() == b.tobytes())
```

Current main prints `False`; with this patch it prints `True` and the transparent
field becomes white.

## How

- One guard in `load_image`; opaque RGB keeps the existing `convert("RGB")` path, so
  EXIF orientation behavior and the returned mode are unchanged.
- A regression test covers RGBA and LA modes, palette transparency, the opaque-RGB
  identity path, translucent-edge blending, and the path-input signature.

## Verification

- `just install` then `just test-fast` (or `uv run pytest tests/test_image_util_alpha.py -m fast`).
- The reproducer above before/after the patch.

## Release note

```release-note
Fix `ImageUtil.load_image` discarding the alpha channel, which let hidden RGB under transparent pixels reach the model and made visually identical RGBA cut-outs load as different images.
```

## Disclosure

The project owner authorized this submission and knew that a pull request was
being opened, but did not author or technically review the change. The issue
investigation, implementation, reproducer, tests, and pull-request preparation
and submission were completed by an owner-authorized AI engineering agent. The
owner's role was to support improvement of the open-source ecosystem, not to
claim technical authorship.

## Summary by CodeRabbit

* **Bug Fixes**
  * Images with transparency now load against a white background, preventing hidden color data in transparent pixels from affecting results.
  * Semi-transparent images are blended correctly while opaque images retain their original colors.

* **Tests**
  * Added coverage for RGBA, grayscale-alpha, palette transparency, translucent pixels, and file-based image loading.

<h3>Greptile Summary</h3>

Updates `ImageUtil.load_image` to composite alpha-bearing images over opaque white before RGB conversion, preventing hidden colors beneath transparent pixels from reaching model inputs.

- Handles RGBA, LA, and palette transparency.
- Preserves the existing conversion path for opaque images.
- Adds regression coverage for hidden colors, partial transparency, palette images, and path inputs.

The PR appears safe to merge with no actionable correctness, security, or repository-rule issues identified.

The compositing guard covers alpha representations produced by realistic supported image formats, preserves opaque-image behavior, and is exercised by compatible regression tests.

<details><summary>Important Files Changed</summary>

| Filename | Overview |
|----------|----------|
| src/mflux/utils/image_util.py | Composites alpha-bearing inputs over white before returning an RGB image; no actionable defect identified. |
| tests/test_image_util_alpha.py | Adds focused regression coverage for supported alpha modes, blending behavior, opaque inputs, and file paths. |

</details>

<sub>Reviews (1): Last reviewed commit: ["Composite alpha-bearing inputs before RG..."](https://github.com/mflux-community/mflux/commit/70f7ff1c1ada2047eb9968a79c8efb257673230c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=64886997)</sub>


IonDen (commented):
Nearly ready, but there's one blocking problem with masks that have a transparent background (see the inline comment). I tested on an M1 Max 32 GB with macOS 27.0, mlx 0.32.0, mflux 0.19.1 and Pillow 12.3.0. The six new tests pass, and five of them fail on the old load_image, so they really catch the bug. The fast test suite gives 1433 passed, 0 failed, 6 skipped (main: 1427, 0, 6), and ruff is clean. CodeRabbit is right about PA and RGBa images; I reproduced it and put a shorter fix in that thread. All 5 CI checks pass on 70f7ff1.

src/mflux/utils/image_util.py
```diff
@@ -161,7 +161,16 @@ def load_image(image_or_path: PIL.Image.Image | StrOrBytesPath) -> PIL.Image.Ima
...
-        return open_oriented(image_or_path).convert("RGB")
+        image = open_oriented(image_or_path)
+        # convert("RGB") drops the alpha channel and exposes whatever RGB sits under fully
+        # transparent pixels, so two files that look identical on screen load as different
+        # arrays. Composite alpha-bearing sources over opaque white first. An opaque RGB source
+        # is unaffected: alpha==255 makes the composite the identity.
+        if image.mode in ("RGBA", "LA") or "transparency" in image.info:
+            rgba = image.convert("RGBA")
+            white = PIL.Image.new("RGBA", rgba.size, (255, 255, 255, 255))
+            return PIL.Image.alpha_composite(white, rgba).convert("RGB")
```
IonDen, line 172: issue (blocking): masks with a transparent background now repaint the whole image

Fill and CatVTON masks also load through load_image ([mask_util.py#L33](https://github.com/mflux-community/mflux/blob/70f7ff1c1ada2047eb9968a79c8efb257673230c/src/mflux/models/flux/variants/fill/mask_util.py#L33), [in_context_mask_util.py#L51](https://github.com/mflux-community/mflux/blob/70f7ff1c1ada2047eb9968a79c8efb257673230c/src/mflux/models/flux/variants/in_context/utils/in_context_mask_util.py#L51)), and in a mask, white means "regenerate this". Take a mask with white strokes on a transparent layer and black under the transparent pixels. Before this PR its background loaded as black, so only the strokes were repainted; now it loads as white, and Fill repaints everything (checked with an 8x8 PNG: pixel (0,0) is (0, 0, 0) on 532b828 and (255, 255, 255) here). Loading masks with `open_oriented(path).convert("RGB")` at those two call sites keeps the old behaviour.

src/mflux/utils/image_util.py
```diff
@@ -161,7 +161,16 @@ def load_image(image_or_path: PIL.Image.Image | StrOrBytesPath) -> PIL.Image.Ima
         # Apply the EXIF Orientation tag before the model sees the pixels: most photos straight
         # off a phone carry a non-1 orientation, and without this the model is conditioned on a
         # sideways image, not merely shown one.
-        return open_oriented(image_or_path).convert("RGB")
+        image = open_oriented(image_or_path)
+        # convert("RGB") drops the alpha channel and exposes whatever RGB sits under fully
+        # transparent pixels, so two files that look identical on screen load as different
+        # arrays. Composite alpha-bearing sources over opaque white first. An opaque RGB source
+        # is unaffected: alpha==255 makes the composite the identity.
+        if image.mode in ("RGBA", "LA") or "transparency" in image.info:
```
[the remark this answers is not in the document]
IonDen: Confirmed on Pillow 12.3.0: an in-memory RGBa or PA image skips the composite and loses its alpha. Pillow's image.has_transparency_data is True for all of these (RGBA, LA, PA, RGBa, La, and palette images with a transparency entry), so the whole check can be `if image.has_transparency_data:`, and the pillow>=12.3.0 floor already has it. La still fails at convert("RGBA"), but it fails the same way at convert("RGB") on main, so that part isn't new.

fxd0h (commented):
Confirmed IonDen's mask point with a synthetic mask (RGBA, transparent background, one white stroke) through `ImageUtil.to_array(load_image(m), is_mask=True)`: main gives a coverage of 0.0625, this branch gives 1.0, so a Fill or CatVTON mask drawn on a transparent layer would repaint the whole image. The fix itself is right for image inputs; it needs to stay out of the mask path. A `composite_alpha` argument on `load_image` defaulting to True, with the two mask loaders passing False, keeps the behaviour change where it was meant, and a test for the mask case would pin it.

src/mflux/utils/image_util.py
anthonywu, line 172: **[P2] Preserve mask-loading behavior**

Scenario — Transparent masks repaint unintended regions

1. A user supplies an RGBA mask with a white selection on transparent black.
2. Both `MaskUtil.create_masked_latents` and `InContextMaskUtil.create_masked_latents` load masks through this method.
3. Compositing over white turns the transparent background into a regeneration region. An 8×8 mask with one opaque white pixel previously selected 1 pixel; this change selects all 64. This was reproduced through `ImageUtil.load_image` followed by `ImageUtil.to_array(..., is_mask=True)`.

Fix: Keep mask loading separate from image compositing and add a transparent-mask regression test.

anthonywu (changes requested):

PR #725, opened 2026-09-17
fix: --low-ram no longer crashes on Lens, whose model never declared a tiling_config

### LINKED ISSUE #724: Lens: --low-ram crashes with AttributeError after the model has loaded
1. Run `mflux-generate-lens --prompt "a cat" --low-ram` on mflux 0.19.2. The same code is in 0.19.0 and 0.19.1.
2. The model loads in full, text encoder included.
3. `CallbackManager.register_callbacks` then builds a `MemorySaver`, and its `__init__` reads `model.tiling_config` (`src/mflux/callbacks/instances/memory_saver.py`, line 26).
4. `LensImage` never sets that attribute. Every other model that registers callbacks gets it from its initializer, and Lens has no initializer. The run stops with `AttributeError: ... object has no attribute 'tiling_config'` before the first step. The call sits outside the `try` block, so there is no memory report either.

Adding `--vae-tiling` avoids the crash, because `_apply_vae_tiling` runs first and creates the attribute.

Reproduced without weights on an M1 Max 32 GB, macOS 27.0, mlx 0.32.0, at f9eec4e. I ran the real `lens_generate.build_parser()` and the real `CallbackManager.register_callbacks` against a stand-in object that carries exactly the attributes `LensImage.__init__` sets. The path became reachable when #579 wired callbacks into the Lens CLI.

Fix: read the attribute with `getattr(model, "tiling_config", None)` in `MemorySaver.__init__`, the way `_apply_vae_tiling` already does on the same model one call earlier.

IonDen:
## What

`mflux-generate-lens --low-ram` loaded the whole model and then stopped with `AttributeError: 'LensImage' object has no attribute 'tiling_config'`. `MemorySaver.__init__` read `model.tiling_config` directly, and Lens is the one model that registers callbacks without setting it: every other family gets it from its initializer, and Lens has none. `MemorySaver` now reads it with `getattr(model, "tiling_config", None)`, the way `_apply_vae_tiling` already does one call earlier, so low-RAM mode installs the default `TilingConfig` on Lens as it does everywhere else.

Nothing changes for the other models. They all declare the attribute, so the read returns what it returned before.

One thing this does not fix: Lens's decode does not read `tiling_config` yet, so tiling itself still is not applied on Lens. That is a separate change.

Fixes #724

- [x] Tests added/updated run in CI by default (selector: `-m "not slow and not high_memory_requirement"`, same as `just test`). Only mark `@pytest.mark.slow` or `@pytest.mark.high_memory_requirement` when a test exceeds CI's time or memory budget (typically weight downloads / image generation).
- [x] `ruff check` and `ruff format` are clean (`uv run ruff` uses the version pinned in the dev dependencies of `pyproject.toml`, which is the single source of truth for pre-commit and CI; `pre-commit run -a` covers it locally).
- [x] Release note block below filled in (or `none` for changes users never see).
- [ ] Docs updated where behavior changed — README examples/table rows are part of the API contract (see `.cursor/rules/RULE.md`). Not needed: no documented behavior changes.

## Release note

```release-note
`mflux-generate-lens --low-ram` no longer crashes with an AttributeError after the model has loaded.
```

## Verification

- The new test, `test_low_ram_installs_default_tiling_on_a_model_that_never_declared_one`, fails on main with the AttributeError at `memory_saver.py:26` and passes with the fix.
- `uv run pytest tests/callbacks -q`: 25 passed.
- `just lint` is clean. `just test-fast`: 1435 passed, 6 skipped.
- M1 Max 32 GB, macOS 27.0, mlx 0.32.0. No weights were loaded. The crash happens during callback registration, which the test drives with a stand-in model.


fxd0h (approved):
Reproduced on main with the Comfy-Org/Lens weights: `mflux-generate-lens --low-ram -q 8` at 512 px loads and then stops with `AttributeError: 'LensImage' object has no attribute 'tiling_config'`. On this branch the same command renders (4 steps, seed 1). Fast suite 1608 passed here.

PR #720, opened 2026-09-11
Update readme - add contributor section

ianscrivener:
## What
Update README.md, adding a "Contributors" section with contribtor buttn images & acknowledgement to Filip

## Checklist (definition of done)

- [ ] Tests added/updated run in CI by default (selector: `-m "not slow and not high_memory_requirement"`, same as `just test`). Only mark `@pytest.mark.slow` or `@pytest.mark.high_memory_requirement` when a test exceeds CI's time or memory budget (typically weight downloads / image generation).
- [ ] `ruff check` and `ruff format` are clean (`uv run ruff` uses the version pinned in the dev dependencies of `pyproject.toml`, which is the single source of truth for pre-commit and CI; `pre-commit run -a` covers it locally).
- [ ] Release note block below filled in (or `none` for changes users never see).
- [ x ] Docs updated where behavior changed — README examples/table rows are part of the API contract (see `.cursor/rules/RULE.md`).
- [ ] New model: shared config wiring (aliases, default steps, mflux-save dispatch, capabilities, completions), thin CLI entrypoint, and `src/mflux/models/<name>/README.md`.
- [ ] New/changed CLI: ignored/rejected options declared (`IGNORED_OPTIONS`/`REJECTED_OPTIONS`) and `warn_ignored_options` actually called in `main()` — `mflux-capabilities` must stay truthful.

## Release note

Update README.md
```release-note
Update README.md
```

## Summary by CodeRabbit

- **Documentation**
  - Added a Contributors section to the README.
  - Added contributor acknowledgments and project origin credit.

<h3>Greptile Summary</h3>

Adds a Contributors section to the main README.
- Links the new section from the table of contents.
- Displays the repository’s contrib.rocks contributor image.
- Credits Filip Strand as the original creator.

<h3>Confidence Score: 5/5</h3>

The documentation-only change appears safe to merge.

The new table-of-contents link resolves correctly, the attribution matches repository metadata, and no behavioral, security, or repository-rule violations were identified.

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| README.md | Adds contributor recognition and original-creator attribution without introducing an actionable issue. |

</details>

<sub>Reviews (1): Last reviewed commit: ["ypdate readme - add contributor section"](https://github.com/mflux-community/mflux/commit/bf16af99768e35d4995170e2157d49a73e67bc75) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62924437)</sub>

PR #731, opened 2026-09-18
test: measure the checkpointing peak under a memory limit so GPU timing cannot move it

### LINKED ISSUE #730: Flaky test: test_ordered_backward_lowers_peak_memory_below_plain_checkpointing reads a peak that depends on GPU timing
The peak-memory test in `tests/training/test_gradient_checkpointing.py` fails on CI now and then, on changes that never touch training code.

1. Open any PR. CI runs `python -m pytest -m "not slow and not high_memory_requirement" -v` on the `macos-14` runner.
2. Most runs pass. Sometimes `test_ordered_backward_lowers_peak_memory_below_plain_checkpointing` fails on `assert ordered_peak < 0.8 * ckpt_peak`, with the ratio a few points over 0.8. Seen on `main` after the README-only #720 (run 35121775168: `156176768 < 0.8 * 189878056`, ratio 0.82) and on #729, a callbacks change (run 35395154750: `146845860 < 0.8 * 173086248`, ratio 0.85).
3. On an M1 Max 32 GB (mlx 0.32.0, macOS 27.0, `main` at f9eec4e) the same test passes every time, with a ratio between 0.50 and 0.71 depending on what ran before it in the same process.
4. The cause is the measurement, not the ordered backward. `mx.get_peak_memory()` is the high-water mark of the buffers MLX counts as active, and a buffer is released by MLX's completion handler after `mx.eval` has returned, not when the array is dropped. So the reading folds in whatever earlier work is still being released, and how much of a step's own memory is alive at once depends on GPU timing. Both only ever push the number up, and the second condition, always the ordered one, pays for it. A small demonstration:

```python
import mlx.core as mx
junk = [mx.zeros((256 * 2**20 // 4,)) for _ in range(4)]
mx.eval(junk)
del junk
print(mx.get_active_memory() // 2**20)  # 1024: the dropped gigabyte is still counted
mx.synchronize()
print(mx.get_active_memory() // 2**20)  # 0
```

Fix: wait for earlier releases with `mx.synchronize()`, then measure each condition under a memory limit that is always exceeded (`mx.set_memory_limit(1)` around the step, restored afterwards). MLX then waits for queued work before continuing, so the peak is the memory the step keeps alive at once on top of what the process already holds: 96 MB plain, 67 MB ordered, ratio 0.69, the same on every run here in either order. The 0.8 bar stays as it is. The fix is #731.

IonDen:
## What

`test_ordered_backward_lowers_peak_memory_below_plain_checkpointing` fails on CI now and then on unrelated changes, most recently on #729 and on `main` after the README-only #720. The ordered backward is fine; the measurement is not. MLX releases a buffer from its completion handler after `mx.eval` has returned, so `mx.get_peak_memory()` folds in whatever earlier work is still being released, and how much of a step's own memory is alive at once depends on GPU timing. Both effects only ever push the reading up, and the second condition, which is always the ordered one, pays for them. On a slow shared runner that is enough to cross 0.8. Fixes #730.

This PR changes only the test. Before each measured step it waits for earlier releases (`mx.synchronize()`), then runs the step under a memory limit that is always exceeded (`mx.set_memory_limit(1)`, restored in a `finally`). MLX then waits for queued work before continuing, so the peak is the memory the step keeps alive at once on top of what the process already holds, and it comes out the same on every run: 96 MB plain, 67 MB ordered, ratio 0.69 here, in either order. The 0.8 bar stays. With the limit the reading no longer depends on timing, so a future failure means either the ordered backward regressed or an earlier test left tens of MB of live arrays in the process. A small new test pins the wait: it drops 256 MB of arrays and checks that active memory is back near its starting point afterwards, which fails without the synchronize.

- [x] Tests added/updated run in CI by default: the changed test and the new `test_wait_for_releases_drops_buffers_still_being_released`, both `fast`. The file runs in about the same time as before (under 2 s).
- [x] Release note block below filled in.
- [x] Docs: no user-facing behaviour changes.

## Release note

```release-note
none
```

## Verification

- `uv run pytest tests/training/test_gradient_checkpointing.py`: 4 passed, three runs in a row; the full fast selector passes.
- On an M1 Max 32 GB (mlx 0.32.0, macOS 27.0): with the limit in place the two peaks read 96 MB and 67 MB on every one of eight trials in each order, ratio 0.693. Without it the same machine gave ratios from 0.50 to 0.71 depending on what ran earlier in the process, and CI gave 0.82 and 0.85 on the two failing runs.
- Mutation checks: disabling the ordered backward makes the ratio test fail at a ratio of 1.0, so it still guards what it was written for; removing the `mx.synchronize()` makes the new test fail three times out of three, with the dropped 256 MB still counted as active.

## Summary by CodeRabbit

- **Tests**
  - Added coverage to verify that released buffers are cleared before memory measurements.
  - Improved peak-memory measurement reliability for gradient checkpointing comparisons by ensuring pending releases are completed first.
  - Updated ordered-backward memory tests to compare individual peak-memory measurements rather than selecting the lowest result.


fxd0h (commented):
The flake does not reproduce on an M5 Max: the memory test passed 10 of 10 runs on main and 10 of 10 on this branch, with the branch costing nothing measurable (13 s against 12 s for the ten runs). So the machine that shows it is CI's, and CI is the judge of the fix; what I can say from here is that measuring under a limit that forces MLX to wait for queued work is the right kind of fix (it removes the timing from the reading instead of widening the margin), the previous limit is restored in `finally`, and the new release test's 256 MB is fine on the runners.

fxd0h (approved):
Approving on the reasoning above; CI's runs will tell.

PR #735, opened 2026-09-19
fix: --vae-tiling and --low-ram tile the decode on FLUX.2 Klein, Lens, Krea 2, Ideogram 4 and Boogu

### LINKED ISSUE #734: --vae-tiling and --low-ram do not tile the decode on FLUX.2 Klein, Lens, Krea 2, Ideogram 4 and Boogu
`--vae-tiling` is accepted by every image command, `--low-ram` turns it on, and its help says it decodes "in overlapping tiles to reduce peak memory during the VAE decode phase". On six decode paths the flag is never read, so the decode still runs in one piece.

1. Run FLUX.2 Klein at 1536x1536 with `--low-ram` or `--vae-tiling` on a 32 GB Mac.
2. The loop finishes, then the VAE decode would need up to about 21 GB on top of the weights, more than the machine has left once the model is loaded. With the 512 px tiles the flag promises, the same decode needs about 3 GB.
3. These calls never pass the model's `tiling_config`: `flux2_klein.py:129`, `flux2_klein_edit.py:185` and `lens_image.py:126` call `decode_packed_latents(...)` without it, and `krea2.py:185`, `ideogram4.py:194` and `boogu_image.py:104` call `self.vae.decode(latents)` directly.
4. Everything they need exists already. `Flux2VAE.decode_packed_latents` takes a `tiling_config` argument, and FLUX.1, Z-Image and Qwen-Image decode through `VAEUtil.decode(..., tiling_config=self.tiling_config)`. #407 asked for this on Klein and was closed with #475, which restored the flag but did not reach FLUX.2's decode calls.
5. Extra memory for one decode, measured on an M1 Max 32 GB with mlx 0.32.0. The table uses random bf16 weights, since decode memory depends on shapes, not values; the real-weight runs at 1024 px came in at or below it (Klein 6.8 GB, Krea 2 7.0 GB untiled).

| VAE (models) | 1024 px | 1536 px | 2048 px | 512 px tiles, any size |
|---|---|---|---|---|
| FLUX.2 (Klein, Klein edit, Lens, Ideogram 4) | 8.6 GB | 21.5 GB | over 28 GB | about 3 GB |
| Qwen (Krea 2) | 7.0 GB | 15.2 GB | 27.1 GB | 3.4 GB |
| FLUX.1 (Boogu) | 7.6 GB | 17.0 GB | 26.3 GB | about 2.7 GB |

A whole Krea 2 q4 run at 1024 px peaked at 22 GB, its weights plus this decode. To reproduce the FLUX.2 row without downloading anything (it prints the process peak: 8.8 GB untiled and 4.0 GB tiled here):

```python
import mlx.core as mx
from mlx.utils import tree_map
from mflux.models.flux2.model.flux2_vae.vae import Flux2VAE
from mflux.models.common.vae.tiling_config import TilingConfig

vae = Flux2VAE()
vae.update(tree_map(lambda p: p.astype(mx.bfloat16), vae.parameters()))
packed = mx.random.normal((1, 128, 64, 64))  # a 1024x1024 image
for tiling in (None, TilingConfig(vae_decode_tile_size=512)):
    mx.eval(vae.parameters(), packed); mx.clear_cache(); mx.reset_peak_memory()
    mx.eval(vae.decode_packed_latents(packed, tiling_config=tiling))
    print("tiled" if tiling else "untiled", round(mx.get_peak_memory() / 2**30, 1), "GB")
```

ERNIE-Image uses the same FLUX.2 VAE but sets `TilingConfig(vae_decode_tiles_per_dim=None)` in its initializer, which turns decode tiling off under every flag. Is that intended?

Fix: pass `self.tiling_config` at the six call sites, as an argument to `decode_packed_latents` for the FLUX.2 VAE and through `VAEUtil.decode` for the others, the way FLUX.1 does. With no flag set, `VAEUtil.decode` makes the same single `vae.decode` call as today, so default output does not change. PR follows.

IonDen:
## What

`--vae-tiling` and `--vae-tile-size` now tile the final VAE decode on FLUX.2 Klein, Klein edit, Lens, Krea 2, Ideogram 4 and Boogu, where the flag was accepted and ignored. `--low-ram` tiles Krea 2 and Boogu; on the FLUX.2 VAE it leaves the decode untiled, because that decoder takes its normalisation statistics per tile and shifts colour between them. The three FLUX.2-VAE models pass `tiling_config=self.tiling_config` to `decode_packed_latents`, which already took it; Krea 2, Ideogram 4 and Boogu decode through `VAEUtil.decode`, as FLUX.1 does. With no flag set the decode is the same single call as before. Lens never set a `tiling_config` attribute, so it now starts at `None` like the other models. Fixes #734.

What tiling costs, measured once per model on a real 1024 px image, decoding the same final latent both ways (M1 Max 32 GB, mlx 0.32.0):

| | Klein 4B (q8) | Krea 2 (q4) |
|---|---|---|
| decode memory, untiled → 512 px tiles | 6.8 → 2.4 GB | 7.0 → 3.4 GB |
| decode time | 1.61 → 1.68 s | 2.47 → 3.17 s |
| tiled vs untiled | PSNR 27 dB, mean difference 7.7 levels | PSNR 44 dB, 97% of pixels within 2 levels |

On Klein each tile comes out with a slightly different colour balance, because the FLUX.2 VAE's GroupNorm layers take their statistics from the area they are given. The shift is a few levels on the large tiles and up to about 19 on the narrow tiles along the bottom edge, with no visible seam at normal size. The FLUX.1 VAE has the same GroupNorm layers and already decodes tiled under `--low-ram`. ERNIE-Image, on the same FLUX.2 VAE, turns decode tiling off in its initializer; I left it as it is and asked about it in the issue. Stepwise previews still decode untiled; that is a separate change.

- [x] Tests added/updated run in CI by default: 15 fast tests in `tests/vae/test_decode_honours_tiling_config.py`. Each model's real `generate_image` runs with fake, no-weights components and a recording VAE: with a tiling config the FLUX.2-VAE models must hand that exact object to `decode_packed_latents` and the others must decode in several tiles; with none, the single untiled call as before. Two more drive the real `Flux2VAE.decode_packed_latents` and check that it tiles. The Ideogram 4 fake model in `test_generate_image_ideogram4.py` now sets `tiling_config`, as the real initializer does.
- [x] Release note block below filled in.
- [x] Docs: the flag's help text already describes this behaviour.

## Release note

```release-note
`--vae-tiling` and `--low-ram` now tile the final VAE decode on Krea 2 and Boogu, where the flag was accepted and ignored, cutting the decode at 1024 px from about 7 GB to about 3 GB. FLUX.2 Klein, Klein edit, Lens and Ideogram 4 tile only when you ask with `--vae-tiling`: their shared decoder takes its normalisation statistics per tile, which shifts colour between tiles, so low-RAM leaves them untiled.
```

## Verification

- `uv run pytest -m "not slow and not high_memory_requirement"`: 1635 passed, 6 skipped. `ruff check`, `ruff format --check` and `ty check`: clean.
- The seven tests that pin the fix fail on `main` and pass here; the two `Flux2VAE` tests pass on both, since they pin the VAE side this PR relies on.
- The real-weights runs in the table: 1024x1024, seed 42, Klein 4B quantized to 8-bit at load with 4 steps, Krea 2 Turbo from `mflux-community/krea-2-turbo-mflux-q4` with 8 steps. Re-decoding untiled reproduces each model's own output (identical on Krea 2, within 1 level on Klein).

## Summary by CodeRabbit

* **New Features**
  * VAE tiling is now consistently supported across additional image-generation models, improving large-image decoding.
  * Low-RAM mode automatically enables tiling only for compatible decoders and preserves explicit tiling settings.

* **Bug Fixes**
  * Standard and packed latent decoding now consistently honors tiling configuration.
  * Prevented implicit tiling for decoders where it may cause visual inconsistencies.

* **Tests**
  * Added coverage for tiled decoding, non-tiled decoding, and low-RAM behavior.

<h2><a href="https://app.greptile.com/api/retrigger?id=66480798"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; the configuration handoffs match the existing VAE interfaces and callback lifecycle.

<h3>Summary</h3>

Makes final VAE decoding honor tiling configuration across FLUX.2 Klein, Klein edit, Lens, Krea 2, Ideogram 4, and Boogu.

- Forwards configuration through the existing packed-latent decoder or shared VAE utility, without changing latent normalization or PiD decoding.
- Initializes Lens with tiling disabled by default and adds 15 regression tests for tiled and untiled paths.
- No actionable correctness or repository-rule violations were identified. Review was static; tests were not run locally.

<details open><summary>Diagram</summary>

```mermaid
flowchart TD
    A["--vae-tiling or --low-ram"] --> B["Model tiling_config"]
    B --> C["Klein / Klein edit / Lens"]
    B --> D["Krea 2 / Ideogram 4 / Boogu"]
    C --> E["decode_packed_latents: unpack and normalize"]
    E --> F["VAEUtil.decode"]
    D --> F
    F --> G{"Decode tiling enabled?"}
    G -->|Yes| H["VAETiler: decode and blend tiles"]
    G -->|No| I["Single VAE decode"]
    H --> J["Image conversion"]
    I --> J
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["fix: --vae-tiling and --low-ram tile the..."](https://github.com/mflux-community/mflux/commit/91c0740d789e6640b283b08b3fe4ca8819955066)</sub>


fxd0h (commented):
Ran it on an M5 Max, 1024 px, seed 42, q8. Fast suite: 1641 passed.

Krea 2 (8 steps): on main `--vae-tiling` gives the plain image to the pixel, so the flag was ignored as the issue says; on this branch it tiles, and the result sits at PSNR 53 dB from the untiled decode, mean 0.11 levels, 99.5% of pixels within 2. Nothing to see. `--low-ram` gives the same tiled image.

Klein 4B (4 steps): PSNR 26.1 dB, mean 9.3 levels, and the difference is not noise but per-tile colour offsets: the four 512 px quadrants of the tiled image sit at mean signed shifts of about (-2, -2, +5), (+6, +1, 0), (-18, -8, -9) and (+8, 0, -13) from the untiled one, and the column means step by 8 levels at x 928 and by 9 at the 512 boundary. On the picture that reads as a lighter vertical band through the sky and the water. That is the GroupNorm point your description makes, and I'd rather it didn't become the default: `--low-ram` now switches tiling on for Klein, Klein edit and Lens, so a low-RAM user of those three goes from a correct image with a 7 GB decode to a seamed one with a 3 GB decode, without asking for it. Two ways out that keep the fix: let explicit `--vae-tiling` do what it does on the FLUX.2-VAE models and have `--low-ram` leave them untiled until the norm statistics are handled (the overlap is 8 latent pixels, which cannot hide a per-tile offset), or make the tiles large enough on that VAE that the offsets stay under a couple of levels; either way the release note should say what low-RAM users of those models get. The Krea 2, Ideogram 4 and Boogu side I'd merge as is.

IonDen:
Thanks for running all five. I measured the same cast and put it in the description, then still let `--low-ram` turn tiling on, so your quadrant numbers are what settle it.

Your first option is in. Low-RAM no longer picks tiling on the FLUX.2 VAE, while `--vae-tiling` and `--vae-tile-size` still tile every model, so the memory route stays open on request.

Two corrections. Ideogram 4 runs the same FLUX.2 VAE, so it joins Klein, Klein edit and Lens on the opt-out side, which leaves Krea 2 and Boogu tiling under low-RAM. Training also kept its own copy of the decision, so Klein preview images had the same problem. Both now read one rule, and it sits on the VAE.

I skipped the larger tiles. The offset is a per-tile normalisation shift rather than a seam, so a wider overlap cannot hide it.

I updated the release note. Nine fast tests cover the rule, and one of them fails if a FLUX.2 model stops carrying the opt-out.

ianscrivener (approved):
<sub>Approved in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/mflux-community/mflux/pull/735)</sub>

PR #736, opened 2026-09-20
feat: Qwen-Image-2.1 support (txt2img, img2img, quantization)

ivanfioravanti:
## Summary

MLX port of **Qwen-Image-2.1** (`Qwen/Qwen-Image-2.1`), the second-generation Qwen Image text-to-image model: a 7.1B single-stream, block-causal DiT with a Qwen3-VL text encoder, a new 64-channel causal VAE (16× spatial compression), and 40-step guidance-free sampling.

- New CLI: `mflux-generate-qwen-2.1` (txt2img + img2img, optional true CFG with `--negative-prompt`)
- Registered in `ModelConfig` (aliases `qwen-image-2.1` / `qwen-2.1` / `qwen-image-21`), `mflux-save`, and the defaults table (40 steps)
- Weight quantization works (`-q 8`, `-q 4`); the Qwen3-VL text encoder stays bf16 (same policy as qwen 1.x and krea2 — quantizing the VL tower degrades conditioning)

## Implementation notes

- Block-causal attention runs as two fast SDPA calls (causal over the text prefix + full attention for the target block), matching the reference processor's segmentation
- `causal_condition`: text tokens are modulated at `t = 0`, image tokens at the sampled `t`, so text activations are step-independent
- 3-axis RoPE with the frame axis frozen at the text position and a zero-centred height/width grid
- Latents are consumed unpatched (plain spatial flatten of the 64-channel latents)
- Timestep sinusoid is computed in fp32 and cast to `ModelConfig.precision` before the embedder; casting by weight dtype would shred the embedding under quantization (packed uint32 weights) — this was a real bug found by the q8 golden test

## Parity & performance (M5 Max, 1024², 40 steps, bf16)

- Latent parity vs the diffusers reference: mean abs diff 3.6e-4 with reference-injected noise/latents
- Channel correlation vs a full reference generation: 0.989–0.992
- ~1.5 s/step, ~78 s end-to-end (MPS diffusers reference: ~85 s); peak ~46 GB bf16, ~30.7 GB with `-q 8`
- Goldens lock mflux-native sampling (MLX RNG differs from PyTorch's, as with every model in this repo)

## Shared code touched (and why)

- `pyproject.toml`, `cli/defaults/defaults.py`, `common/config/model_config.py`, `common/cli/save.py`, `common/weights/loading/weight_definition.py` — required wiring for a new model
- `common/latent_creator/latent_creator.py` — added `Qwen21LatentCreator` to the `LatentCreatorType` union

## Intentionally not ported

[...]

PR #741, opened 2026-09-21
feat: add Qwen 2.1 reference editing, RGBA output and prefix caching

dreampuf:
## What

Extend the Qwen-Image-2.1 support introduced in #736 with instruction-based reference-image editing, RGBA output, and prefix KV caching. The new `mflux-generate-qwen-2.1-edit` command and `QwenImage21Edit` API accept up to ten reference images; for example, a portrait and a clothing reference can produce the same person wearing the requested shirt.

The existing `mflux-generate-qwen-2.1` command, strength-based img2img, model registration, `mflux-save` dispatch, and image goldens retain their current behavior. This branch is based directly on `8c00dab` and includes no unrelated local commits.

### Implementation and compatibility

- Add Qwen3-VL vision processing, DeepStack features and multimodal positions; splice reference VAE latents into block-causal attention with zero-timestep conditioning and optional prefix caching.
- Preserve RGBA through reference preprocessing, the 64-channel VAE, and output serialization. The vision encoder gets a separate white-composited RGB copy.
- Keep the tested editing components under `qwen21/reference` and the public pipeline under `qwen21/variants/edit`. This retains the existing text-only implementation and its checkpoint layout while adding a complete vision-capable pipeline. The reference pipeline loads checkpoint configurations and validates tensor names/shapes.
- Support quantization and complete local exports through `QwenImage21Edit.save_model`. Existing text-only `mflux-save` exports omit the vision tower and are not interchangeable with these exports; the README documents both paths.
- Shared changes are limited to opt-in vision dtype preservation (default off for existing users), model-specific metadata serialization/replay, and CLI/model inventory wiring. No dependency or lockfile changes.

### Validation and limits

Real checkpoint tests cover 1024 text generation, transparent PNG output, single- and two-reference edits at 512, a successful 896×1152 clothing edit, and pixel-identical 8-bit export/reload. The documentation records checkpoint/reference revisions, settings, memory use, and numerical comparisons.

The 1024×1024 panda snow/dual-subject examples did not follow the requested changes, including with expanded prompts. A controlled official Diffusers denoising run reproduced the snow-edit failure closely (PSNR 49.91 dB against dense MLX). That comparison shares verified VAE condition latents and decoding, so it is not presented as an untouched end-to-end Diffusers run. The successful clothing example does not replace those failures. A 2048 run exercised four steps and tiled decoding only; full-step 2048 quality is unverified. Training, LoRA loading, and the optional prompt-rewriting models are outside this PR.

896×1152 two-reference clothing edit

The example uses the public inputs and prompt from the [ComfyUI editing workflow](https://github.com/Comfy-Org/workflow_templates/blob/main/templates/image_qwen_image_2_1_image_edit.json), 8-bit weights, 25 steps, guidance 1, and seed 1070478148268574. The showcase has the same pixels as the tested output, with metadata removed. It is not an image golden or a pixel comparison against ComfyUI.

- [x] Tests added/updated run in CI by default. Numerical Diffusers tests skip when the optional reference package is absent; a separate pinned-reference run executes them.
- [x] Release note block filled in.
- [x] Docs updated, including CLI/API examples, export compatibility and measured validation limits.
- [x] Reuse existing shared model registration/defaults/save dispatch; add a thin editing CLI and capability/model inventory coverage. Existing argparse-driven completions discover the new options.
- [x] Conditional scheduler/negative-prompt behavior declared; runtime warns when an explicitly supplied negative prompt is ignored at guidance 1.

## Release note

```release-note
Qwen-Image-2.1 now supports instruction-based editing with up to ten reference images, RGBA output, and prefix KV caching through mflux-generate-qwen-2.1-edit. The new QwenImage21Edit Python API can quantize and export complete vision-capable checkpoints; the existing generation command and text-only exports remain available.
```

## Verification

- CI default selector (`not slow and not high_memory_requirement`): **1667 passed, 10 skipped, 63 deselected**.
- Fast selector: **1500 passed, 10 skipped, 230 deselected**.
- Focused model/reference suite with Diffusers `80c7ed262aeffbeb43ef13ae04baeb9b84515a69`: **35 passed**, using MPS references.
- `UV_OFFLINE=1 just lint`, Ruff format check, type checking, and `just build`.
- Installed editing CLI: saved 8-bit checkpoint + original 512-pixel/eight-step metadata reproduces the pre-integration PNG pixel for pixel. A separate 40-step, 512-pixel reference-image snow edit also matches its successful pre-integration PNG exactly.
- Capability discovery reports full coverage for the new command; model inventory exports all 36 entries.

The PR worktree installs its own editable mflux distribution and reuses the existing local dependency packages. Its fresh offline `uv sync` could not find the public-PyPI Torch wheel URL in cache, so tests used `MFLUX_PRESERVE_TEST_OUTPUT=1 uv run --no-sync python -m pytest -m fast`, equivalent to the recipe's test command after synchronization. See `src/mflux/models/qwen21/reference/VALIDATION.md` for the full evidence and environment limitations.

## Summary by CodeRabbit

- **New Features**
  - Added Qwen Image 2.1 editing with instruction-based and multiple-reference image generation.
  - Added a dedicated editing command supporting up to 10 reference images, configurable resolution, KV caching, and PNG, WebP, or TIFF outputs.
  - Generation metadata now preserves model-specific parameters.
  - Added RGBA support documentation and reference-editing guidance.

- **Bug Fixes**
  - Improved dtype preservation for Qwen3-VL vision processing.
  - Restored additional generation settings from saved metadata when not explicitly provided.

<h2><a href="https://app.greptile.com/api/retrigger?id=66855228"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR appears safe to merge after considering two non-blocking CLI edge cases involving abbreviated metadata overrides and late guidance validation.

<h2>Findings</h2>

1. P2&nbsp;**Abbreviated Override Gets Reversed** <a href="https://github.com/mflux-community/mflux/pull/741#discussion_r4060492086">▶</a>
2. P2&nbsp;**Invalid Guidance Loads Model** <a href="https://github.com/mflux-community/mflux/pull/741#discussion_r4060492098">▶</a>

<details><summary>Fix with agent prompt</summary>

`````markdown
### Issue 1
src/mflux/cli/parser/parsers.py:468
When replaying metadata, argparse accepts an unambiguous abbreviation such as `--no-use-kv`, but `_option_was_provided` recognizes only the full spelling. A sidecar containing `"use_kv_cache": true` therefore overwrites the explicitly parsed `False` value and silently re-enables caching. Detect the parsed option independently of exact raw-token matching, or disable long-option abbreviations for this parser.

### Issue 2
src/mflux/models/qwen21/cli/qwen21_edit_generate.py:85-89
Values such as `--guidance 0`, `--guidance nan`, or `--guidance inf` pass argument parsing and all preflight checks. The large model is then loaded before `generate_image` raises an uncaught `ValueError`, wasting substantial startup time and memory. Validate the finite, at-least-one guidance constraint in the existing pre-load validation block so invalid invocations fail immediately with a parser error.

---

`````

</details>

<h3>Summary</h3>

This PR adds the Qwen-Image-2.1 reference-editing command and API, including multi-reference conditioning, RGBA output, prefix KV caching, model export, metadata replay, and Qwen3-VL dtype preservation.
- Registers the new CLI and model capability inventory.
- Adds opt-in dtype preservation through the shared Qwen3-VL vision stack.
- Extends generated-image metadata with model-specific replay parameters.
- Adds fast parity, layout, cache, RGBA, and parser coverage.
- Two non-blocking CLI validation and metadata-precedence issues remain.

<details open><summary>Diagram</summary>

```mermaid
flowchart LR
  CLI[mflux-generate-qwen-2.1-edit] --> PARSE[Parse arguments and metadata]
  PARSE --> RESOLVE[Resolve Qwen 2.1 configuration]
  RESOLVE --> MODEL[Load QwenImage21Edit]
  MODEL --> ENCODE[Encode prompt and references]
  ENCODE --> CACHE[Build or reuse prefix KV cache]
  CACHE --> SAMPLE[Denoising loop]
  SAMPLE --> VAE[RGBA VAE decode]
  VAE --> OUTPUT[Image and replay metadata]
```
</details>

<sub>Reviews (1) · Last reviewed commit: ["feat: add Qwen 2.1 reference editing and..."](https://github.com/mflux-community/mflux/commit/9cc2491540635e9e0304141c5c6b9d2dd9768668)</sub>


fxd0h (commented):
Ran the branch on an M5 Max (torch 2.13, transformers 5.15, no diffusers, so the transformer parity cases skip). The fast suite is 1697 passed and 7 skipped except the six that CI also fails, and those are two separate problems.

CI's is the environment. All six die with `MPS backend out of memory (MPS allocated: 0 bytes, other allocations: 377 MiB, max allowed: 7.93 GiB). Tried to allocate 1.50 KiB`, which is the runner's virtualised Metal, not your code. The file is the first in the repo to put torch tensors on `device="mps"`; everything else that touches torch in tests stays on CPU. The configs here are tiny (hidden 32, three layers), so `cpu` costs nothing: with mps swapped for cpu the run takes 1.4 s instead of 8.3 s.

Mine is numerical, and it survives the swap to CPU. `test_text_encoder_and_deepstack` fails for the three float32 cases, here and on every rerun: 2 of 160 elements differ, worst 1.23e-4 against `atol=1e-4`, at values around 0.027, so about 0.45% relative. The bfloat16 cases pass. It isn't the GELU split, which the port gets right: `nn.GELU()` in transformers is the exact one and your mergers use `nn.gelu`, the blocks use `gelu_approx` for `gelu_pytorch_tanh`, and the two differ by at most 4.1e-4 over [-3, 3], which is what the x10 on `linear_fc1` is there to expose. What is left is float32 accumulation order between MLX and torch, amplified by that same x10. Either the float32 case wants a looser atol, or the scaling wants to be smaller.

The shared-code changes read as contained to me. `generation_parameters` is a new optional dict threaded through `ImageUtil.to_image` and `GeneratedImage`, with a guard that raises if a model tries to overwrite a standard metadata key, and no existing caller passes it. `preserve_input_dtype` on the Qwen3-VL vision blocks defaults to False, so the existing Qwen path is untouched, and the new encoder is the only one that turns it on. I have not reviewed the block-causal attention, the prefix cache or the reference-latent splicing against diffusers; that needs a diffusers revision with QwenImage21, which I do not have here.

PR #742, opened 2026-09-21
Fix ci GitHub action

ianscrivener:
## What

• add Qwen-Image-2.1 to the CI Manifest script - `scripts/ci_extract_models.py` 
• improve the Agent directions & Skills to reminde AI Agents to do this task

- [ X ] Tests added/updated run in CI by default (selector: `-m "not slow and not high_memory_requirement"`, same as `just test`). Only mark `@pytest.mark.slow` or `@pytest.mark.high_memory_requirement` when a test exceeds CI's time or memory budget (typically weight downloads / image generation).
- [ X ] `ruff check` and `ruff format` are clean (`uv run ruff` uses the version pinned in the dev dependencies of `pyproject.toml`, which is the single source of truth for pre-commit and CI; `pre-commit run -a` covers it locally).
- [ ] Release note block below filled in (or `none` for changes users never see).
- [ ] Docs updated where behavior changed — README examples/table rows are part of the API contract (see `.cursor/rules/RULE.md`).
- [ ] New model: shared config wiring (aliases, default steps, mflux-save dispatch, capabilities, completions), thin CLI entrypoint, and `src/mflux/models/<name>/README.md`.
- [ ] New/changed CLI: ignored/rejected options declared (`IGNORED_OPTIONS`/`REJECTED_OPTIONS`) and `warn_ignored_options` actually called in `main()` — `mflux-capabilities` must stay truthful.

## Release note
```release-note
Add Qwen-Image-2.1 to the CI Manifest script (`scripts/ci_extract_models.py`) .. and improve the Agent directions & Skills to remind AI Agents to do this task.
```

## Verification
None

## Summary by CodeRabbit

- **New Features**
  - Added registry metadata for the Qwen Image 2.1 model, including its model family, active status, supported command, and standard quantization options.

- **CI Improvements**
  - Model-manifest validation now runs for pull requests as well as other supported workflow events.
  - Added checks to ensure newly registered models are correctly represented in CI-generated manifests.

- **Documentation**
  - Updated contributor guidance and checklists for model registration and CI verification.


fxd0h (commented):
Checked it here: on main `scripts/ci_extract_models.py` exits with "no OVERLAY entry for: qwen-image-2.1", and on this branch it writes 36 models with the row you added; the aliases, repo id and CLI name all match the registry and pyproject. The manifest gap is real and the fix is right.

Two things before it goes in.

`.gitignore` now ignores `AGENTS.md`, and that file is tracked (it has been since #323). Git keeps serving the tracked copy, so nothing breaks today, but anyone who deletes it and writes a new one will find git silently ignoring it. The `.agents/skills/gitnexus-*` lines are fine, they match what `.claude` and `CLAUDE.md` already do, and `.agents/skills/remediating-dependabot/SKILL.md` stays tracked as it should. I'd drop just the `AGENTS.md` line.

`pull_request: {}` makes this run on every PR on a macos-14 runner, doc-only ones included. The sync stays off (`REPO=... && EVENT=push` leaves a PR in the else branch, sync=false, so no token is needed and forks are safe), which is the right guard. It's the runner cost I'd think about, since this is the third macOS job per PR. If the point is to catch a missing OVERLAY row at PR time, `paths:` on `src/mflux/models/common/config/model_config.py` and `scripts/ci_extract_models.py` would catch every case that matters and skip the rest.

The release-note check is red because the block in the description is empty; `none` in it turns it green, since this is CI and agent docs.

fxd0h (dismissed):
The AGENTS.md line is gone, so my blocking point is settled. Ran the extract here again on your head: it writes the 36 models with qwen-image-2.1 in them, and main still exits with "no OVERLAY entry". The only thing left is the release-note check, which is red because the block in the description is empty; `none` inside it turns it green for a CI and agent-docs change. One nit while you are in there: the last commit left .gitignore without a trailing newline.

anthonywu:
looks like the agent that generated this PR body failed to put the release note body in the fenced code block, I went ahead and made an edit

anthonywu (approved):
@ianscrivener I fixed the failed release note check. going ahead with merging.

PR #746, opened 2026-09-22
fix: use @cache instead of @lru_cache(max_size=1) for zero-arg functions

akx:
## What

Small cleanup I noticed on the way to do something else: [`functools.cache`](https://docs.python.org/3.14/library/functools.html#functools.cache) has been a thing since Python 3.9; it's documented to be faster than a sized `lru_cache`, and arguably easier to read.

Follows up on #706.

Nothing new, internal change.

## Release note

```release-note
none
```

## Verification

Not run, trivial change.

## Summary by CodeRabbit

* **Refactor**
  * Simplified the handling of model and configuration lookups.
  * Preserved existing lookup behavior and caching characteristics.
  * No user-facing functionality or interface changes.

<h2><a href="https://app.greptile.com/api/retrigger?id=67352163"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge because the new decorators are behaviorally equivalent for these zero-argument functions and are supported by the repository's minimum Python version.

<h3>Summary</h3>

This PR replaces four single-entry `lru_cache` decorators on zero-argument registry lookup functions with the equivalent `functools.cache`.
- Preserves one-time computation of CLI model choices and canonical choices.
- Preserves one-time computation of model-resolution names and keys.
- Remains compatible with the package's Python 3.10 minimum.

<sub>Reviews (1) · Last reviewed commit: ["fix: use @cache instead of @lru\_cache(ma..."](https://github.com/mflux-community/mflux/commit/628f0b8869a737835fc42f927e50d69e7daa16a0)</sub>


anthonywu (approved):
small, safe, but perf-wise is a no-op, accepting as a idiomatic python style improvement

akx:
> small, safe, but perf-wise is a no-op

If these functions were hotter, 3% might make an actual difference! 😄 

```
$ python3 -m timeit -s 'from functools import lru_cache
            @lru_cache(maxsize=1)
            def foo():
              return 8
            ' 'foo()'
10000000 loops, best of 5: 28.5 nsec per loop
$ python3 -m timeit -s 'from functools import cache
            @cache
            def foo():
              return 8
            ' 'foo()'
10000000 loops, best of 5: 27.8 nsec per loop
```

PR #749, opened 2026-09-23
feat: Qwen-Image-2.1 edit/instruction variant with prefix KV cache

flyingtimes:
## What

Implements the edit/instruction variant for Qwen-Image-2.1 — the biggest remaining capability gap listed in this model's README ("Not yet supported: the edit/instruction variant (needs the Qwen3-VL vision tower) …"). Also includes the official prefix KV cache for editing, a fix for a t2i text-encoder weight-mapping bug discovered while testing, the new `mflux-generate-qwen-2.1-edit` CLI, unit tests, and README updates.

No additional model download is required: the Qwen3-VL vision tower needed for editing ships inside the existing `text_encoder` shards (`model.visual.*`), and this PR maps those weights for the edit variant.

- [x] Tests added/updated run in CI by default (`-m "not slow and not high_memory_requirement"`): `tests/image_generation/test_qwen_image_21_edit_transformer.py` is weight-free and `@pytest.mark.fast`.
- [x] `ruff check` and `ruff format` clean (ruff 0.16.3, the pinned version).
- [x] Release note block filled in.
- [x] Docs updated: `src/mflux/models/qwen21/README.md` (editing docs, prefix KV cache, `Not yet supported` now lists only LoRA mappings and PID decoding).
- [x] New model wiring: N/A — this is a new variant of the existing `qwen-image-2.1` model (reuses its `ModelConfig`, weights, and completions); a thin CLI entrypoint was added.
- [x] New/changed CLI: no option is accepted-but-ignored (the img2img-only flags come from `add_image_to_image_arguments`, which this CLI does not add), so nothing is declared in `IGNORED_OPTIONS`; all accepted options are honored.

## Release note

```release-note
Qwen-Image-2.1: new `mflux-generate-qwen-2.1-edit` command for natural-language image editing with up to 10 reference images, including a prefix KV cache that makes multi-reference editing ~3.6x faster. Also fixes randomly-initialized text-encoder MLP weights in Qwen-Image-2.1 txt2img (the weights were never mapped).
```

## Verification

All commands run inside the repo venv (mlx 0.32.0, transformers 5.15.0) on M2 Ultra, bf16, `Qwen/Qwen-Image-2.1` local weights.

**Unit tests (weight-free, CI default selector):**

```
uv run pytest tests/image_generation/test_qwen_image_21_edit_transformer.py -q
# 3 passed
uv run pytest tests -m fast -q
# 1502 passed, 6 skipped (no regressions)
uv run ruff check && uv run ruff format --check
# clean
```

**Numeric parity vs references (transformers 5.x / diffusers `pipelines/qwenimage21`):**

| Check | Result |
|---|---|
| `Qwen21ImageProcessor` vs HF `Qwen3VLProcessor` preprocessing | byte-identical (diff 0.0) |
| Vision tower embeddings | mean abs diff 0.022 (bf16 noise; ref scale 0.19) |
| Deepstack features [0..2] | diff 0.003-0.008 |
| mrope positions vs `get_rope_index` | exactly equal |
| Full encoder fp32 vs transformers | mean abs diff 0.0019 (ref scale 1.57) |
| Edit forward vs validated t2i forward, text-only layout | exact 0.0 |
| KV cache vs uncached, 8-step edit (512², single ref) | pixel diff mean 0.055/255, max 8/255 |

**End-to-end edits (40 steps unless noted):**

- Single-image instruction edit at 512²/20 steps — prompt honored, scene preserved:
  https://raw.githubusercontent.com/flyingtimes/mflux/demo-assets/demo/edit_multi.jpg (two-image cross-reference replacement with different aspect ratios)
- Three-image composition (animal from image 1 + person from image 2 placed in the scene from image 3, identities preserved):
  https://raw.githubusercontent.com/flyingtimes/mflux/demo-assets/demo/edit_multi3_meet.jpg (1024², 40 steps)

**Prefix KV cache A/B (identical seed and settings):**

```
use_kv_cache=False: 14.9 s total,  512² 8-step single-reference edit
use_kv_cache=True:   8.9 s total   -> 1.67x
# pixel diff mean 0.06/255 (bf16 reduction order), max 17/255

three-reference 1024² 40-step edit:
use_kv_cache=False: 28.0 s/step (18.7 min total)
use_kv_cache=True:   7.7 s/step ( 5.2 min total)  -> 3.6x
```

**Attention memory:** the edit prefill uses the reference's exact multi-pass segmented
attention (one SDPA per template-ordered run; text runs causal over their prefix via a
small run×prefix additive mask, image runs unmasked over their prefix). No quadratic dense
mask tensor is materialized — peak mask memory is a few MB instead of up to ~3.8 GiB at ten
1024² references.

Example outputs (from the fork's `demo-assets` branch, generated with this PR):

3-image edit
2-image edit

## Summary by CodeRabbit

* **New Features**
  * Added Qwen Image 2.1 image editing through a new command-line workflow.
  * Supports one or more reference images, prompts, custom dimensions, seeds, guidance, and optional metadata output.
  * Automatically resizes images, supports RGBA editing masks, and derives output dimensions from reference images when unspecified.
  * Added image understanding and prefix KV caching for faster generation.

* **Documentation**
  * Added guidance on image-editing usage, image inputs, output dimensions, and KV caching.


flyingtimes:
Thanks for the review — all findings are addressed in 6384021 and 4435d0f:

1. **use_kv_cache override** — fixed; the parameter now controls the cache. Re-measured a valid A/B with the fixed code: 14.9 s (uncached) vs 8.9 s (cached) on a 512² 8-step single-reference edit (1.67×), pixel diff mean 0.06/255 (bf16 reduction order).
2. **CLI dimensions** — fixed; at `--width auto --height auto` (the parser defaults) dimensions stay unresolved so `generate_image` derives its ~1MP defaults from the last condition image.
3. **README flag** — fixed (`--image-paths`).
4. **Dense quadratic mask** — replaced with the reference's exact multi-pass segmented prefill (one SDPA per template-ordered run; text runs causal over their prefix via a small run×prefix mask, image runs full over their prefix). No quadratic mask tensor is materialized at all — peak mask memory drops from ~3.8 GiB (ten 1024² refs, bf16) to a few MB. The mask is also no longer cached.
5. **Empty image_paths** — `generate_image` now raises a clear `ValueError` at entry; no artificial upper bound added, per the suggestion.
6. Nitpicks applied: `build_mrope_positions`/`_get_text_encoder_mapping_with_visual`/`_calculate_dimensions` moved onto their classes, and long docstrings shortened to comments.

PR description rewritten against the repository template with the required What / Checklist / Release note / Verification sections and demo images.

src/mflux/models/qwen21/cli/qwen21_edit_generate.py
```diff
@@ -0,0 +1,81 @@
...
+        model=qwen,
+        latent_creator=Qwen21LatentCreator,
+    )
+
+    try:
+        image_paths = [str(p) for p in args.image_paths]
+        # "auto" (the parser default) keeps dimensions unresolved so the variant derives
+        # its ~1MP defaults from the last condition image's aspect ratio
+        width = height = None
+        if str(args.width) != "auto" and str(args.height) != "auto":
```
anthonywu, line 53: [P1] Preserve the automatic ~1 MP output size

The shared parser converts the default `"auto"` to `ScaleFactor(1)`, whose string representation is `"1x"`. This condition therefore evaluates to true even when neither dimension was supplied, and `DimensionResolver` uses the source image dimensions instead of allowing `generate_image()` to derive the documented ~1 MP target. I reproduced a 3024×4032 output target from a phone-sized JPEG with no dimension flags, which substantially increases attention cost and memory requirements.

Could we preserve whether each dimension was explicitly supplied and pass `None` for automatic dimensions, while retaining explicit scale-factor support? A parser-to-generation regression test using a large source image would catch this.

src/mflux/models/qwen21/variants/edit/qwen_image_21_edit.py
```diff
@@ -0,0 +1,288 @@
...
+        height: int | None = None,
+        width: int | None = None,
+        guidance: float = 1.0,
+        negative_prompt: str | None = None,
+        scheduler: str = "linear",
+        output_resolution: int = 1024,
+        use_kv_cache: bool = True,
+    ) -> GeneratedImage:
+        # Normalize inputs to PIL up front (same normalization order as the reference).
+        images = [img if isinstance(img, Image.Image) else Image.open(img) for img in image_paths]
```
anthonywu, line 65: [P2] Apply EXIF orientation before preparing reference images

For a portrait JPEG stored landscape with EXIF orientation 6, `DimensionResolver` correctly resolves the displayed portrait dimensions, but this raw `Image.open()` leaves the pixels unrotated. Both the vision and VAE paths then encode a sideways reference. I reproduced CLI dimensions of 3024×4032 while this code reads the reference as 4032×3024; the Python API also derives its default aspect ratio from the unrotated size.

Could we use the existing EXIF-aware image-loading helper before calculating aspect ratios or preparing either encoder input, preserving the RGBA alpha channel? An orientation-6 fixture would make a useful regression test.

tests/image_generation/test_qwen_image_21_edit_transformer.py
```diff
@@ -0,0 +1,79 @@
...
+        num_inference_steps=4,
+    )
+    rng = np.random.default_rng(7)
+    text = mx.array(rng.standard_normal((1, 12, 4096)).astype(np.float32)).astype(mx.bfloat16)
+    latents = mx.array(rng.standard_normal((1, 16, 64)).astype(np.float32)).astype(mx.bfloat16)
+
+    out_t2i = transformer(t=0, config=config, hidden_states=latents, encoder_hidden_states=text)
+    out_edit = transformer.__call_edit__(t=0, config=config, target_latents=latents, layout=[("text", text)])
+
+    np.testing.assert_array_equal(np.array(out_t2i.astype(mx.float32)), np.array(out_edit.astype(mx.float32)))
```
anthonywu, line 79: Suggested weight-free regression tests

Could we add fast tests for these paths as well as the existing text-only transformer parity test?

1. **CLI automatic dimensions:** create a synthetic 4032×3024 reference, run the real parser/main with the model constructor and generation mocked, and capture `generate_image()` arguments. With no dimension flags, assert `width` and `height` remain `None`, allowing the API's ~1 MP sizing. Also cover explicit integer dimensions and explicit scale factors so the fix preserves those behaviors.

2. **EXIF orientation through both input paths:** create an asymmetric image with distinguishable colored corners, save it with EXIF orientation 6, and use it as the reference. Intercept preprocessing/VAE inputs in the edit generation path (stub the heavy models) and assert that both receive the correctly rotated content and portrait aspect ratio. Compare against `ImageOps.exif_transpose()` applied to the reopened fixture, avoiding JPEG round-trip differences. Test both a file path and a PIL input. Include an RGBA case to ensure normalization preserves alpha for the VAE while the vision path composites it over white. Testing only the existing orientation helper would miss whether this new pipeline actually calls it.

3. **Reference-image KV-cache parity:** use a small randomly initialized transformer with a layout containing text, two reference-image runs of different shapes, and trailing text. Compare uncached output with cache extraction at the first timestep, then compare a cached pass with a fresh uncached pass at a different timestep and with different target latents. Use an appropriate numerical tolerance, not exact equality. This directly exercises the new default cache path, which the current text-only test does not cover.

Local evidence: I reproduced the dimension/orientation mismatch using a synthetic orientation-6 JPEG (CLI target 3024×4032; raw reference pixels 4032×3024), without full-weight inference. A separate small fp32 two-reference transformer check gave maximum absolute differences of about 7.2e-7 for prefill and 8.4e-7 for the subsequent cached pass. These checks are not yet committed regression tests.

src/mflux/models/qwen21/cli/qwen21_edit_generate.py
flyingtimes: Fixed in 5b71592. `resolve_output_dimensions()` now inspects the parsed values' types instead of their string form: both dimensions auto (`ScaleFactor(1)`) map to `(None, None)` so `generate_image()` derives the ~1MP target from the last condition image; explicit `ScaleFactor` values resolve through `DimensionResolver` against that image; plain integers pass through and the variant applies its own /16 rounding. Covered by `test_cli_auto_dimensions_and_scale_factors` (auto, integer, and scale-flag combinations).

src/mflux/models/qwen21/variants/edit/qwen_image_21_edit.py
flyingtimes: Fixed in 5b71592. Condition images are now loaded via `open_oriented()`, so both the aspect-ratio derivation and the vision/VAE inputs see the displayed orientation. RGBA is preserved for the VAE (alpha channels matter to the causal VAE); the vision path still receives its own preprocessed RGB copy. Covered by the new `test_qwen_image_21_edit_exif.py`, which pushes an orientation-6 PNG through the full `generate_image()` path with stubbed encoders and asserts the rotated content reaches both the vision preprocessor and the VAE.

tests/image_generation/test_qwen_image_21_edit_transformer.py
flyingtimes: Both added in 5b71592:

1. `test_cli_auto_dimensions_and_scale_factors` (tests/image_generation/test_qwen_image_21_edit_transformer.py) runs the real CLI `main()` with the model constructor and generation mocked, and asserts the captured `generate_image()` arguments: no flags → `width`/`height` stay `None`; explicit integers pass through; explicit scale factors resolve against the reference.

2. `test_exif_orientation_reaches_both_encoders` (tests/image_generation/test_qwen_image_21_edit_exif.py) builds an asymmetric landscape PNG carrying EXIF Orientation 6, runs `generate_image()` with the vision preprocessor/VAE/transformer stubbed, and asserts: the vision path sees the portrait RGB image at the area-normalized size, the VAE input is a portrait RGBA tensor whose marker quadrant proves a 90° CW rotation (not CCW), and alpha is preserved opaquely.

All 1511 fast tests pass on the merged branch.

flyingtimes:
Thanks @anthonywu for the careful review — all three points are addressed in 5b71592:

**[P1] Automatic ~1 MP output size** — `resolve_output_dimensions()` is now ScaleFactor-type-aware (the previous `str()` comparison never matched `"1x"`). Both dimensions auto → `(None, None)` so `generate_image()` derives the ~1MP target from the last condition image; explicit scale factors resolve through `DimensionResolver`; plain integers pass through.

**[P2] EXIF orientation** — condition images load via the existing `open_oriented()` helper before aspect-ratio derivation and encoder preparation, so both the vision and VAE paths see the displayed orientation. RGBA is preserved for the VAE.

**Regression tests** — both requested suites added, no weights required:
- `test_cli_auto_dimensions_and_scale_factors`: runs the real CLI `main()` with the model mocked; asserts auto flags keep `width`/`height` `None`, integers pass through, scales resolve against the reference.
- `test_qwen_image_21_edit_exif.py`: an orientation-6 PNG flows through `generate_image()` with stubbed encoders; asserts the vision path receives the portrait RGB image at the area-normalized size, the VAE input is portrait RGBA with the marker quadrant proving a correct 90° CW rotation, and alpha is preserved.

Also included the dtype casts from CodeRabbit's finding (vision/deepstack embeds cast back to `hidden_states.dtype`). Full fast suite passes on the merged branch: 1511 passed, 6 skipped.

flyingtimes:
Follow-up round: I re-verified the feature against the reference pipeline (diffusers `QwenImage21Pipeline`) using an execution matrix of real inputs — extreme aspect ratios, every PIL mode (L/LA/RGBA/P+alpha/CMYK/I;16/1-bit), mixed dimension flags, multi-image sets — with only the heavy models stubbed. That surfaced four issues, all fixed in 86b8021 with weight-free regression tests:

1. **Per-axis output dimensions** — `generate_image(width=1024, height=None)` overwrote both axes from the aspect ratio (produced 736 wide). Now an explicit axis is kept and only the missing one derives from the last condition image, matching the reference's `height = height or calculated` semantics. This also makes the Python API and the CLI consistent for partially-specified dimensions.

2. **Extreme aspect ratios crash with unhelpful errors** — a 300:1 panorama died inside vision preprocessing (`absolute aspect ratio must be smaller than 200, got 277`), and ratios beyond ~4000:1 died with `height and width must be > 0` after `_calculate_dimensions` rounded a side to zero. Both are now rejected up front with a `ValueError` naming the image and its ratio, checked on the /32-rounded dimensions that actually reach the encoder (rounding can push a ~199:1 input past the encoder's own limit).

3. **`generation_time` metadata always 0.0** — the variant read `elapsed` from a local tqdm that was never iterated; it now reads the Config-owned tqdm the denoising loop advances (as the t2i variant does).

4. **`guidance > 1` without a negative prompt silently skipped CFG** — now logs a warning, mirroring the reference pipeline's message.

Positive findings from the same matrix, for the record: all PIL modes flow through both encoders correctly, multi-image mixed-mode sets build consistent layouts, CLI mixed flags (int + auto, scale + scale) resolve correctly, and the KV cache/CFG branch structure matches the reference. Full fast suite: 1515 passed, 6 skipped.

fxd0h (commented):
Ran the branch here (M5 Max): the fast suite is 1688 passed, and the existing Qwen 2.1 text-to-image tests are among them, which matches what I see in the diff: the new `kv_mode` and `segments` arguments default to None and the old path falls through untouched, so a user who never edits gets the same forward pass. I approved the CI runs so lint and tests could report; tests are green, lint is red on the three new test files only (unused imports, an unsorted import block, an unused `out_extract`, `sys` imported twice) and `ruff format` wants to touch the same three files. `uv run ruff check --fix` plus `ruff format` clears all of it.

One thing I'd change before merging. `DimensionResolver.is_auto` treats `ScaleFactor(1)` as "not given", and the parser does produce that for the default. But it produces the same object for an explicit `--width 1x`, and the help text on those flags says the default is the source image size, which is what `1x` means on every other edit command. Here both become `(None, None)` and the ~1 MP target, so someone asking for the reference's own size gets a resize. The parser already has `_option_was_provided`, so the two cases are separable; otherwise the help text for this command has to say its default is ~1 MP, not the source size.

What I could not check: parity with `QwenImage21Pipeline` (no diffusers revision with it here) and a real render (no Qwen-Image-2.1 weights on this machine), so the prefix cache and the segmented prefill are reviewed by reading only. The EXIF fix is where it should be, applied to every reference image before anything measures it.

For the maintainers rather than the author: this and #741 both add reference-image editing for Qwen-Image-2.1, with prefix caching in both. Someone needs to pick one direction before either author invests another round.

flyingtimes:
Thanks @fxd0h for running the branch and the CI approvals — both points addressed in 4281421:

**Lint** — `ruff check --fix` + `ruff format` applied to the three new test files (unused imports, unsorted import block, duplicate `sys` import), and the unused `out_extract` binding removed with a comment noting the extract pass only fills the cache. Repo-wide `ruff check` passes.

**Explicit `1x` vs the auto default** — fixed via `_option_was_provided` as you suggested. `resolve_output_dimensions` now takes `dims_specified`: with no dimension flag on the command line, both axes stay `None` and the variant derives its ~1MP target (the reference-pipeline default); once any dimension flag is given, the shared flags' documented semantics apply — scale factors resolve against the reference image, so `--width 1x --height 1x` yields the source's own displayed size, and an axis left on its default means 1x the source size. The ScaleFactor(1)-sniffing `is_auto` helper is gone. New test case pins `1x` to the source size.

On the #741 overlap: happy to defer to the maintainers on direction — whichever implementation they prefer to move forward with, and glad to help reconcile anything useful from the other branch either way.

flyingtimes:
New capabilities on top of the review fixes (b5ec489), all validated against the real checkpoint on M2 Ultra (q8, 20 steps, 1024x1024):

**1. Mask-guided inpainting / outpainting** — `generate_image(..., mask_image=...)` (CLI `--mask-image`): white = repaint, black = preserve, aligned with the first condition image. During denoising the unmasked latent tokens are pulled onto the reference's own noised flow trajectory each step, and the decoded image is composited with the original pixels outside the mask — measured **bitwise-identical outside the mask**. Outpainting falls out of the same mechanism (paste onto a larger canvas, mask the border).

**2. Auto-masking via visual grounding** — `generate_image(..., auto_mask="the dress")` (CLI `--auto-mask`): the text-encoder checkpoint ships a full Qwen3-VL language model whose `lm_head.weight` is **untied and was previously unmapped**; `Qwen21TextEncoder.locate_object()` now loads it and decodes greedily over a per-layer KV cache, parses the JSON bbox (handles fraction / absolute-pixel / 0-1000-normalized coordinate regimes), feathers it (box grown 3% per side) and feeds the inpainting path. On the validation image "the dress" returned `[359, 168, 580, 783]`, which lands exactly on the dress. One integration note: models that assumed tied embeddings were silently dropping `lm_head.weight`; nothing else consumed it, so the mapping addition is inert for the t2i path (strict=False update).

**3. First-block step cache** — `generate_image(..., use_step_cache=True, step_cache_threshold=0.12)` (CLI `--use-step-cache`): block 0 always runs; when the accumulated relative-L1 signal of consecutive steps stays under the threshold, blocks 1..N reuse the previous step's target hidden with the output norm re-applied at the current timestep. Skipping engages only on the KV-cached path so a prefix-cache extract can never be interrupted. Measured **1.21x end-to-end** at threshold 0.12 (~1.5x denoising-only; the same run also paid the grounding prefill). Output is close to but not pixel-identical with the baseline — it is opt-in.

Validation artifacts: side-by-side outputs in this run — manual-mask baseline, auto-mask + step cache (both show the dress repainted red with pose/face/background exactly preserved); unmasked-region pixel diff 0.0 (baseline) and 0.007/255 (auto-mask, from the wider feathered box). Weight-free fast tests cover the blend math, coordinate parsing regimes, rasterization, cache skip/recompute, and an auto-mask end-to-end run with stubbed models: full fast suite 1526 passed, 6 skipped.

flyingtimes:
Third round of capabilities (0218c10), again all validated on the real checkpoint (M2 Ultra, q8, 20 steps):

**Edit strength** — `generate_image(..., strength=0.6)` (CLI `--strength`): 1.0 stays the full re-denoise; lower values start partway down the sigma schedule from the reference's own latent noised to that sigma (the img2img machinery via `Config.init_time_step`). Measured gradient on the validation portrait, mean pixel diff vs original: **0.3 → 3.8 (subtle retouch, composition identical), 0.6 → 3.1, 1.0 → 22.5**. Composes with inpaint masks.

**Prompt enhancement** — `--enhance-prompt`: the official `prompt_rewrite` serving recipe implemented with the in-memory Qwen3-VL (no separate 9B checkpoint; our port follows the official system prompt's rules and its JSON answer format, including the tolerated `rewrited_prompt` typo key and balanced-JSON scan). Measured: *"change her dress to a red evening gown"* → a 40-word description naming the fitted silhouette, floor-length hem, lace neckline, and explicit preservation constraints, in 4.7 s; any parse failure falls back to the original instruction. The A/B output with *"make the dress golden"* shows the rewritten prompt's satin/embroidery detail actually reflected in the image.

**Self-verification** — `--verify` (+ `--verify-retries N`): after generating, the original and the output are shown together to the same Qwen3-VL, which returns a structured verdict stored on `GeneratedImage.verification` and surfaced by the CLI and api_server; failed verdicts can trigger seed-jittered retries. Grounded decoding was generalized to multi-image prompts (`locate_object` now delegates to a shared `generate()`). On the validation pass-case it correctly returned `{"instruction_applied": true, "outside_unchanged": true}`; an engineered "impossible" case turned out to be a *feasible* edit for this model (it did replace the background), and the verifier correctly called it applied — the fail→retry path is covered by weight-free unit tests, and honestly engineering a real failure case that survives 20 steps is itself getting hard.

Fast suite: 1533 passed, 6 skipped. CLI surface now: `--mask-image / --auto-mask / --use-step-cache / --strength / --enhance-prompt / --verify / --verify-retries`.

